PR #4, opened 2018-05-19
[docs] Fix misleading build/hooks -> hooks/build

avo7:


avo7:
Nah, just a couple of minutes ;) Thanks for project, man!

PR #9, opened 2018-05-25
add command to deploy in local and give parameters to serverless invoke

Fabex:
In order to deploy on local machine with serveless I need to give some paremeters to the invoke function

### DIFF
--- a/bref
+++ b/bref
@@ -52,6 +52,104 @@ $app->command('init', function (SymfonyStyle $io) {
     ]);
 });
 
+$app->command('local [data]', function(?string $data, SymfonyStyle $io) {
+    $fs = new Filesystem;
+    $commandRunner = new CommandRunner;
+
+    if (!$fs->exists('serverless.yml') || !$fs->exists('bref.php')) {
+        throw new Exception('The files `bref.php` and `serverless.yml` are required to deploy, run `bref init` to create them');
+    }
+
+    // Parse .bref.yml
+    $projectConfig = [];
+    if ($fs->exists('.bref.yml')) {
+        /*
+         * TODO validate the content of the config, for example we should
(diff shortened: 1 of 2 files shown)


bref
```diff
@@ -52,6 +52,104 @@ $app->command('init', function (SymfonyStyle $io) {
...
+     * - handler.js
+     * - .bref/**
+     */
+
+    // Run build hooks defined in .bref.yml
+    $buildHooks = $projectConfig['hooks']['build'] ?? [];
+    foreach ($buildHooks as $buildHook) {
+        $io->writeln('Running ' . $buildHook);
+        $commandRunner->run('cd .bref/output && ' . $buildHook);
+    }
```
mnapoli: Thanks for the PR! If I'm not mistaken everything above this is duplicated from the `deploy` command? Maybe that code could be shared instead, moved to a class maybe? I'm thinking a `generate()` method in a  `src/Console/Deployer` class for example.

bref
```diff
@@ -52,6 +52,104 @@ $app->command('init', function (SymfonyStyle $io) {
     ]);
 });
 
+$app->command('local [data]', function(?string $data, SymfonyStyle $io) {
```
mnapoli: Could you add a comment above this line? Something that says something like "Runs the lambda locally".

bref
Fabex: Yes you are right, it's a big duplication of deploy command. I would just propose you this feature in a simple way to saw if it's was a good idea.
I will refacto this and send you a new commit

bref
Fabex: ok no problem

mnapoli:
Thanks, I have built upon this PR (the code wasn't working) and opened #11. I have kept your commit, I'll be closing this one and merging #11.

PR #14, opened 2018-06-08
Enable opcache and use it with the file cache

mnapoli:
Opcache usually works in memory, compiling PHP code to opcode and caching it in memory between HTTP requests. Since on lambdas the PHP process is terminated between requests, the classic behavior of opcache doesn't work (the cache is cleared every time).

To make opcache work we enable its "file cache" mode where opcache writes the opcode into a file (with the `opcache.file_cache` option). When the PHP process starts, it reads the opcodes from the file cache.

### DIFF
--- a/bin/php/Dockerfile
+++ b/bin/php/Dockerfile
@@ -38,6 +38,8 @@ RUN ./configure \
     # --enable-mbstring is included here because otherwise there's no way to get pecl to use it properly (see https://github.com/docker-library/php/issues/195)
     --enable-mbstring \
     --enable-mysqlnd \
+    # Allows to use the opcache.file_cache option
+    --enable-opcache-file \
     --enable-soap \
     --enable-zip \
     --with-curl \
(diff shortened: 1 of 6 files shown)


mnapoli:
OK I think I understand why I don't see any difference in performances, what I did is not enough for opcache to get loaded. I need to bundle the `opcache.so` file too it seems.

mnapoli:
On the API platform demo, when getting a list of books (uses the database):

- without opcache: 220ms
- with opcache: 105ms

This is very good news, it means opcache in CLI mode is very useful too.

fgilio:
Impressive! Thanks 🙌🏼

staabm:
to be honest, I dont get why the file-based opcache works. will the opcache file, which gets created on the first visit be persisted with the lambda and is available on subsequent requests?

or do you somehow pre-generated the opcache while deployment or similar, and therefore the opcache-file-cache is part of the "software-package" as a whole?

mnapoli:
:+1: and I want to explore the possibility of generating the opcache file before deploying. That would probably require using a Docker image with the exact same PHP version and same file paths, but I was told on Twitter that this could be doable.

PR #18, opened 2018-06-12
Speed up building the lambda

mnapoli:
With those changes the lambda builds much faster. For example with a simple use case (23 seconds faster):

capture d ecran 2018-06-12 a 13 05 58

You can check [the Blackfire profile](https://blackfire.io/profiles/compare/9812960e-34d6-4650-a65e-e6c6ff0b4fb3/graph) if you are interested.

This will help when:

- deploying
- running the lambda locally

### DIFF
--- a/src/Filesystem/DirectoryMirror.php
+++ b/src/Filesystem/DirectoryMirror.php
@@ -1,7 +1,7 @@
 <?php
 declare(strict_types=1);
 
-namespace Bref;
+namespace Bref\Filesystem;
 
 use Symfony\Component\Filesystem\Exception\IOException;
 use Symfony\Component\Filesystem\Filesystem;
--- a/composer.json
+++ b/composer.json
@@ -25,8 +25,7 @@
         "aws/aws-sdk-php": "^3.44",
         "psr/http-server-handler": "^1.0",
         "zendframework/zend-diactoros": "^1.6",
-        "jolicode/jolinotif": "^2.0",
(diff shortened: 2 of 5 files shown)


src/Filesystem/DirectoryMirror.php
```diff
@@ -0,0 +1,127 @@
...
+        $targetFiles = iterator_to_array($this->indexFilesCTimeByRelativePath($target));
+
+        return [$sourceFiles, $targetFiles];
+    }
+
+    private function indexFilesCTimeByRelativePath(iterable $files) : \Traversable
+    {
+        foreach ($files as $file) {
+            /** @var SplFileInfo $file */
+            yield $file->getRelativePathname() => $file->getCTime();
```
mnapoli: Fun part here :)

src/Filesystem/DirectoryMirror.php
```diff
@@ -0,0 +1,127 @@
...
+
+use Symfony\Component\Filesystem\Exception\IOException;
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Finder\Finder;
+use Symfony\Component\Finder\SplFileInfo;
+
+/**
+ * @author Matthieu Napoli <matthieu@mnapoli.fr>
+ */
+class DirectoryMirror
```
mnapoli: This could be an open source package when I get the motivation…

PR #22, opened 2018-06-12
Simplify how to invoke lambdas locally

mnapoli:
This follows https://github.com/mnapoli/bref/issues/17#issuecomment-396575392 (ping @t-geindre)

### Before

Test a basic lambda locally (`simpleHandler`):

```shell
vendor/bin/bref local
# This will generate the lambda locally in a separate directory, it will NOT use the dev setup
```

Test a CLI command locally (`cliHandler`):

```shell
php bref.php <my-command>
# This will NOT generate the lambda, instead the dev setup is used
```

Test a HTTP application locally (`httpHandler`):

```shell
php -S 127.0.0.1 bref.php
# This will NOT generate the lambda, instead the dev setup is used
```

### After

Test a basic lambda locally (`simpleHandler`):

```shell
php bref.php bref:invoke
# This will NOT generate the lambda, instead the dev setup is used
```

**(unchanged)** Test a CLI command locally (`cliHandler`):

```shell
php bref.php <my-command>
# This will NOT generate the lambda, instead the dev setup is used
```

**(unchanged)** Test a HTTP application locally (`httpHandler`):

```shell
php -S 127.0.0.1 bref.php
# This will NOT generate the lambda, instead the dev setup is used
```

### Summary of the changes

Bref now auto-adds a `bref:invoke` command in the CLI handler. This command can be invoked locally to test the lambda's `simpleHandler`.

Advantages:

- much much faster (the lambda is not generated like it was going to be deployed)
- uses the development environment, like the other methods
- more consistent: all helpers for testing the lambda involve invoking `bref.php` directly

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,10 @@ $payload = json_decode($result->get('Payload')->getContents(), true);
 Bref provides a helper to invoke the lambda locally, on your machine instead of the serverless provider:
 
 ```shell
-$ vendor/bin/bref local
+$ php bref.php bref:invoke
+
+# If you want to pass event data:
+$ php bref.php bref:invoke --event='{"name":"foo"}'
 ```
 
 ## Deletion
@@ -202,11 +205,19 @@ Hello Bob
 
 As you can see, all arguments and options after `bref cli --` are forwarded to the CLI command running on lambda.
 
-To test your CLI commands locally, simply run:
+To test your CLI commands locally (on your machine), simply run:
 
 ```shell
 $ php bref.php <commands and options>
 ```
+
+As you may have seen above, there is a special command that is automatically added to your CLI application:
+
+```shell
+$ php bref.php bref:invoke
+```
+
+That special command lets you invoke the `simpleHandler` on your development machine.
 
 ## Build hooks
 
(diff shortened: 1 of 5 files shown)


t-geindre:
I think the local command is still useful to check the package content before deploying it. But maybe some kind of a dry run on the deploy command could do the job.

Otherwise, I like this new way to run the lambda locally as it avoids the useless build step.

mnapoli:
Thanks for the feedback. What would you test before deploying? That's maybe something which needs a broader solution 🤔 I was thinking about test helpers for PHPUnit for example.

t-geindre:
For example, if I am working on the build step (hook), I might want to check several times the generated package content.

And I think in that case it doesn't make sense to deploy the lambda each time. 

That's why I was thinking about a dry run.

mnapoli:
OK so maybe a `bref generate` command would make sense? (it would generate the lambda but not deploy it)

Or even a `bref deploy --dry-run` ? That avoids a new command, I think it's better…

t-geindre:
Yes a `bref deploy --dry-run` is exactly what I had in mind.

mnapoli:
:+1: I'll merge this and open a new issue.

PR #23, opened 2018-06-12
Add new commands

### LINKED ISSUE #19: Write a command to remove a lambda
Currently we have to run `serverless remove`.

The goal would be to have a `bref remove` command that runs `serverless remove`.

The `bref info` command works like that so it can be a good example: https://github.com/mnapoli/bref/blob/d4cf7a8470c63213b0c37c33dfd9d675aadc00d1/bref#L119-L121

### LINKED ISSUE #20: Write a command to invoke the lambda remotely
Currently we have to run `serverless invoke -f main`.

The goal would be to have a `bref invoke` command that runs `serverless invoke -f main`.

The `bref info` command works like that so it can be a good example: https://github.com/mnapoli/bref/blob/d4cf7a8470c63213b0c37c33dfd9d675aadc00d1/bref#L119-L121

### LINKED ISSUE #21: Write a command to show the lambda's logs
Currently we have to run `serverless logs -f main`.

The goal would be to have a `bref logs` command that runs `serverless logs -f main`.

The `bref info` command works like that so it can be a good example: https://github.com/mnapoli/bref/blob/d4cf7a8470c63213b0c37c33dfd9d675aadc00d1/bref#L119-L121

davidavz:
Add new commands : 

- bref remove (fix #19)
- bref invoke (fix #20)
- bref logs (fix #21)

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,9 +86,7 @@ $ curl https://xxxxx.execute-api.xxxxx.amazonaws.com/dev/
 Triggering a lambda manually:
 
 ```shell
-# "main" is the name of the function created by default
-# you can have several functions in the same projects
-$ serverless invoke -f main
+$ vendor/bin/bref invoke
 ```
 
 Triggering a lambda from another PHP application:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -116,7 +116,7 @@ $ vendor/bin/bref local
 ## Deletion
 
 ```shell
-$ serverless remove
+$ bref remove
```
mnapoli: Could you use `vendor/bin/bref` to be consistent with the rest of the documentation?

mnapoli:
Thanks!

PR #28, opened 2018-06-15
Allow PHP archive to provide a php.ini file

### LINKED ISSUE #26: Load extension dynamically
As done in #14 with `opcache.so`, I tried to load more extensions dynamically.

I was able to compile Redis and MongoDB extensions with a bit tweaked version of `bin/php/Dockerfile` and `bin/php/build.sh`.

Then I created a tar archive containing those two extensions in the `ext` directory.

And finally I used the following hook to make php binary load those extensions:

```yaml
hooks:
    build:
        # Add custom extensions to php.ini
        - 'echo "extension=mongodb.so" >> .bref/php.ini'
        - 'echo "extension=redis.so" >> .bref/php.ini'
```

A `phpinfo()` confirms that those two extensions are correctly loaded.

I think this could be done in a better way, especially the `php.ini` update.

What do you think?

t-geindre:
Bref allow us to provide an archive containing our version of PHP, that's why I think we should also be able to provide the related php.ini file.

Bref install a default php.ini file containing opcahe configuration but there is no guaranty that related extensions will be available in the archive.

Just to keep bref working with existing archives, I left the default php.ini as a fallback if there's no php.ini file in the archive.

As mentioned in #26, this would allow us to embed more extensions in the archive and load them.

fix #26

### DIFF
--- a/src/Console/Deployer.php
+++ b/src/Console/Deployer.php
@@ -159,8 +159,13 @@ class Deployer
             ->mustRun();
         // Set correct permissions on the file
         $this->fs->chmod('.bref/output/.bref/bin', 0755);
-        // Install our custom php.ini
-        $this->fs->copy(__DIR__ . '/../../template/php.ini', '.bref/output/.bref/php.ini');
+        // Check if binary package provides a php.ini
+        $phpIniFile = __DIR__ . '/../../template/php.ini';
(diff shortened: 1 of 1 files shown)


t-geindre:
closing, see https://github.com/mnapoli/bref/issues/26#issuecomment-398688736

PR #29, opened 2018-06-20
Dynamic extensions loading

### LINKED ISSUE #26: Load extension dynamically

t-geindre:
This add PHP configuration and dynamic extensions loading:

```yaml
php:
    configuration:
        max_execution_time: 300
    extensions:
        - redis
        - mongodb
```
I couldn't find any library which can handle the several usages of the same key in an INI file:

```ini
extension=redis.so
extension=mongodb.so
```
(even `parse_ini_file()` only gets the last value)

I'm not very proud of the part of this code which generate the INI content, better ideas are welcomed.

fix #26

### DIFF
--- a/bin/php/Dockerfile
+++ b/bin/php/Dockerfile
@@ -17,7 +17,9 @@ RUN yum --releasever=2017.03 install \
     openssl-devel \
     libpng-devel \
     libjpeg-devel \
-    curl-devel -y
+    curl-devel \
+    findutils \
+    php-pear -y
 
 RUN curl -sL https://github.com/php/php-src/archive/$PHP_VERSION.tar.gz | tar -zxv
 
@@ -54,3 +56,5 @@ RUN ./configure \
 RUN make -j 5
 
 RUN make install
+
+RUN pecl install mongodb redis
--- a/bin/php/Dockerfile
+++ b/bin/php/Dockerfile
@@ -18,6 +18,7 @@ RUN yum --releasever=2017.03 install \
     libpng-devel \
     libjpeg-devel \
     curl-devel \
+    # The pecl command uses find
     findutils \
     php-pear -y
 
--- a/bin/php/build.sh
+++ b/bin/php/build.sh
@@ -22,8 +22,7 @@ container=$(docker create php-build)
 
 docker -D cp $container:/php-src-$PHP_VERSION_GIT_BRANCH/sapi/cli/php .
 mkdir -p ext
-docker -D cp $container:/usr/local/lib/php/extensions/no-debug-non-zts-20170718/opcache.a ext/opcache.a
-docker -D cp $container:/usr/local/lib/php/extensions/no-debug-non-zts-20170718/opcache.so ext/opcache.so
+docker -D cp $container:/usr/local/lib/php/extensions/no-debug-non-zts-20170718/. ext/
 docker rm $container
 
 tar czf $PHP_VERSION_GIT_BRANCH.tar.gz php ext
--- a/src/Console/Deployer.php
+++ b/src/Console/Deployer.php
@@ -155,7 +155,14 @@ class Deployer
         // Set correct permissions on the file
         $this->fs->chmod('.bref/output/.bref/bin', 0755);
         // Install our custom php.ini
-        $this->fs->copy(__DIR__ . '/../../template/php.ini', '.bref/output/.bref/php.ini');
+        $this->fs->copy(__DIR__ . '/../../template/php.ini', $phpConfigFile = '.bref/output/.bref/php.ini');
+        // Inject additional config in php.ini
+        $this->injectPhpConfig(
+            $phpConfigFile,
+            $projectConfig['php']['configuration'] ?? [],
+            $projectConfig['php']['extensions'] ?? []
+        );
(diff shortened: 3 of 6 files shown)


bin/php/Dockerfile
```diff
@@ -17,7 +17,9 @@ RUN yum --releasever=2017.03 install \
     openssl-devel \
     libpng-devel \
     libjpeg-devel \
-    curl-devel -y
+    curl-devel \
+    findutils \
```
t-geindre: The `pecl` uses the `find` command, so we need to install it first.

bin/php/build.sh
```diff
@@ -22,8 +22,7 @@ container=$(docker create php-build)
 
 docker -D cp $container:/php-src-$PHP_VERSION_GIT_BRANCH/sapi/cli/php .
 mkdir -p ext
-docker -D cp $container:/usr/local/lib/php/extensions/no-debug-non-zts-20170718/opcache.a ext/opcache.a
-docker -D cp $container:/usr/local/lib/php/extensions/no-debug-non-zts-20170718/opcache.so ext/opcache.so
+docker -D cp $container:/usr/local/lib/php/extensions/no-debug-non-zts-20170718/. ext/
```
mnapoli: 👌 so much simpler!

bin/php/Dockerfile
mnapoli: 👍 could you put that in a comment in the code? (that way the explanation is not lost once this is merged)

src/Console/Deployer.php
```diff
@@ -241,4 +247,25 @@ private function copyProjectToOutputDirectory() : void
...
+            $value = $key . '=' .$value;
+        });
+
+        $extensions = array_map(function ($extension) {
+            if (!$this->fs->exists('.bref/output/.bref/bin/ext/' . $extension . '.so')) {
+                throw new \Exception('PHP Extension "'. $extension . '" does not exist');
+            }
+
+            return 'extension=' . $extension . '.so';
+        }, $extensions);
```
mnapoli: > I couldn't find any library which can handle the several usages of the same key in an INI file:

Is that supposed to be done with sections? E.g. here is how it's done in my config files on my system:

```ini
[opcache]
zend_extension=/usr/local/opt/php/lib/php/20170718/opcache.so
```

Maybe we should put each extension with a section before? (the section could have the same name of the extension)

Also have you tried if it works with both redis and mongodb enabled?

As for PHP configuration this is different: https://stackoverflow.com/a/4651918/245552 The last value is taken, so that's fine for us because that means users can override values in the default `php.ini`.

src/Console/Deployer.php
```diff
@@ -241,4 +247,25 @@ private function copyProjectToOutputDirectory() : void
...
+            $value = $key . '=' .$value;
+        });
+
+        $extensions = array_map(function ($extension) {
+            if (!$this->fs->exists('.bref/output/.bref/bin/ext/' . $extension . '.so')) {
+                throw new \Exception("The PHP extension '$extension' is not available yet in Bref, please open an issue or a pull request on GitHub to add that extension");
+            }
+
+            return 'extension=' . $extension . '.so';
+        }, $extensions);
```
mnapoli: > I couldn't find any library which can handle the several usages of the same key in an INI file:

Is that supposed to be done with sections? E.g. here is how it's done in my config files on my system:

```ini
[opcache]
zend_extension=/usr/local/opt/php/lib/php/20170718/opcache.so
```

Maybe we should put each extension with a section before? (the section could have the same name of the extension)

Also have you tried if it works with both redis and mongodb enabled?

As for PHP configuration this is different: https://stackoverflow.com/a/4651918/245552 The last value is taken, so that's fine for us because that means users can override values in the default `php.ini`.

t-geindre:
I used https://github.com/matomo-org/component-ini to generate `php.ini`. Here is an example of generated file:

```ini
[flags]
opcache.enable = 1
opcache.enable_cli = 1
opcache.file_cache = "/tmp/.bref/opcache"
opcache.file_cache_only = 1
opcache.validate_permission = 0
opcache.validate_timestamps = 0
opcache.memory_consumption = 128
opcache.max_accelerated_files = 10000
extension_dir = "/var/task/.bref/bin/ext"
zend_extension = "opcache.so"

[redis]
extension = "redis.so"
```
> Also have you tried if it works with both redis and mongodb enabled?

I've tried on a real lambda, a `phpinfo()` shows that those two extensions are correctly loaded.

mnapoli:
Really nice solution with the INI!

Thanks!

PR #35, opened 2018-06-25
Add --stage option to deploy command

### LINKED ISSUE #30: Add --stage option on deploy command
`serverless deploy` command has a `--stage` option.

Adding this option to `bref deploy` and forward its value to `serverless deploy` could be useful to deploy lambdas on different stages.

t-geindre:
fix #30

### DIFF
--- a/src/Console/Deployer.php
+++ b/src/Console/Deployer.php
@@ -78,7 +78,7 @@ class Deployer
             $progress->display();
             $serverlessCommand = 'serverless deploy';
             if (null !== $stage) {
-                $serverlessCommand .= ' --stage ' . $stage;
+                $serverlessCommand .= ' --stage ' . escapeshellarg($stage);
             }
             $process = new Process($serverlessCommand, '.bref/output');
             $process->setTimeout(null);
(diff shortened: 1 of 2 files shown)


src/Console/Deployer.php
```diff
@@ -76,7 +76,11 @@ public function deploy(SymfonyStyle $io, bool $dryRun) : void
         if (!$dryRun) {
             $progress->setMessage('Uploading the lambda');
             $progress->display();
-            $process = new Process('serverless deploy', '.bref/output');
+            $serverlessCommand = 'serverless deploy';
+            if (null !== $stage) {
+                $serverlessCommand .= ' --stage ' . $stage;
```
mnapoli: We could use `escapeshellarg()` just in case? (I don't know if there can be special characters in stages but it doesn't cost much to use it)

mnapoli:
That looks good, I've opened #37 to avoid blocking this pull request waiting for the documentation (I'll probably address it myself later). I've added a minor comment and then it's good to merge.

t-geindre:
Yes I thoutgh about documenting this new feature, but as you said in #37, the README is becoming a bit crowded.

PR #36, opened 2018-06-25
Remove unused extensions at build time

### LINKED ISSUE #32: Remove unused PHP extensions
Since we know which extension is activated, shouldn't we remove those which are not ?

This way, the generated archive would be lighter ?

t-geindre:
As said in #32, the idea is to make the lightest package possible.

fix #32

### DIFF
--- a/src/Console/Deployer.php
+++ b/src/Console/Deployer.php
@@ -164,15 +164,7 @@ class Deployer
             $projectConfig['php']['extensions'] ?? []
         );
         // Remove unused extensions
-        foreach (glob('.bref/output/.bref/bin/ext/*.so') as $extensionFile) {
-            if ($extensionFile === '.bref/output/.bref/bin/ext/opcache.so') {
-                continue;
-            }
-            $extensionName = substr($extensionFile, strrpos($extensionFile, '/') + 1, -3);
-            if (!array_key_exists($extensionName, $phpConfig)) {
-                $this->fs->remove($extensionFile);
-            }
-        }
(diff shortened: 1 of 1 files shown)


src/Console/Deployer.php
```diff
@@ -157,12 +157,22 @@ private function generateArchive(SymfonyStyle $io, ProgressBar $progress) : void
...
             '.bref/output/.bref/php.ini',
             $projectConfig['php']['configuration'] ?? [],
             $projectConfig['php']['extensions'] ?? []
         );
+        // Remove unused extensions
+        foreach (glob('.bref/output/.bref/bin/ext/*.so') as $extensionFile) {
+            if ($extensionFile === '.bref/output/.bref/bin/ext/opcache.so') {
+                continue;
+            }
+            $extensionName = substr($extensionFile, strrpos($extensionFile, '/') + 1, -3);
```
mnapoli: This could maybe simplified using `basename()`: https://3v4l.org/hMOhU

src/Console/Deployer.php
```diff
@@ -157,12 +157,22 @@ private function generateArchive(SymfonyStyle $io, ProgressBar $progress) : void
...
+            if (!array_key_exists($extensionName, $phpConfig)) {
+                $this->fs->remove($extensionFile);
+            }
+        }
```
mnapoli: Like the other parts in this file could you move this logic to a private function? (e.g. `$this->removeUnusedExtensions($phpConfig)`)

PR #41, opened 2018-07-09
Add note regarding translations directory

nealio82:
When using symfony/website-skeleton, an exception is thrown unless the `translations` dir is included in `serverless.yml`


mnapoli:
That's perfect thanks!

PR #42, opened 2018-07-15
Add note regarding Twig cache

nealio82:
Twig sometimes needs to perform write operations on the cache dir.

Rather than setting the entire application cache to `/tmp/cache` and losing the benefit of the pre-compiled config cache as well as the Twig cache, the preferred solution should be to disable Twig caching where possible.

### DIFF
--- a/docs/Symfony.md
+++ b/docs/Symfony.md
@@ -71,7 +71,6 @@ package:
     - 'translations/**'
 ```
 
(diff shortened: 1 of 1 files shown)


mnapoli:
Awesome wording thank you! This solution isn't ideal but it's better than nothing.

PR #43, opened 2018-07-23
Add INTL extension

t-geindre:
This PR isn't meant to be merged yet, but just to open a discussion.

Adding the INTL extension to PHP makes it to depend on ICU libraries. So I added a `lib` directory to the tar archive where the ICU's library files are stored. Then, in `handler.js`, I added this `lib` directory to `LD_LIBRARY_PATH`. This is working fine.

The issue here, IMO, is that it adds 4 files for about 25MB extra load:
```bash
$ du libicu* -h 
20M     libicudata.so.50
2,0M    libicui18n.so.50
52K     libicuio.so.50
1,5M    libicuuc.so.50
```

@mnapoli what do you think?

This extension can be replaced by [symfony/polyfill-intl-icu](https://github.com/symfony/polyfill-intl-icu) and [symfony/polyfill-intl-normalizer](https://github.com/symfony/polyfill-intl-normalizer) but not all classes and functions are provided by those packages (eg [Transliterator](http://php.net/transliterator)).

### DIFF
--- a/bin/php/Dockerfile
+++ b/bin/php/Dockerfile
@@ -20,7 +20,12 @@ RUN yum --releasever=2017.03 install \
     curl-devel \
     # The pecl command uses find
     findutils \
-    php-pear -y > /dev/null
+    php-pear \
+    # C++ & ICU are required by INTL
+    libicu \
+    libicu-devel \
+    c++ \
+    gcc-c++ -y > /dev/null
 
 RUN curl -sL https://github.com/php/php-src/archive/$PHP_VERSION.tar.gz | tar -zxv
 
@@ -46,6 +51,7 @@ RUN ./configure \
     --enable-opcache-file \
     --enable-soap \
     --enable-zip \
+    --enable-intl \
     --with-curl \
     --with-openssl \
     --with-zlib \
--- a/bin/php/build.sh
+++ b/bin/php/build.sh
@@ -31,12 +31,19 @@ docker -D cp $container:/php-src-php-$PHP_VERSION/sapi/cli/php .
 mkdir -p ext
 docker -D cp $container:/usr/local/lib/php/extensions/no-debug-non-zts-20170718/. ext/
 
+# Fetch ICU libraries required by INTL extension
+mkdir -p lib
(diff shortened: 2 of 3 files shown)


mnapoli:
This is great stuff! I think Bref should provide an environment as close as what we can get on any other host, so having the `intl` extension is very good. Using polyfills is not ideal, I guess it will have an impact on performances too, and it means users have to be aware of the limitations which is not great.

So I think we should provide the `intl` extension. But *ideally* (i.e. not necessarily in this PR) we could have a guide to help reduce the size of the lambda (because it's easy to hit the size limit) and we could make `intl` optional.

Regarding the 25mb size, [here are the AWS limits](https://docs.aws.amazon.com/lambda/latest/dg/limits.html):

- 50mb compressed (zip file)
- 250mb uncompressed

25mb is a big share. Do you know how much it is when compressed?

t-geindre:
> we could make intl optional.

I tried to do so using the PECL version of INTL. But, as stated [here](https://pecl.php.net/package/intl), the PECL version is not maintained anymore.

I don't know any other solution to provide INTL as a separate extension file. But, I'm sure it's possible since OS packages managers provide a dedicated package for this extension (eg `apt-get install php-intl`).

> 25mb is a big share. Do you know how much it is when compressed?

| Archive        | Without INTL           | With INTL  |
| ------------- |:-------------:| -----:|
| php-7.2.5.tar.gz      | 17M | 28M |
| Lambda Zip on our projet      | 34.4M |  45.9M |

t-geindre:
[This](http://php.net/manual/fr/install.pecl.phpize.php) could be the solution.

t-geindre:
Ok so I managed to compile INTL extension as a separate `intl.so` file using the procedure decribed in the [documentation](http://php.net/manual/fr/install.pecl.phpize.php).

We still need to embed the ICU's library files if this extension is enabled.

I think we could add a `dependencies.yml` file to the php archive as follow:

```yaml
extensions:
    intl:
        - libicudata.so.50
        - libicui18n.so.50
        - libicuio.so.50
        - libicuuc.so.50
```

This way, the package building process could remove any library which is not required.

mnapoli:
Nice! That could be a good solution indeed!

t-geindre:
See #44

PR #44, opened 2018-07-23
Add INTL extension

t-geindre:
Following up #43, here is a new PR to add INTL as an optional extension.

What's been done:
 * Add `intl.so` build
 * Fetch ICU's libraries and add it to tar
 * According to enabled extensions, the deployer now remove useless libraries
 * Doc update

Since `max_execution_time` has no effect in CLI, I've also updated the PHP configuration example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,16 +252,17 @@ You can also define `php.ini` configuration flags and activate extensions:
 ```yaml
 php:
     configuration:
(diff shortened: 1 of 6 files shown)


mnapoli:
This is awesome thanks, I'll review that tonight!

PR #53, opened 2018-08-03
Add missing stage option

### LINKED ISSUE #52: Add --stage option to other commands
We can deploy on a specific stage using the `--stage=` option with `deploy` command.

This options should be available on other commands:
- `cli`
- `info`
- `remove`
- `logs`
- `invoke`

t-geindre:
fix #52 

It turns out I need this option sooner than I thought, so here is a PR!

This adds the `--stage` option on the following commands:
- `cli`
- `info`
- `remove`
- `logs`
- `invoke`

I haven't updated the documentation since those commands are self documented when running `vendor/bin/bref`.

### DIFF
--- a/bref
+++ b/bref
@@ -79,8 +79,9 @@ $app->command(
 /**
  * Run a CLI command in the remote environment.
  */
-$app->command('cli [arguments]*', function (array $arguments, SymfonyStyle $io) {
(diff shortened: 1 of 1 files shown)


mnapoli:
Awesome thanks!

PR #58, opened 2018-08-28
 Add multipart/form-data support

vdechenaux:
Hi,

I saw there is no multipart/form-data support, are you interested with that ?

--

There is a detail, if you want to handle binary files you have to do this in your API Gateway settings:
image

You can also use a plugin to let serverless framework configure it for you, but I did not tested it: https://github.com/maciejtreder/serverless-apigw-binary

If you want to merge this PR, maybe we should add this to the readme ?

### DIFF
--- a/tests/Bridge/Psr7/RequestFactoryTest.php
+++ b/tests/Bridge/Psr7/RequestFactoryTest.php
@@ -5,6 +5,7 @@ namespace Bref\Test\Bridge\Psr7;
 
 use Bref\Bridge\Psr7\RequestFactory;
 use PHPUnit\Framework\TestCase;
+use Psr\Http\Message\UploadedFileInterface;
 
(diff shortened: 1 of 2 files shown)


tests/Bridge/Psr7/RequestFactoryTest.php
```diff
@@ -97,17 +98,27 @@ public function test POST JSON body is not parsed()
...
+bar\r
+--testBoundary\r
+Content-Disposition: form-data; name=\"bim\"\r
+\r
+baz\r
+--testBoundary--\r
+",
         ]);
         self::assertEquals('POST', $request->getMethod());
-        self::assertNull(null, $request->getParsedBody());
```
vdechenaux: It was always `null` :stuck_out_tongue_winking_eye:

Issue #59, opened 2018-09-03
RFC: Session handling when using Symfony framework


nealio82:
Obviously, with Lambda being totally ephemeral, we need to store sessions in some external data store when performing functions which require storing state between requests (such as logging in to  an admin)

One such way is to use the Symfony `PdoSessionHandler` https://symfony.com/doc/current/doctrine/pdo_session_storage.html

When testing locally using the `php -S 127.0.0.1:8000 bref.php` command, this works fine. However, running in the Lambda environment the session doesn't work.

Although the session appears to be created, for some reason the response object (`$symfonyResponse = $this->httpKernel->handle($symfonyRequest);`) in `/src/Bridge/Symfony/SymfonyAdapter.php` doesn't contain the session cookie / `Set-Cookie` header, and so the client never sends the `PHPSESSID` cookie back on the next request.

Additionally, when using Symfony, the framework has its own interface for accessing the its own internal session handling, meaning you can't use `session_id()` or other `session_...` methods for interacting with the session handling.

Adding the cookie header is easily fixed by adding the following code to `SymfonyAdapter.php`:
```
$symfonyResponse->headers->setCookie(
    new Cookie(
        session_name(),
        $this->httpKernel->getContainer()->get('session')->getId()
    )
);
```
However, during the login cycle, requests are redirected using `301` and `302` redirects, and unless the session id is passed to Symfony, a new session id is generated each time and thus is never re-fetched from the data store. This means that the user record is stored authenticated in the database, but as the client has received a new `PHPSESSID` cookie, the session is never retrieved back from the data store and the user is not able to access the logged-in content. 

This can also be fixed by adding the following to `SymfonyAdapter.php`:
```
if (!is_null($symfonyRequest->cookies->get(session_name()))) {
    $this->httpKernel->getContainer()->get('session')->setId(
        $symfonyRequest->cookies->get(session_name())
    );
}
```

However, this means that in order to work with Symfony's sessions, the `handle` method now looks like this:
```
public function handle(ServerRequestInterface $request): ResponseInterface
{
    $httpFoundationFactory = new HttpFoundationFactory;

    $symfonyRequest = $httpFoundationFactory->createRequest($request);

    if (!is_null($symfonyRequest->cookies->get(session_name()))) {
        $this->httpKernel->getContainer()->get('session')->setId(
            $symfonyRequest->cookies->get(session_name())
        );
    }

    $symfonyResponse = $this->httpKernel->handle($symfonyRequest);

[...]

PR #60, opened 2018-09-11
Add session handling into `SymfonyAdapter`

nealio82:
As per #59, some additional injection is needed on the `Request` and `Response` objects in order to handle the `PHPSESSID` cookie sent in the request headers, or to add the `Set-Cookie` header to the response when using the Symfony framework with Lambda / Bref

### DIFF
--- a/src/Bridge/Symfony/SymfonyAdapter.php
+++ b/src/Bridge/Symfony/SymfonyAdapter.php
@@ -71,7 +71,14 @@ class SymfonyAdapter implements RequestHandlerInterface
         $symfonyResponse->headers->setCookie(
             new Cookie(
                 session_name(),
-                $this->httpKernel->getContainer()->get('session')->getId()
+                $this->httpKernel->getContainer()->get('session')->getId(),
+                0,
+                "/",
+                null,
+                false,
+                true,
+                false,
+                Cookie::SAMESITE_LAX
             )
         );
     }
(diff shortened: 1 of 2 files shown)


src/Bridge/Symfony/SymfonyAdapter.php
```diff
@@ -43,4 +50,29 @@ public function handle(ServerRequestInterface $request) : ResponseInterface
...
+            );
+        }
+    }
+
+    /**
+     * @param $symfonyResponse
+     */
+    private function addSessionCookieToResponse($symfonyResponse): void
+    {
+        $symfonyResponse->headers->setCookie(
```
staabm: will this cookie contain a http-only flag (or other security measures)?

nealio82: @staabm yes, it will. i've updated the unit test to reflect this. is there anything else you feel i need to be added for security?

staabm: SameSite Lax oe Strict ?

nealio82: the Symfony default for the `Cookie` class is `httponly=true` and `SameSite` is null. I've added `Cookie::SAMESITE_LAX` into the adapter but I'm not 100% certain we want to override the defaults too much..?

mnapoli: @staabm is it all good for you here? I haven't studied cookies in enough details to be of more help. (the rest of the PR is good to merge for me)

mnapoli:
Thanks both!

PR #62, opened 2018-09-18
add pre-build hook to deployment

nealio82:
The Symfony best practices suggest to run `yarn encore production` during deployment (https://symfony.com/doc/current/frontend/encore/simple-example.html). However, when adding this as a build hook in `.bref.yml` the process blows up with:

```
In Process.php line 223:

  The command "yarn encore production" failed.

  Exit Code: 1(General error)

  Working directory: .bref/output

  Output:
  ================
  yarn run v1.9.4
  $ ~/Projects/bref-sf4/.bref/output/node_modules/.bin/encore production
  info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.


  Error Output:
  ================
  internal/modules/cjs/loader.js:583
      throw err;
      ^

  Error: Cannot find module '../lib/config/parse-runtime'
      at Function.Module._resolveFilename (internal/modules/cjs/loader.js:581:15)
```

Running the command in the current working directory *before* the build artefact is created runs smoothly. Therefore the ability to add it as a `pre-build` hook in the deployment process allows us to automate the generation of compiled assets during the build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,6 +295,11 @@ You can execute additional scripts by using a *build hook*. Those can be defined
 
 ```yaml
 hooks:
+    # pre-build hooks are run *before* the deployment package is created (in the current working directory)
+    pre-build:
+        - 'yarn encore production'
+        
+    # build hooks are run on the deployment package *during* its creation (in .bref/)  
     build:
         - 'npm install'
 ```
--- a/src/Console/Deployer.php
+++ b/src/Console/Deployer.php
@@ -126,6 +126,17 @@ class Deployer
         }
         $progress->advance();
 
+        // Run pre-build hooks defined in .bref.yml
+        $progress->setMessage('Running pre-build hooks');
+        $progress->display();
+        $buildHooks = $projectConfig['hooks']['pre-build'] ?? [];
+        foreach ($buildHooks as $buildHook) {
+            $progress->setMessage('Running pre-build hook: ' . $buildHook);
+            $progress->display();
+            $this->runInWorkingDirectory($buildHook);
+        }
+        $progress->advance();
+
         $progress->setMessage('Building the project in the `.bref/output` directory');
         $progress->display();
         $this->copyProjectToOutputDirectory();
@@ -195,6 +206,13 @@ class Deployer
             $this->runLocally($buildHook);
         }
         $progress->advance();
+    }
+
+    private function runInWorkingDirectory(string $command) : void
+    {
+        $process = new Process($command);
+        $process->setTimeout(null);
+        $process->mustRun();
     }
 
     private function runLocally(string $command) : void


nealio82:
note: The original issue persists even when explcitly calling `yarn install` immediately before calling `yarn encore production` in the `build` hooks config

```
  build:
    - 'yarn install'
    - 'yarn encore production'
```

nealio82:
A colleague and I tried a few different ways to get yarn running in the temp directory but so far have been unable to. We're on a tight schedule atm but I can try again today.

> (e.g. here to avoid having all the production artifacts in the dev environment).
I see your point. Our thinking was that we could have the option to add a pre-build hook with the understanding that doing so would create the built JS in the working directory and then copy it to the deployment package during the execution.

> Why do we need to add a pre-build hook?
Solving the yarn problem would allow us to use `/vendor/bin/bref deploy` as our single source of truth for building and deploying everything, saving us from needing to remember to run `yarn encore production` each time, or having to create a `deploy.sh` script which would do the packaging and then deployment

nealio82:
From digging in a little deeper, it appears that the following _might_ work:

```
hooks:
  build:
    - 'rm -rf node_modules/*'
    - 'npm install'
    - 'yarn encore production'
    - 'APP_ENV=prod php bin/console cache:clear --no-debug --no-warmup'
    - 'APP_ENV=prod php bin/console cache:warmup'
```

the `.bin/encore` package has relative paths inside it to `@symfony/webpack-encore/lib/config/parse-runtime.js`, which appears to be copied into the build package but (i'm assuming) not any symlinks. According to Stack Overflow, `yarn install` doesn't create symlinks either, which is why adding `yarn install` before `yarn encore production` doesn't work.

However, simply adding `npm install` also fails because of existing files:
```
 npm ERR! path ~/Projects/bref-sf4/.bref/output/node_modules/sockjs/node_module
  s/.bin/uuid
  npm ERR! code EEXIST
  npm ERR! Refusing to delete ~/Projects/bref-sf4/.bref/output/node_modules/sock
  js/node_modules/.bin/uuid: is outside ~/Projects/bref-sf4/.bref/output/node_mo
  dules/sockjs/node_modules/uuid and not a link
  npm ERR! File exists: ~/Projects/bref-sf4/.bref/output/node_modules/sockjs/nod
  e_modules/.bin/uuid
  npm ERR! Move it away, and try again.
```

Thus the workaround seems to be to delete `node_modules`, run `npm install`, and then run `yarn encore production` afterwards.

However I'll try it out a few more times before closing this PR

nealio82:
The JS is served from S3, as per https://symfony.com/doc/current/frontend/encore/cdn.html

We're using the `serverless-s3-assets` plugin for the serverless framework (included in `serverless.yml`), which takes the compiled JS and uploads it to S3 during the deployment. As the serverless framework upload process is initated by `bref deploy`, it makes sense to add the compilation to the build steps as well.

In short, we're utilising `bref deploy` to initiate updating the whole application stack (RDS, VPC, subnets, S3, etc)

mnapoli:
Makes sense! I'm rewriting my blog with bref and I'll be facing exactly that issue with uploading assets so I'll try the same plugin!

nealio82:
FTR, my colleague @eddmann (who knows a lot more node-y stuff than I do) came up with this solution in the end, which is much nicer than the `rm -rf node_modules && npm install` method:

```
hooks:
  build:
    - 'yarn add @symfony/webpack-encore'
    - 'yarn encore production'
    - 'APP_ENV=prod php bin/console cache:clear --no-debug --no-warmup'
    - 'APP_ENV=prod php bin/console cache:warmup'
```

PR #65, opened 2018-09-21
Fix empty session value in Symfony adapter

eddmann:
Sorry about this change.
The mock session handler used within the tests takes the NULL value (if no cookie is present) and sets it as the id. This allows the subsequent check upon response to function correctly.
However, when using the native `session_id` function in production, the NULL value is not returned and `session_id` returns an empty string.
So as to meet these requirements we should instead replace NULL with an empty string to signify no session being present.


mnapoli:
No worries, thanks for the fix!

PR #66, opened 2018-09-30
add support for HTTP_HOST server var passed in from Lambda event

nealio82:
The `$_SERVER['HTTP_HOST']` global isn't available within Bref for obvious reasons, however it is useful in many applications to have access to the hostname the script is running on.

The PSR-7 `Request` object allows us to create `$server` variables but they need to be explicitly converted from the Lambda event in order to be used.

When using Symfony's router to generate absolute URLs, the URLs will all point to `localhost` unless `HTTP_HOST` is set.

### DIFF
--- a/src/Bridge/Psr7/RequestFactory.php
+++ b/src/Bridge/Psr7/RequestFactory.php
@@ -91,6 +91,10 @@ class RequestFactory
             'REQUEST_URI' => $uri,
         ];
 
+        if (isset($headers['Host'])) {
+            $server['HTTP_HOST'] = $headers['Host'];
+        }
+
         return new ServerRequest(
             $server,
             $files,
--- a/tests/Bridge/Psr7/RequestFactoryTest.php
+++ b/tests/Bridge/Psr7/RequestFactoryTest.php
@@ -267,4 +267,16 @@ RAW
         self::assertEquals('POST', $request->getMethod());
         self::assertEquals(['foo' => 'bar'], $request->getParsedBody());
     }
+
+    public function test HTTP_HOST is set() {
+        $request = RequestFactory::fromLambdaEvent([
+            'headers' => [
+                'Host' => 'www.example.com',
+            ],
+        ]);
+
+        $serverParams = $request->getServerParams();
+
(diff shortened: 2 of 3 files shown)


nealio82:
I'm really not sure why this is failing, because my original changes didn't touch `CliTest.php` or `Process` / deploy script at all 🤔

mnapoli:
It's weird! The build with the lowest dependencies is passing, could it be that the tests are failing on all branches?

nealio82:
Yeah, I'm not sure, although I think it could be. Could it be a minor version upgrade in a dependency that's breaking it?

nealio82:
yes, tests break on master too (managed to replicate locally)

mnapoli:
Mhh I don't understand, I checked out master and pulled the last changes, then ran `composer update` and tests still work fine on my end (MacOS). I need to dig into this :/

nealio82:
In order to replicate I had to do the following:

- Set PHPStorm's interpreter to 7.2
- Allow PHPStorm to download and use its own local `composer.phar` appropriate for the interpreter settings (instead of using my global composer)
- Run `php composer.phar install`
- Run phpunit tests

nealio82:
Although, now, having run `composer install` in the same way again, I have this:

```
Testing started at 09:15 ...
/usr/local/Cellar/php/7.2.9_2/bin/php /Users/neal/www/bref/vendor/phpunit/phpunit/phpunit --no-configuration Bref\Test\CliTest /Users/neal/www/bref/tests/CliTest.php --teamcity
PHPUnit 6.5.13 by Sebastian Bergmann and contributors.



Time: 188 ms, Memory: 4.00MB

OK (1 test, 2 assertions)

Process finished with exit code 0
```

mnapoli:
That's weird, I cannot reproduce any issue locally. #68 passed the tests on Travis too, and now master is green after the merge…

I've restarted the jobs, we'll see. Could you try merging master into your branch if tests continue to fail after this?

mnapoli:
Andddd now it's green. There is one of those shown as failing, I cannot get a URL working so I'll just ignore it and merge. I should have tried that sooner, sorry for the time wasted!

Thanks!

PR #68, opened 2018-10-06
Bref init: check that serverless is installed and AWS configred

vdechenaux:

PR #69, opened 2018-10-07
Bref init: Automatically `git add bref.php serverless.yml` + gitignore /.bref/

vdechenaux:
496a990 is a little extra, related to #68 :speak_no_evil:

### DIFF
--- a/bref
+++ b/bref
@@ -65,14 +65,39 @@ $app->command('init', function (SymfonyStyle $io) {
         $io->writeln('Creating bref.php');
         $fs->copy(__DIR__ . '/template/bref.php', 'bref.php');
     }
-    $io->success([
-        'Project initialized and ready to deploy using `bref deploy`',
-        'If you are using git, you will need to commit the following files:',
-        '- bref.php',
-        '- serverless.yml',
-        // TODO Do this automatically
-        'You can add `/.bref/` to your .gitignore.',
-    ]);
+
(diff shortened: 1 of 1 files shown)


bref
```diff
@@ -31,28 +31,19 @@ $app->command('init', function (SymfonyStyle $io) {
         return 1;
     }
 
-    $configured = true;
     if (getenv('AWS_ACCESS_KEY_ID') === false || getenv('AWS_SECRET_ACCESS_KEY') === false) {
         // Credentials are not configured with environment variables, check aws cli
-        if ($exeFinder->find('aws') !== null) {
```
vdechenaux: This is not needed, `Process` will exits with status 1 if not found. It simplifies the block :)

bref
```diff
@@ -65,14 +56,39 @@ $app->command('init', function (SymfonyStyle $io) {
...
-        'You can add `/.bref/` to your .gitignore.',
-    ]);
+
+    /*
+     * We check if this is a git repository to automatically run some commands:
+     * - git add bref.php serverless.yml
+     * - add /.bref/ to .gitignore
+     */
+    if ((new Process('git rev-parse --is-inside-work-tree'))->run() === 0) {
+        (new Process('git add bref.php serverless.yml'))->run();
```
staabm: this could also be combined into a single process command, with `&&`

PR #70, opened 2018-10-07
Added REQUEST_TIME_FLOAT server var

staabm:
As supported by native php

http://php.net/manual/en/reserved.variables.server.php

> 'REQUEST_TIME_FLOAT'
The timestamp of the start of the request, with microsecond precision


mnapoli:
Thanks, the tests are failing though.

staabm:
thx for the pointer. didn't realize it.

I just looked up https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-mapping-template-reference.html which does not define a microsecond precision timestamp in the requestContext.

should we still add it, since php devs are used to have it available - even if it is somehow simulated by the framework and not aws lambda native?

staabm:
I feel we should add it, so "regular php libs" can be re-used in bref context which expect this key to be existant.

mnapoli:
I agree it makes sense, and it costs very little to have this. Thanks!

PR #73, opened 2018-10-17
Added enriched error handling for lambda non-proxy

krcrawford:
Out of the box, bref works great. I found that I wanted to lean more on non-proxied lambda integrations and take advantage of the error handling in the response integrations.
Adding this technique allowed my api gateway to effectively massage the responses without a simple status code override.
Not sure if this is good for going forward, but it helped me out a bit.
Alternatively, perhaps an eventual allowance for use of a custom `handler.js` template in the same fashion as the current `.bref.yml` override strategy.
Thanks for any consideration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,7 +261,9 @@ If you want to, you can define those 3 handlers in the same application. On exec
 
 The filesystem on lambdas is read-only (except for the `/tmp` folder). You should not try to write application logs to disk.
 
-The easiest solution is to push logs to AWS Cloudwatch (Amazon's solution for logs). Bref (and AWS Lambda) will send to Cloudwatch anything you write on `stdout` (using `echo` for example) or `stderr`. If you are using Monolog this means you will need to configure Monolog to write to the output (contribution welcome: clarify with an example).
+The easiest solution is to push logs to AWS Cloudwatch (Amazon's solution for logs). Bref (and AWS Lambda) will send to Cloudwatch anything you write on `stdout` (using `echo` for example) or `stderr`. If you are using Monolog this means you will need to configure Monolog to write to the output.
+For Symfony applications, see a Monolog usage example in [deploying Symfony applications](docs/Symfony.md).
+Additional Monolog strategy contributions welcome. Please clarify with examples.
 
 If you have more specific needs you can of course push logs to anything, for example Logstash, Papertrail, Loggly, etc.
 
--- a/README.md
+++ b/README.md
@@ -261,9 +261,7 @@ If you want to, you can define those 3 handlers in the same application. On exec
 
 The filesystem on lambdas is read-only (except for the `/tmp` folder). You should not try to write application logs to disk.
 
-The easiest solution is to push logs to AWS Cloudwatch (Amazon's solution for logs). Bref (and AWS Lambda) will send to Cloudwatch anything you write on `stdout` (using `echo` for example) or `stderr`. If you are using Monolog this means you will need to configure Monolog to write to the output.
-For Symfony applications, see a Monolog usage example in [deploying Symfony applications](docs/Symfony.md).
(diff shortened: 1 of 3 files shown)


mnapoli:
Thank you for the PR that's interesting.

> Adding this technique allowed my api gateway to effectively massage the responses without a simple status code override.

I haven't use this, could you explain what you can do with API Gateway? Because currently the HTTP status code is correctly forwarded to the client so I'm not sure why you need this change.

krcrawford:
As I currently understand the application, the AWS API Gateway expects the lambda error responses in the format:
```
{
    "errorMessage": "<replaceable>string</replaceable>",
    "errorType": "<replaceable>string</replaceable>",
    "stackTrace": [
        "<replaceable>string</replaceable>",
        ...
      ]
}
```
### TL;DR
It's not just returing the HTTP status code, but the content, json keys and values, seems to be very relevant and specific
Seems that only a `callback(Error)` from node will provide the necessary failure subject required to regex match for dynamic integration response types
 
[Handle Lambda Errors in API Gateway](https://docs.aws.amazon.com/apigateway/latest/developerguide/handle-errors-in-lambda-integration.html)
If and when this is returned, the Lambda Regex will use the `errorMessage` attribute value as the regex subject. See [selectionPattern](https://docs.aws.amazon.com/apigateway/api-reference/resource/integration-response/#selectionPattern) here. That being said, it seems that additionally, it may require the specific header: `X-Amzn-Function-Error`, which seemingly can't be passed by the lambda without intentionally triggering an `Error` within the handler. I have also attempted to provide this header in my Symfony application with no noticeable behavior change, but at that point its a nested value in the `headers` attribute of Bref's `LambdaResponse`, and not a "super" header built into the Lambda response.

Other notables:
>However, make sure that the error response does not contain any newline (\n) character in such cases

So clean up, `preg_replace('/\s+/', ' ', $jsonData)` or something to that effect, your PHP json responses before you output to your lambda.
Based on the above, I'm going to remove the `replace` operation from the PR incase it still has momentum.

krcrawford:
Another alternative might be to introduce an error handler in `LamdaResponse::toJson`:
...perhaps an event listener
```
    ...
    if (!($this->statusCode >= 200 && $this->statusCode < 300)) {
        return json_encode([
            'erorMessage' => '<some-message>',
            'errorType' => 'Error',
            'stackTrace' => [
                ... aggregate symfony exception stack trace...
            ],
        ]);
    }
    ...
```
Which may or may not work well, introducing an observer into bref, but then again, this all seems to require the specific Lambda error header mentioned in my previous comment.
Nevermind, this would likely not work.

mnapoli:
Thanks for the link. What I don't understand is that a 404 response from PHP (e.g. Symfony) shows correctly as a 404 response in the browser. API Gateway did not turn that into a 200 response.

So I don't understand why we need this 🤔 	

Or maybe we need this for 500 HTTP responses only? (i.e. actual application errors)

krcrawford:
Is it possible you were using the proxy integration? I believe that proxy integration ignores any response integration processing and simple moves to the model application, thereby just passing your failure through?

mnapoli:
Oh right yes I always use the proxy integration.

Is there a reason not to use that integration? (I thought it was the "old" way which was more complex)

krcrawford:
I don't see any reason to _not_ use the proxy integration, personally. The scenario in this PR is predicated upon using the non-proxy integration and providing that flexibility if the consumer sees fit.
The non-proxy is certainly more complex.
In any case, thanks for taking the time. It's understandable if you'd rather not support non-proxied integration, but this should help anyone working in that context if you think bref should have it.
You've done some great work with bref, so thanks!

mnapoli:
No worries, I'll close this for now to support only the proxy integration for now 👍

PR #74, opened 2018-10-18
adding path option in InvokeCommand

MarcAtEnopTeam:
## Why

Because it is nice to get event load from files on local invoke

### How

Update InvokeCommand with path option

### DIFF
--- a/src/Cli/InvokeCommand.php
+++ b/src/Cli/InvokeCommand.php
@@ -58,7 +58,7 @@ class InvokeCommand extends Command
                 throw new \RuntimeException('The `--path` option reference an invalid file path: ' . $option);
             }
             $fileContent = file_get_contents($path);
-            if (!$fileContent){
+            if (!$fileContent) {
                 throw new \RuntimeException('Unable to get file content:' . $option);
             }
             $event = json_decode($fileContent, true);
--- a/src/Cli/InvokeCommand.php
+++ b/src/Cli/InvokeCommand.php
@@ -43,33 +43,27 @@ class InvokeCommand extends Command
         $simpleHandler = ($this->invokerLocator)();
 
         $event = [];
+
         if ($option = $input->getOption('event')) {
-            $event = json_decode($option, true);
-            if ($event === null) {
+            if (null === $event = json_decode($option, true)) {
                 throw new \RuntimeException('The `--event` option provided contains invalid JSON: ' . $option);
             }
         }
+
         if ($option = $input->getOption('path')) {
-            $path = $option;
-            if (0 !== strpos($option, '/') && false === (bool)preg_match('#^[a-z]:#i', $path)) {
-                $path = realpath(getcwd() . DIRECTORY_SEPARATOR . $option);
-            }
-            if (!$path) {
+            if (!$path = realpath($option)) {
                 throw new \RuntimeException('The `--path` option is an invalid path: ' . $option);
             }
             if (!is_readable($path)) {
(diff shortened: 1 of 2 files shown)


src/Cli/InvokeCommand.php
```diff
@@ -42,10 +43,27 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+                throw new \RuntimeException('The `--event` option provided contains invalid JSON: ' . $option);
+            }
+        }
+        if ($option = $input->getOption('path')) {
+            $path = realpath(getcwd() . DIRECTORY_SEPARATOR . $option);
+            if (!$path) {
+                throw new \RuntimeException('The `--path` option is an invalid path: ' . $option);
+            }
+            if (!is_readable($path)) {
+                throw new \RuntimeException('The `--path` option reference an invalid file path: ' . $option);
```
staabm: or misses permission

mnapoli:
Thanks!

Maybe it could make sense to reuse the `--event` parameter to accept either:

- a JSON string
- or a file path

We can then check if the path exists, or if it can be JSON-decoded, else throw an exception. There is no risk of confusion between the 2.

That would be simpler I guess, no new parameter to learn about. WDYT?

src/Cli/InvokeCommand.php
```diff
@@ -42,12 +43,33 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+        if ($option = $input->getOption('event')) {
+            $event = json_decode($option, true);
             if ($event === null) {
-                throw new \RuntimeException('The `--event` option provided contains invalid JSON: ' . $input->getOption('event'));
+                throw new \RuntimeException('The `--event` option provided contains invalid JSON: ' . $option);
             }
         }
+        if ($option = $input->getOption('path')) {
+            $path = $option;
+            if (0 !== strpos($option, '/')) {
```
staabm: Does this work on windows?

MarcAtEnopTeam: good point

mnapoli:
> The idea behind it, is keeping a compliance-like with the serverless --path parameter which serves the same thing.

Yep ok that sounds fine, sorry for the delay.

src/Cli/InvokeCommand.php
```diff
@@ -42,12 +43,33 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+        if ($option = $input->getOption('event')) {
+            $event = json_decode($option, true);
             if ($event === null) {
-                throw new \RuntimeException('The `--event` option provided contains invalid JSON: ' . $input->getOption('event'));
+                throw new \RuntimeException('The `--event` option provided contains invalid JSON: ' . $option);
             }
         }
+        if ($option = $input->getOption('path')) {
+            $path = $option;
+            if (0 !== strpos($option, '/') && false === (bool)preg_match('#^[a-z]:#i', $path)) {
```
mnapoli: Does that mean that I cannot use an absolute path? If so explain that in a code comment? (it's not obvious on first read)

And is there a reason we want to limit that?

MarcAtEnopTeam: That mean opposite, in case of an absolute, we don't need to do resolve that path , so we are going to just continue the treatment. But after all, doing that look useless. Maybe just realpath is enough ?

```suggestion
           if (!$path = realpath($option)) {
                throw new \RuntimeException('The `--path` option is an invalid path: ' . $option);
            }
```

MarcAtEnopTeam: Is outdated due to changes.

mnapoli:
Thanks!

PR #75, opened 2018-10-19
Monolog symfony example

krcrawford:
In this PR, I've updated some of the docs to include a Monolog usage example for Symfony applications.

### DIFF
--- a/docs/Symfony.md
+++ b/docs/Symfony.md
@@ -87,6 +87,50 @@ public function getLogDir()
 
 The best solution however is not to write log on disks because those are lost. You should use a remote log collector (ELK stack) or a cloud solution like Cloudtrail, Papertrail, Loggly, etc.
 
+While `stdout` and `stderr`echo statements get logged directly to CloudWatch, you can easily use Monolog to write to CloudWatch using `php://stderr`.
+
+```shell
+composer require symfony/monolog-bundle
+```
+
+```yaml
+#config/packages/prod/monolog.yaml
+monolog:
+  handlers:
+    # use a handler name of your choice, but here we are using "stderr"
+    stderr:
+      type: stream
+      path: 'php://stderr'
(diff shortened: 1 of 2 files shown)


krcrawford:
@GromNaN Fair enough. This was really in response to the [README.md](https://github.com/mnapoli/bref/blob/master/README.md) as requested here:
>If you are using Monolog this means you will need to configure Monolog to write to the output (contribution welcome: clarify with an example).

I agree that the minimalist logger would be a leaner solution, as you suggest.

@mnapoli Do you want to continue providing docs that suggest the use of Monolog? I may introduce another PR around removing the Monolog suggestion and suggesting the minimalist logger (unless @GromNaN creates a PR)

docs/Symfony.md
```diff
@@ -87,6 +87,50 @@ public function getLogDir()
 
 The best solution however is not to write log on disks because those are lost. You should use a remote log collector (ELK stack) or a cloud solution like Cloudtrail, Papertrail, Loggly, etc.
 
+While `stdout` and `stderr`echo statements get logged directly to CloudWatch, you can easily use Monolog to write to CloudWatch using `php://stderr`.
```
staabm: Missing space before „echo“

docs/Symfony.md
```diff
@@ -87,6 +87,50 @@ public function getLogDir()
...
+
+```shell
+composer require symfony/monolog-bundle
+```
+
+```yaml
+#config/packages/prod/monolog.yaml
```
staabm:
```suggestion
# config/packages/prod/monolog.yaml
```

mnapoli:
>  I may introduce another PR around removing the Monolog suggestion and suggesting the minimalist logger instead

That sounds even better indeed! If you can do that that's awesome thanks!

mnapoli:
Closing this PR in the meantime.

PR #81, opened 2018-10-31
Refactor the PSR-7 request and lambda response creation code to allow for custom implementations

jvis:
Hi @mnapoli, thanks for your work on Bref. It's really great!

I recently had a use-case where I needed to set `"isBaseEncoded" => true` on the lambda response array and realised there was no easy way to override that code.

This pull request abstracts the PSR-7 request and lambda response creation code to allow for custom implementations (eg. to achieve the `isBaseEncoded` example above).

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -6,8 +6,10 @@ namespace Bref;
 use Bref\Bridge\Psr7\RequestFactory;
 use Bref\Cli\InvokeCommand;
 use Bref\Cli\WelcomeApplication;
-use Bref\Http\LambdaResponse;
+use Bref\Http\LambdaResponseArraySerializer;
 use Bref\Http\WelcomeHandler;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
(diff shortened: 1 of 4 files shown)


mnapoli:
Hi, thank you!

Is there a reason we don't set `"isBaseEncoded" => true` all the time instead?

jvis:
Base64 encoded responses require some additional configuration on the API Gateway side also: https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-payload-encodings.html

Using `isBase64Encoded` all the time would likely add unnecessary complexity, as most use-cases won’t need it.

However, my pull request would allow the developer to use that setting (or any other custom response) if desired.

mnapoli:
Oh OK. I need to take some time to review this, this is quite a big change and it complexifies the whole thing, I'd love to find a simpler solution 🤔 

In any case thanks for the work, I'll look into this!

mnapoli:
Hi! I'll be closing this since with v0.3 HTTP will be supported through a custom runtime via PHP-FPM. Encoding the response in base64 is possible but only when implementing a HTTP runtime manually.

PR #96, opened 2018-11-22
Add support for the PHP extension BCMath

### LINKED ISSUE #95: Request for extension `bcmath`
This is needed for (some?) watchers in Laravel Telescope. (At least the `EventWatcher`).

Still working on getting Telescope usable, so not an urgent request as may not be possible anyway (or at least not for the watchers that need it).

I'm mostly getting Telescope working as a reasonably complex app (along with Nova) to see where the pain points are with Bref.

oppiansteve:
Closes #95

### DIFF
--- a/bin/php/Dockerfile
+++ b/bin/php/Dockerfile
@@ -51,6 +51,7 @@ RUN ./configure \
     --enable-opcache-file \
     --enable-soap \
     --enable-zip \
+    --enable-bcmath \
     --with-curl \
     --with-openssl \
     --with-zlib \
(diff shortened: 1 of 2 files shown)


bin/php/Dockerfile
```diff
@@ -72,6 +73,13 @@ RUN ./configure
 RUN make
 RUN make install
 
+# Compile the bcmath extension because it cannot be installed with pecl
+WORKDIR /php-src-$PHP_VERSION/ext/bcmath
+RUN phpize
+RUN ./configure
+RUN make
+RUN make install
+
```
mnapoli: Thank you for the PR! Do you know why/if it is necessary to compile the extension given there is `--enable-bcmatch` above?

In the documentation it says to use `--enable-bcmatch` and that's it (https://secure.php.net/manual/en/bc.installation.php).

PR #102, opened 2018-12-02
Fix Issue 101

MickaelTH:
Add Timeout on Symphony Progress when untar file that could timeout on slow machine.


mnapoli:
Thanks!

Issue #100, opened 2018-11-30
Run PHP using the official integration API on AWS


mnapoli:
[AWS announced the possibility to use any programming language on Lambda](https://aws.amazon.com/fr/blogs/aws/new-for-aws-lambda-use-any-programming-language-and-share-common-components/). This is awesome! That means a simplification in Bref, (probably) better performances and a more official support of PHP.

Stackery announced they are working on the PHP runtime and this is available in this repository: https://github.com/stackery/php-lambda-layer

The questions are:

- should we use Stackery's runtime?
- do we get better performances (if not maybe we should hold off)?
- do we loose features? (e.g. ability to use some extensions or customize php.ini)

Let's use this issue to track information about all this.

At the moment I have been trying Stackery's PHP layer and here is what I noted:

- the "bootstrap" (the executable file that communicates with AWS and orchestrates the execution) [is written in PHP](https://github.com/stackery/php-lambda-layer/blob/master/bootstrap) which is good (easy to contribute and maintain)
  - would it make sense to have this small bootstrap file as a binary for performances? (this is an optimization that can be done later I guess)
- no tests of the runtime, e.g. I wasn't able to set response headers so don't know if it's supported (I'll be assuming no)
- response headers did not work
- uses [PHP's built-in webserver](http://php.net/manual/en/features.commandline.webserver.php): it's marked as a development server, but here since we don't have concurrent requests on a single Lambda maybe it's not that bad of an idea? Probably worth looking into performances and alternatives though?
- very very little set of extensions installed (for me a simple app was not running because the tokenize functions were not available), especially critical ones like `json`, `intl`, etc.
- supports only PHP 7.1
- marked as not ready for production yet

Update: this runtime does not seem to be made or maintained by PHP developers judging from the discussions in the issues/PR. I don't consider it viable at the moment.

What's interesting is that creating a runtime for AWS is in the end pretty easy. Our [build script](https://github.com/mnapoli/bref/tree/master/bin/php) is almost ready, and more powerful that what can be found there.

I'll be trying out more things, if you have info to share feel free to post it here.

mnapoli:
Update:

- I managed to compile and publish my own PHP 7.2 runtime, with all the extensions currently provided by Bref (much better since PHP 7.1 is becoming EOL)
- Stackery's bootstrap file is very buggy and WIP, this first version is not usable at all, I don't know what is planned there
- I am exploring writing another bootstrap from scratch that does not use/require pcntl and forks, will report when I have more info

[...]

PR #106, opened 2018-12-05
[WIP] Lambda layer for the PHP runtime

mnapoli:
This is related to #100 

The idea is to create a Lambda layer for the PHP runtime.


runtime/publish.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh
+
+set -e
+
+# TODO dynamically change the region
+REGION=us-east-2
+BREF_VERSION=0.3
+PHP_VERSION=7.2.5
```
staabm: any chance to get a more recent php version?

mnapoli: Oh yeah we just have to change that line. However I thought it would be fair to compare to the same version used in Bref currently for the benchmarks.

Let's keep that comment open so that I remember to update this line before merging.

barryvdh:
We should probably keep in mind what the easiest way is for people to add their own php.ini and extensions, without requiring a new build. See https://github.com/stackery/php-lambda-layer/pull/6 and https://github.com/stackery/php-lambda-layer/issues/3
But that also depends on the chosen method of implementation.

mnapoli:
Yes I haven't done that yet but this is planned.

runtime/build.sh
```diff
@@ -0,0 +1,105 @@
...
+# --enable-ftp: because ftp_ssl_connect() needs ftp to be compiled statically (see https://github.com/docker-library/php/issues/236)
+# --enable-mbstring: because otherwise there's no way to get pecl to use it properly (see https://github.com/docker-library/php/issues/195)
+# --enable-opcache-file: allows to use the `opcache.file_cache` option
+# --with-mhash: https://github.com/docker-library/php/issues/439
+./configure \
+    --enable-option-checking=fatal \
+    --enable-static=yes \
+    --enable-shared=no \
+    --disable-cgi \
+    --disable-fpm \
```
barryvdh: This does make it hard to test with FPM, right?

mnapoli: Yes I ported the original build script, I haven't had time to build FPM and do the benchmark yet ;)

barryvdh: No was trying to see if I could create the benchmark for that, but not sure how to build the entire layer myself.

barryvdh: Yeah I tried and did compile it with fpm, but actually have no idea how to set it up further, so gave that up for now :(

runtime/build.sh
```diff
@@ -53,6 +53,7 @@ cd php-src-php-$PHP_VERSION
     --enable-option-checking=fatal \
     --enable-static=yes \
     --enable-shared=no \
+    --with-config-file-path=/opt \
     --disable-cgi \
```
barryvdh: Can you enable cgi?

runtime/build.sh
```diff
@@ -0,0 +1,106 @@
...
+cd /tmp/bref/php-src-php-$PHP_VERSION/ext/intl
+phpize > /dev/null
+./configure > /dev/null
+make > /dev/null
+make install
+
+# Copy the generated binary into /opt/bin (this directory is automatically included in the $PATH)
+# See https://docs.aws.amazon.com/lambda/latest/dg/configuration-layers.html#configuration-layers-path
+mkdir /opt/bin
+cp /tmp/bref/php-src-php-$PHP_VERSION/sapi/cli/php /opt/bin/
```
barryvdh: And also copy php-cgi perhaps? `cp /tmp/bref/php-src-php-$PHP_VERSION/sapi/cgi/php-cgi /opt/bin/`

mnapoli: Right I could if it's useful for tests right now, but at the end we want to make the layer as slim as possible (because it impacts cold starts and the limit on lambda is not that high).

Is it still useful to add it and publish a new layer or am I too late? (the benchmarks have been run it seems)

mnapoli:
@barryvdh I have separated the runtimes into 2:

- default (php-fpm)
- loop (`php -L`, mode G)

Could you explain what you did for your PHP-FPM layer? Is it just about removing `--disable-fpm` and adding the `php-fpm` binary in `/otp`?

Makefile
```diff
@@ -1,3 +1,8 @@
-# Build all the PHP versions supported by Bref
-build-php:
-	bin/php/build.sh 7.2.5
+# Build the PHP runtimes
+runtimes: runtime-default runtime-loop
+runtime-default:
+	cd runtime/default && sh publish.sh
+runtime-loop:
+	cd runtime/default && sh publish.sh
```
sandrokeil: It should be `cd runtime/loop`

barryvdh:
Yes I just enabled fpm and copied sapi/fpm/php-fpm to opt/bin

Makefile
mnapoli: Woops thanks!

PR #107, opened 2018-12-05
Append querystring to uri

barryvdh:
I'm not 100% sure, but doesn't the Query string belong in the URI? According to https://www.php-fig.org/psr/psr-7/

Eg. if I dump my $_SERVER in a normal request, the querystring is appended to the REQUEST_URI


barryvdh:
Encountered this because the API gateway in my 'f' example wasn't passing the query parameters correctly when forwarding the PSR-7 request.

mnapoli:
Thanks!

mnapoli:
In Diactoros this is also the case: https://github.com/zendframework/zend-diactoros/blob/master/test/RequestTest.php#L88

Merging!

PR #110, opened 2018-12-06
Fix 'path'

barryvdh:
I think this should just be path, the other one contains something like '{proxy+}

### DIFF
--- a/src/Bridge/Psr7/RequestFactory.php
+++ b/src/Bridge/Psr7/RequestFactory.php
@@ -24,7 +24,7 @@ class RequestFactory
         $bodyString = $event['body'] ?? '';
         $parsedBody = null;
         $files = [];
-        $uri = $event['requestContext']['path'] ?? '/';
+        $uri = $event['path'] ?? '/';
         $headers = $event['headers'] ?? [];
         $protocolVersion = $event['requestContext']['protocol'] ?? '1.1';
 


mnapoli:
Thanks, I faced the same thing earlier this week but didn't have time to investigate further. I had the problem when running locally with AWS SAM.

But on API Gateway I think it was correct, do you get the problem when deployed?

Also IIRC this path doesn't contain the `/dev` prefix from API Gateway, so this doesn't represent the real URL.

barryvdh:
Not the real path, but does represent the expected value, right? So if you deploy your website you don't want to update your route to prefix it with Prod, right?

mnapoli:
Agreed, but that creates side-effects. Like for example in Symfony you would have links pointing to wrong URLs.

Here is an example: you are on the home page at `/dev`. Symfony believes it's `/`. If you generate links in a template to `/about` then this URL will not work (it should be `/dev/about` instead).

I'm not 100% sure what to do here 😕

barryvdh:
Ah yeah, not sure how symfony usually handles being in a subdirectory.

nealio82:
I'm not 100% sure on this, but at the moment the only way I can see working across all three environments (php built-in server, SAM local, Lambda) would be to have an option in the `index.php` file which allows you to turn on prepending the stage into the URI when in development mode (eg, can we detect if we're in SAM local and adjust accordingly?).

eg (using Symfony as an example): 
- in Symfony you prepend all your routes with `/dev` and use the built-in server. all routes have `'/dev/...`
- in your `index.php` you turn on the config option to prepend all routes with `/dev`. therefore `$event['path']` `/about` becomes `/dev/about` when using SAM local
- on Lambda without a custom domain name, `dev` is provided in the url so the routes still work

When ready to 'go live' later on:
- in symfony you remove all the `/dev` route prefixes
- in `index.php` you turn off the 'stage/dev' config option. SAM local now has 'normal' routes again
- on Lambda the stage name is set on the API config, and the framework goes back to seeing non-prefixed routes

I need to dig into it / think about it more - sometimes this makes perfect sense as a solution to me, other times it seems completely unnecessary.

barryvdh:
It's a bit confusing that the local preview doesn't have the /Prod prefix, and the path is wrong. Is that actually a bug with the local preview?

mnapoli:
OK I think this change landed in v0.3.

What I suggest is to close this and see when testing if problems arise with the new version.

PR #111, opened 2018-12-09
use $event['path'] to get route

nealio82:
`RequestFactory` currently uses `$event['requestContext']['path']` to figure out the route, although using the new SAM Local / runtime API system with the following config:

```
Resources:
    DemoFunction:
        Properties:
            Events:
                # ...
                HttpSubPaths:
                    Type: Api
                    Properties:
                        Path: /{proxy+}
                        Method: ANY
```

results in `No route found for "POST /{proxy+}"` when `POST`ing using SAM local (although still works on prod).

Dumping `$event['requestContext']` on SAM local shows:

```
["path"]=>
    string(9) "/{proxy+}"
```

But we now have access to both `$event['path']` _and_ `$event['pathParameters']['proxy']` instead, which both appear to give the current URL

### DIFF
--- a/src/Bridge/Psr7/RequestFactory.php
+++ b/src/Bridge/Psr7/RequestFactory.php
@@ -24,7 +24,7 @@ class RequestFactory
         $bodyString = $event['body'] ?? '';
         $parsedBody = null;
         $files = [];
-        $uri = $event['requestContext']['path'] ?? '/';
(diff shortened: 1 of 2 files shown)


mnapoli:
Thanks! Please see the discussion in #110 this is about the same thing. I'm not sure what to do here 😕

Issue #99, opened 2018-11-25
Consider Serverless framework vs AWS SAM


mnapoli:
Bref is currently using the [serverless framework](https://serverless.com/) internally to deploy lambdas. I want to explore whether using AWS SAM could be an alternative.

|  | Serverles | AWS SAM |
|--|:---------:|:-------:|
| Cross-provider | ✅ | ❌ (limited to AWS) |
| Easy deployment from CLI | ✅ | ✅ |
| Deployment using a CloudFormation stack | ✅ | ✅ |
| Allow extra CloudFormation resources | ✅ | ✅ |
| Allow to run hooks before deployment | ✅ | ❓ |
| Run preview locally | ✅ | ✅ |
| Run preview in Docker | ❌ (not natively) | ✅ |
| Integration with CodePipeline/CodeDeploy | ❌ | ✅ |
| Deployment strategies | ❌ | ✅ |
| Integration with AWS SAR (serverless application repository) | ❌ | ✅ |
| Deployment of different stages | ✅ | ❓ |
| Deployment of different stages to different accounts | ✅ | ❓ |

Other lines to come, feel free to comment to add questions.

This comes with 2 questions:

- should Bref only target AWS?

  I'm leaning on a very strong yes here considering how everything even today with serverless (which is generic) is strictly tied to AWS.
- should Bref hide the details of the deployment tools, or should Bref integrates in the deployment tool?

  What I'm getting at is that Bref could be a "whole in one" solution, hiding serverless/SAM (so that we control the details and we can provide a very simple user experience). Or Bref could be the "PHP part" that fits inside serverless/SAM (so that we can benefit from everything those frameworks have to offer).

nealio82:
One huge positive here is that you’ll be able to more accurately replicate the Lambda environment locally. I find with debugging that needing to deploy between changes _really_ slows down the feedback loop.

With regard to ‘should we only target AWS?’; this is, as you’re no doubt aware, a tough question. AFAIK we can currently only run Bref on Lambda ATM anyway, so I doubt the current user base will be affected. However, what would that mean for the project if AWS natively supported PHP at some point in the future? Would we switch back to server less?

mnapoli:
Agreed with the feedback loop, honestly what they did in SAM-CLI is pretty amazing, it works really well.

> what would that mean for the project if AWS natively supported PHP at some point in the future? Would we switch back to serverless?

I would love for PHP to be supported natively! Bref has 3 goals right now:

- PHP support through the JS shim
[...]

PR #113, opened 2018-12-16
v0.3 - Official PHP support on Lambda, better performances and AWS SAM

mnapoli:
This pull request will contain all the changes planned for v0.3 to keep master (v0.2.*) stable in the meantime.

**The future website for Bref 0.3: [bref.sh](https://bref.sh)**

### Summary of changes

AWS has released an API to provide support for any language in Lambda. Bref will move from its Javascript-based hack to this new official API.

Bref will provide a PHP runtime **optimized for CLI scripts** (workers, CLI tasks, simple PHP lambdas, etc.) as well as another PHP runtime **optimized for HTTP applications** (based on PHP-FPM).

It is very likely that Bref will switch from the "serverless framework" to using AWS SAM instead for deployment and configuration (see #99). On top of advanced integration to other AWS services, this brings very useful **tools for local development** based on Docker.

With v0.3 it will be even easier to integrate HTTP applications and support any framework. **Performances will also dramatically increase** thanks to [research and benchmarks](https://github.com/mnapoli/bref-bootstrap-benchmarks) as well as new experimental modes to run PHP faster in some scenarios (more information to come).

Finally, Bref v0.3 will come with a completely **rewritten and extensive documentation**.

### v0.3

Check out the **[v0.3 Kanban board](https://github.com/mnapoli/bref/projects/2)**.

### Detailed changes

- #106 PHP runtime for AWS Lambda (built through the official support for 3rd party languages)
- #112 Client for the Lambda runtime API
- #115 PHP-FPM runtime optimized for HTTP applications
- #122 A console runtime for Console applications
- #124 The `λ()` function was replaced by a `lambda()` function
- #127 Website and documentation
- #126 Migrate to AWS SAM for deployments
- #129 an industrialized build pipeline for the runtimes


nealio82:
This is amazing! What's the long-term plan for building the runtime? Will `bref` provide stock runtimes for people who are just trying things out?

Also, will building the runtime be invoked as part of the first `bref deploy` or does the user always need to do it themselves?  What about for using SAM local? It might be nice to have a single `bref develop` command in the `Makefile` which creates the runtime and sets up the config for SAM Local.

mnapoli:
@nealio82 I finally answered in #114: https://github.com/mnapoli/bref/issues/114#issuecomment-448400235

About building runtimes: ideally you shouldn't have to worry about this, it just works (see the WIP instructions in the diff: https://github.com/mnapoli/bref/pull/113/files#diff-279a8ca151371a6a63058dc411985dd1R9). And locally it just works too, SAM is incredibly awesome for this. The only thing I'm worried with SAM when running locally is performances because you get a cold start every time. But you can also still run Symfony/Laravel the only way locally, it still works that way. We'll see as we use it more.

Makefile
```diff
@@ -1,3 +1,17 @@
-# Build all the PHP versions supported by Bref
-build-php:
-	bin/php/build.sh 7.2.5
+.EXPORT_ALL_VARIABLES:
+
+BREF_VERSION := 0.3
+REGION := us-east-2
+PHP_VERSION := 7.2.12
```
staabm:
```suggestion
PHP_VERSION := 7.2.13
```

Makefile
```diff
@@ -1,3 +1,17 @@
...
+PHP_VERSION_SHORT := 72
+
+# Build the PHP runtimes
+runtimes: runtime-default runtime-fpm runtime-loop
+runtime-default:
+	cd runtime/default && sh publish.sh
+runtime-fpm:
+	cd runtime/fpm && sh publish.sh
+runtime-loop:
```
staabm: was the loop runtime intentionally added?

runtime/default/bootstrap
```diff
@@ -0,0 +1,11 @@
+#!/opt/bin/sh
```
staabm:
```suggestion
#!/usr/bin/env sh
```

?

runtime/default/build.sh
```diff
@@ -0,0 +1,105 @@
+#!/bin/sh
```
staabm: different then in the file above

runtime/default/build.sh
```diff
@@ -0,0 +1,105 @@
...
+    php-pear \
+    libicu \
+    libicu-devel \
+    c++ \
+    gcc-c++
+
+# Compile PHP
+mkdir /tmp/bref
+cd /tmp/bref
+curl -sL https://github.com/php/php-src/archive/php-$PHP_VERSION.tar.gz | tar -zxv
```
staabm: wget fallback when curl is not available?

runtime/default/php.ini
```diff
@@ -0,0 +1,25 @@
+extension_dir=/opt/php/extensions
+display_errors=On
```
staabm: will those be rendered to the end-user like in regular php?

src/Runtime/LambdaRuntime.php
```diff
@@ -0,0 +1,154 @@
...
+     * Wait for the next lambda invocation and retrieve its data.
+     *
+     * This call is blocking because the Lambda runtime API is blocking.
+     *
+     * @see https://docs.aws.amazon.com/lambda/latest/dg/runtimes-api.html#runtimes-api-next
+     */
+    private function waitNextInvocation(): array
+    {
+        $handler = curl_init("http://{$this->apiUrl}/2018-06-01/runtime/invocation/next");
+        curl_setopt($handler, CURLOPT_FOLLOWLOCATION, true);
```
staabm: would it make sense to use http2 for this requests?

src/Runtime/LambdaRuntime.php
```diff
@@ -0,0 +1,154 @@
...
+    /**
+     * Wait for the next lambda invocation and retrieve its data.
+     *
+     * This call is blocking because the Lambda runtime API is blocking.
+     *
+     * @see https://docs.aws.amazon.com/lambda/latest/dg/runtimes-api.html#runtimes-api-next
+     */
+    private function waitNextInvocation(): array
+    {
+        $handler = curl_init("http://{$this->apiUrl}/2018-06-01/runtime/invocation/next");
```
staabm: intentional non-secure http?

src/functions.php
```diff
@@ -1,17 +1,24 @@
...
  * @param callable $handler This callable takes a $event parameter (array) and must return anything serializable to JSON.
- *
- * @see \Bref\Application::simpleHandler()
  */
 function λ(callable $handler): void
 {
-    $app = new Application;
-    $app->simpleHandler($handler);
-    $app->run();
+    lambda($handler);
```
staabm: this will be a very common used function.

Would it be worth to duplicate the impl to save a function call?

Makefile
mnapoli: Yes

runtime/default/bootstrap
mnapoli: Any reason to prefer one or the other? Also are we certain that the alternative works on Lambda? (compiling, publishing the version, creating a test lambda, deploying and testing takes a good amount of time just for validating this little change)

runtime/default/build.sh
mnapoli: This one doesn't run in Lambda, it runs on the build machine.

runtime/default/build.sh
mnapoli: Do you have a replacement that works for that line?

runtime/default/bootstrap
staabm: I dont know about lambda. But using `/usr/bin/env sh` Is a way which also supports shells which are installed in non standard paths. Its good practice to use.

runtime/default/php.ini
mnapoli: No this is for the default runtime which doesn't run over PHP-FPM or similar. Any error in there will end up in Cloudwatch.

I need to test this out more to make it perfect, this is part of #87

src/Runtime/LambdaRuntime.php
mnapoli: Can you check the Lambda runtime documentation and tell me if there is an error here? Since I follow the  official documentation I'll assume that it is correct in the meantime.

src/Runtime/LambdaRuntime.php
mnapoli: HTTP2 for the lambda runtime API? Is it working or documented somewhere?

src/functions.php
mnapoli: That made me think more about it and I opened #124 to actually remove this alias.

runtime/default/bootstrap
mnapoli: I understand.

Until we have a reason to choose one, I don't want to compile, publish a layer, create a test lambda, deploy and test that it works for no perceived gain. Let's revisit when we have more information.

src/Runtime/LambdaRuntime.php
staabm: All the runtime docs reference http and dont mention https at all

src/Runtime/LambdaRuntime.php
staabm: I asked for feedback on the docs and hope they will clarify what is supported (https and http2)

mnapoli: Honestly I don't see any reason they will want to support HTTPS (or HTTP2) in this context since this is an internal (low-level) API.

mnapoli: Let's mark this resolved for now and reopen if there's any reason to.

demo/cli.php
```diff
@@ -0,0 +1,11 @@
+<?php declare(strict_types=1);
+
+require __DIR__.'/../vendor/autoload.php';
+
+ini_set('display_errors', '1');
```
staabm:
```suggestion
ini_set('display_errors', 'stderr');
```

mnapoli: Could you explain this suggestion?

Juhlinus: 'stderr' outputs the errors in the console, whereas the '1' would output to file.

mnapoli: Right but why would it make sense here, in the context of AWS Lambda?

PR #115, opened 2018-12-18
PHP-FPM runtime

mnapoli:
Second part of #100. This is the PHP-FPM runtime for HTTP applications.

TODO:

- [x] Request tests
- [x] Response tests


barryvdh:
FYI, I'm not really sure about the library here. The alternative is https://github.com/hoaproject/Fastcgi which I later found also supoorts file sockets, bit needs a prefix (file:// or something). It's a bit more lower level, the current one abstracts it a bit too much perhaps

mnapoli:
Yeah I did a bit of a search yesterday and there are not a lot of alternatives unfortunately (https://packagist.org/search/?q=fastcgi+client).

Hoa is a very solid project but I'm not sure the author is involved anymore. So I'm not sure about this…

barryvdh:
If it works, not sure why it needs to be changed ;) It's pretty low-level (just passing the parameters), so I guess not much would change there. The hollodotme maps everything in a request object, which isn't PSR-7. So you need to convert everything to that object, which isn't totally clear.

This was basically it: https://github.com/mnapoli/bref-bootstrap-benchmarks/pull/7/commits/1de6d9e55425f81dc4d918f8d1b5aff6830e7630

mnapoli:
Oh wait I completely missed the commit you are referencing. I'll look into that because that looks much simpler indeed. Thanks!

barryvdh:
Yeah only not sure if the parameters are correct, and you need to use `file:///<socket path>` or something.

mnapoli:
Opened #117 to give it a try later, that way we can start testing. Thanks!

Hywan:
> Hoa is a very solid project but I'm not sure the author is involved anymore. So I'm not sure about this…

I think I've to write a blog post to clarify since you're not the only asking :-). I slowed down my participation in all my open source projects due to health issue, but now I'm back. I need to catch up everything. But Hoa is definitely not dead!

mnapoli:
@Hywan awesome!

PR #119, opened 2018-12-19
Replace hollodotme/fast-cgi-client with hoa/fastcgi

### LINKED ISSUE #117: Try the Hoa FastCGI library
Followup of #115 to simplify the FPM runtime.

See https://github.com/mnapoli/bref/pull/115#issuecomment-448510349

mnapoli:
Closes #117 

@hywan trying out Hoa\FastCGI :) Can you confirm if that package is maintained? Also if you see any optimization to be done here I'd be happy to hear it :) (I've seen `Hoa\Socket\Client::PERSISTENT` and I'm not sure if it's appropriate here).


Hywan:
Hello :-),

Yes, this package is still maintained, it is used by multiple companies and many projects. We are rebuilding many stuff in Hoa and it takes time, hence the few activities on `master`s, but we should reach the end of this period soon I hope.

`hoa/fastcgi` is interesting because it relies on `hoa/socket`, and we should, at some point in the future, get native async instructions in `hoa/socket`, so `hoa/fastcgi` will suddenly be async. I don't have any schedule though.

Also, it's good to note that I'm not very proud of the API of `hoa/fastcgi`. It works, it's solid, but, eh, you know. So if you have remarks or feedbacks, they will be much welcomed.

Thanks!

src/Runtime/PhpFpm.php
```diff
@@ -140,31 +142,45 @@ private function eventToFastCgiRequest(array $event): AbstractRequest
...
+            'REQUEST_METHOD' => $event['httpMethod'],
+            'REQUEST_URI' => $uri,
+            'SCRIPT_FILENAME' => $this->handler,
+            'SERVER_SOFTWARE' => 'bref',
+            'REMOTE_ADDR' => '127.0.0.1',
+            'REMOTE_PORT' => '80',
+            'SERVER_ADDR' => '127.0.0.1',
+            'SERVER_NAME' => $serverName,
+            'SERVER_PROTOCOL' => $protocol,
+            'CONTENT_LENGTH' => mb_strlen($requestBody, 'UTF-8'),
```
Hywan: If the content length is expressed as UTF-8 counting, then the content-type must be aware of that assumption. Something like adding `; charset=utf-8` or similar, thoughts?

mnapoli:
Perfect thank you for letting us know (and doing all this!).

src/Runtime/PhpFpm.php
mnapoli: I'm not even sure we should set that every time, see http://php.net/manual/fr/reserved.variables.server.php#110763

@barryvdh I picked that from your implementation, do you know what to do here?

To be honest I have no idea, any suggestion is welcome. Does it make sense to add `; charset=utf-8` to the content-type automatically? The goal would be to stay as close as possible to Apache/Nginx + PHP-FPM.

Hywan: I think you should get them from the HTTP request to be honest :-).

PR #122, opened 2018-12-24
Add console runtime for Symfony Console and Laravel Artisan

### LINKED ISSUE #120: Runtime or layer for CLI applications
With v0.3 we get rid of the principle where we have `bref.php` that receives every event and decides what to execute ("is it a HTTP call, simple lambda event or CLI execution?").

Now we have 2 runtimes:

- PHP
- PHP-FPM

What's missing from Bref v0.2 is the ability to run console commands (Symfony Console or Laravel Artisan) in a lambda.

To keep that feature we could create a runtime that makes it super easy to do so.

For example:

```yaml
...
Resources:
    ConsoleFunction:
        Properties:
            Handler: bin/console
            Runtime: provided
            Layers:
                - 'arn:aws:lambda:us-east-2:416566615250:layer:php-cli-72:1'
```

(notice the handler is `bin/console` and the PHP runtime is `php-72-cli`, not `php-72`)

In that layer we would have a `bootstrap` that takes the incoming event and convert it to a CLI call.

The only thing that would differ from the `php-72` runtime is the `bootstrap` so we could either:

- create a separate `php-72-cli` runtime
- create a layer that only overrides the `/opt/bootstrap` file (so you use the `php-72` layer *plus* the `php-cli` layer on top of it), it might be easier to maintain

WDYT?

mnapoli:
Closes #120 

I implemented this as a thin layer that comes on top of the PHP runtime so that it's much easier to maintain and not coupled to a specific PHP version.

To use this: include the PHP runtime and add the "console" layer on top of it.

Usage example:
 
```yaml
Resources:
     MyFunction:
         Properties:
             [...]
             Handler: bin/console
             Layers:
                 - '<PHP layer ARN here>'
                 - '<console layer ARN here>'
```
 
Then to execute a command:
 
```
vendor/bin/bref cli <function name> -- doctrine:migrate
```

capture d ecran 2018-12-26 a 12 37 28

TODO:

- [x] runtime
- [x] `bref cli` command
- [x] publish runtime


demo/console.php
```diff
@@ -0,0 +1,21 @@
...
+
+$silly = new Application;
+
+$silly->command('hello [name]', function (string $name = 'World!', OutputInterface $output) {
+    $output->writeln('Hello ' . $name);
+});
+$silly->command('phpinfo', function (OutputInterface $output) {
+    ob_start();
+    phpinfo();
+    $phpinfo = ob_get_contents();
```
staabm:
```suggestion
    $phpinfo = ob_get_clean();
```
And drop next line

runtime/command/README.md
```diff
@@ -0,0 +1,23 @@
...
+            [...]
+            Handler: bin/console
+            Layers:
+                - '<PHP layer ARN here>'
+                - '<command layer ARN here>'
+```
+
+Then to execute a command:
+
+```
```
staabm:
```suggestion
```php
```

mnapoli:
I created a follow up in #125, for now it works.

PR #128, opened 2018-12-29
add 'what is serverless?' page

nealio82:
Added a 'what is serverless' page including reasoning & costing examples behind why switching to serverless hosting from traditional VPS systems might be a good idea

### DIFF
--- a/docs/what-is-serverless.md
+++ b/docs/what-is-serverless.md
@@ -1,0 +1,149 @@
+# What is serverless?
+
+These days the term **serverless** is used to describe a lot of different variations on the same general theme; that you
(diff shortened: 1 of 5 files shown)


nealio82:
I agree - maybe the cost analysis belongs elsewhere (case studies would be an excellent choice), and it'd almost certainly make more sense to merge the 'what is serverless?' with 'what is bref?'

i'd love to see a getting started / tutorials / cookbook section in the docs; i'm more than happy to work on these if you haven't already..?

mnapoli:
> getting started / tutorials / cookbook section in the docs

I'm not sure I follow, what would be in there?

Issue #130, opened 2018-12-29
RequestFactory::fromLambaEvent not aggregating ALL request context


krcrawford:
Is there a reason that additional values from the request context in the lambda event are not read into the server global?
My case, is that I'd like to see the `$_SERVER['REMOTE_ADD']` set to the `event['requestContext']['identity']['sourceIp']`, which in turn will open the `trustedProxies` feature in Symfony.
`RequestFactory::fromLambdaEvent` is not ingesting more information currently, so is a PR welcome to expand that, or is this an intentional strategy?

PR #131, opened 2018-12-31
Add remote addr and check x forwarded

krcrawford:
Add setting of `$_SERVER['REMOTE_ADDR']` and checks for `X-Forwarded-For` header (setting REMOTE_ADDR to last X-Forwarded-For entry)
References issue #130 
Laravel and Symfony both rely on a `REMOTE_ADDR`, so unless these frameworks start evaluating deeper conditions in `$_SERVER` items availability (the frameworks don't fallback to X-Forwarded-For, at least directly), then trusted proxies wouldn't work with bref _without_ something similar to this PR.
Interested in different thoughts, so please share if you have alternatives or see problems.


src/Bridge/Psr7/RequestFactory.php
```diff
@@ -95,6 +95,23 @@ public static function fromLambdaEvent(array $event): ServerRequestInterface
...
+            $server['REMOTE_ADDR'] = $event['requestContext']['identity']['sourceIp'];
+        }
+
+        if (isset($headers['X-Forwarded-For'])) {
+            $ips = [$headers['X-Forwarded-For']];
+            if (false !== strpos($ips[0], ',')) {
+                $ips = preg_split('/,\s?/', $ips[0]);
+                $ip = $ips[count($ips) - 1];
+                if (\filter_var($ip, FILTER_VALIDATE_IP)) {
+                    $server['REMOTE_ADDR'] = $ip;
```
staabm: re-assignment of the ip should only be based on a ip whitelist

krcrawford: I see your point, as I intend to use this along with a trusted proxy list. When using symfony, whitelisting doesn't work without a `REMOTE_ADDR`. Perhaps you have another suggestion to better mesh the `REMOTE_ADDR` property? In truth, aren't all IPs fairly spoofable? With that in mind, I'm just looking to get the initial values passed into symfony, then I'll whitelist.

mnapoli:
Thank you for the PR, this is indeed something we need to take advantage of.

You made a PR against master which is v0.2, v0.3 is the next version and changes a lot of things. Would you be able to send a PR against this new version? (this is the `0.3` branch).

Here is where I turn the API Gateway event into a FastCGI request for PHP-FPM: https://github.com/mnapoli/bref/blob/0.3/src/Runtime/PhpFpm.php#L159-L173 I guess this is where we can pass more info right?

mnapoli:
Closing for now.

Issue #129, opened 2018-12-29
PHP Layer(s) Runtime Pipeline for v3.0


bubba-h57:
We are opening this issue for discussion pertaining to how *bref* builds and publishes runtimes for the supported layers. We are prepared to provide a pull request implementing all features discussed here.

- [ ] Consolidate Build Process
- [ ] Implement Build Pipeline
- [ ] Use [mult-stage docker builds](https://docs.docker.com/develop/develop-images/multistage-build/) where possible
- [ ] Use the [Offical AWS Docker Image](https://github.com/aws/amazon-linux-docker-images/tree/2017.03) for our build
- [ ] Move the *bref* Layer(s) from `/opt` to `opt/bref`
- [ ] Update the PHP Build to no longer use the _[deprecated]_ bundled [libzip](http://php.net/manual/en/zip.installation.php)
- [ ] Update the Layer(s) to [OPENSSL 1.1.1a](https://github.com/openssl/openssl/releases)
- [ ] Update the Layer(s) to [CURL 7.63.0](https://github.com/curl/curl/releases) w/OPENSSL 1.1.1a
- [ ] Update the Layer(s) to [LIBSSH 1.8.0](https://github.com/libssh2/libssh2/releases)
- [ ] Update the Layer(s) to [JPEG-TURBO 2.0.1](https://github.com/libjpeg-turbo/libjpeg-turbo/releases)
- [ ] Update the Layer(s) to [PNG16 1.6.36](https://libpng.sourceforge.io/)
- [ ] Update the Layer(s) to [WEBP 1.0.1 ](https://storage.googleapis.com/downloads.webmproject.org/releases/webp/index.html)
- [ ] Update the Layer(s), add the [Sodium](http://php.net/manual/en/book.sodium.php) Extension
- [ ] Update the Layer(s),  add [IMAGEMAGICK 7.0.8-20](https://github.com/ImageMagick/ImageMagick)
- [ ] Update the Layer(s),  add the [IMAGICK](https://github.com/mkoppanen/imagick) Extension
- [ ] Update the Layer(s), add the [pdo-pgsl](http://php.net/manual/en/ref.pdo-pgsql.php) and [pgsl](http://php.net/manual/en/book.pgsql.php) Extensions
- [ ] Compile PHP to natively load Configuration File `/opt/bref/etc/php/php.ini`
- [ ] Compile PHP to natively scan `/opt/bref/etc/php/config.d:/var/task/php/config.d` for additional `.ini` files
[...]

PR #132, opened 2019-01-02
 PHP Layer(s) Runtime Pipeline for v3.0

bubba-h57:
Implements all discussed in issue #129


bubba-h57:
Here is a copy of the [php-default-7.3.0.zip](https://github.com/mnapoli/bref/files/2721657/php-default-7.3.0.zip) package I created with the build system. Makes it a bit easier to unzip the package and take a look at the contents without having to build it yourself.

I also have the `/opt/bin` directory populated with symlinks to the `/opt/bref/bin` directory, ensuring that all the binaries simply end up on the path.

runtime/experimental/compiler.Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+RUN yum groupinstall -y "Development Tools"  --setopt=group_package_types=mandatory,default
+RUN yum install -y  jq \
+                    gperf \
+                    expect \
+                    gtk-doc \
+                    texlive \
+                    python35 \
+                    docbook2X \
+                    findutils \
+                    python35-pip
```
mnapoli: Are you able to document why each of these dependency is needed? That helps a lot to maintain in the long run.

runtime/experimental/compiler.Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+
+RUN yum clean all
+
+# Install Ninja and Meson
```
mnapoli: Could you document why we need those?

runtime/experimental/compiler.Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+RUN curl -Ls https://github.com/ninja-build/ninja/releases/download/v1.8.2/ninja-linux.zip >> /tmp/ninja.zip && \
+    cd /tmp && unzip /tmp/ninja.zip && \
+    cp /tmp/ninja /usr/local/bin && \
+    /usr/bin/pip-3.5 install meson
+
+# Install the rust toolchain
```
mnapoli: Same here, I'm not sure why we need rust

runtime/experimental/compiler.Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+RUN curl https://sh.rustup.rs -sSf | sh -s -- -y
+
+# We need a newer cmake than is available, so lets build it ourselves.
```
mnapoli: > We need a newer cmake than is available

I think it's good that we explain why?

runtime/experimental/makefiles/apcu.mk
```diff
@@ -0,0 +1,23 @@
+.DEFAULT_GOAL := make_apcu
+
+url_apcu = https://github.com/krakjoe/apcu/archive/v${VERSION_APCU}.tar.gz
+upper_apcu = $(shell echo apcu| awk '{print toupper($0)}')
+build_dir_apcu = ${DEPS}/apcu
```
mnapoli: Those variables used in all the `*.mk` files make it very hard to follow. Can we inline them to simplify further? At least those that are only used once.

runtime/experimental/makefiles/curl.mk
```diff
@@ -0,0 +1,45 @@
...
+		--enable-ipv6 \
+		--enable-openssl-auto-load-config \
+		--enable-cookies \
+		--with-gnu-ld \
+		--with-ssl \
+		--with-libssh2
+
+build_curl:
+	cd ${build_dir_curl} && \
+	$(MAKE) install
```
mnapoli: Why is the `curl` extension (and other similar extensions) compiled from source and not installed via `--with-curl` like [here](https://github.com/mnapoli/bref/blob/ebdf5331deadb2d7a2a794b9175430026dc7d779/runtime/default/build.sh#L56-L69)?

runtime/experimental/Makefile
```diff
@@ -0,0 +1,73 @@
+SHELL := /bin/bash
+
+compiler:
+	docker build -f ${PWD}/compiler.Dockerfile -t bref/runtime/compiler:latest .
+
+shell/compiler:
+	docker run --interactive --tty -v ${PWD}:/host --entrypoint /bin/bash -e "TERM=xterm-256color" bref/runtime/compiler:latest
```
mnapoli: I think we can get rid of those "shell/" commands to simplify the Makefile, maybe we could document the `docker run` commands in the readme instead so that they can still be used?

runtime/experimental/helpers/export.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+cp /tmp/php-$(php -r 'echo phpversion();').zip /export/php-fpm-$(php -r 'echo phpversion();').zip
+cp /tmp/php-cli-$(php -r 'echo phpversion();').zip /export/php-default-$(php -r 'echo phpversion();').zip
```
mnapoli: Is it possible to simplify here (and elsewhere) by not having the version name in the temporary files/folders?

runtime/experimental/makefiles/php.mk
```diff
@@ -0,0 +1,93 @@
+SHELL := /bin/bash
+.DEFAULT_GOAL := make_php
+
+url_php = https://github.com/php/php-src/archive/php-${VERSION_PHP}.tar.gz
+build_dir_php = ${DEPS}/php
+
+fetch_php:
+	mkdir -p ${build_dir_php}
+	${CURL} -Ls ${url_php} | tar $(shell ${TARGS} ${url_php}) ${build_dir_php} --strip-components=1
```
mnapoli: I'm trying to wrap my head around everything. For example here why is this `${CURL}` and not simply `curl`?

What is the `${TARGS}` variable?

What is `$(shell ${TARGS} ${url_php})` resolving to?

runtime/experimental/versions.ini
```diff
@@ -0,0 +1,127 @@
...
+imagick = "3.4.3"
+memcached = "1.0.18"
+php_memcached = "3.1.3"
+
+[php70]
+php = "7.0.33"
+imagemagick = "7.0.8-20"
+imagick = "3.4.3"
+memcached = "1.0.18"
+php_memcached = "3.1.3"
```
mnapoli: We can drop PHP 7.0 as it's not supported anymore.

runtime/experimental/Makefile
```diff
@@ -0,0 +1,73 @@
...
+shell/distro: 
+	docker run --interactive --tty -v ${PWD}:/host --entrypoint "/bin/bash" -e "TERM=xterm-256color" bref/runtime/dist:latest
+
+export: distro
+	mkdir -p ${PWD}/exports
+	docker run --interactive --tty -v ${PWD}/exports:/export -e "TERM=xterm-256color" bref/runtime/dist:latest
+	ls -al ${PWD}/exports
+
+clean:
+	docker system prune -f
```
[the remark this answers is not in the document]
bubba-h57: Good point.

runtime/experimental/makefiles/curl.mk
bubba-h57: When php-curl is built against a different openssl library than PHP itself is built against, that complicates things.

In order to have PHP built against a curl library that is built against a sane Openssl version, we have to build it ourselves. For developers who have legal compliance requirements, this is important.

Building against the php-curl extension against the latest curl ensures that our developers have access to the latest curl feature-set, as well as the collective bugfixes.

**Technical Details**
That `--with-curl` is actually building against the old system curl ([version 7.51.0, released Nov 3, 2016](https://github.com/curl/curl/releases?after=curl-7_52_0)), those shared libraries/executables are already installed on the system by default, and the development files are being add with 
 https://github.com/mnapoli/bref/blob/ebdf5331deadb2d7a2a794b9175430026dc7d779/runtime/default/build.sh#L29

A cursory glance at https://curl.haxx.se/changes.html#7_52_0 through https://curl.haxx.se/changes.html#7_63_0 looks like over 1,000 bugfixes and perhaps a hundred or so other changes.

Additionally, the system curl is compiled against [OpenSSL 1.0.1k-fips 8 Jan 2015](https://github.com/openssl/openssl/releases?after=OpenSSL_0_9_8ze) as well as a number of other old libs (zlib, libssh2) that differ from the version we will compile PHP against.

runtime/experimental/Makefile
bubba-h57: Certainly. They were simply convenient for developing the build system itself.

runtime/experimental/helpers/export.sh
bubba-h57: It is.

I was envisioning a scenario where the runtime system and default layers are managed internally to bref. That is, we would build and publish all the PHP Version layers that we support out of the box, to all regions we support out of the box. In that scenario, it would be very rare for someone outside of bref to build/publish.

If we keep the version in the names, we can look in the S3 bucket(s) to see a history of the various versions we have uploaded. Ultimately, I envision having some associated meta-data with the packages in S3 that designate the `layer:version` that package is associated with.

In that scenario, it seemed good to automate building all the php packages and then uploading/registering them.

Which sorta brings up the question of how we afford to pay for S3 hosting. I will open an issue to chase that rabbit though. 

Ultimately, we can certainly simplify it, it will just add further complication if we desire to automate the building of multiple versions. It also complicates tracking the php packages, by version, in an S3 bucket. But certainly doable.

runtime/experimental/makefiles/php.mk
bubba-h57: We set some baseline Environment Variables in the compiler container:

https://github.com/mnapoli/bref/blob/9ea33c127d7ff2055c5b1f18266e7ad7904ebb41/runtime/experimental/compiler.Dockerfile#L63-L82

`${TARGS}` contains the path to our [get_tar_args.sh helper script](https://github.com/mnapoli/bref/blob/9ea33c127d7ff2055c5b1f18266e7ad7904ebb41/runtime/experimental/helpers/get_tar_args.sh). That script determines what arguments `tar` will need in order to properly unpack the sourcecode. 
https://github.com/mnapoli/bref/blob/9ea33c127d7ff2055c5b1f18266e7ad7904ebb41/runtime/experimental/compiler.Dockerfile#L78

${CURL} contains the path to the `curl` exectable we compiled ourselves. 
https://github.com/mnapoli/bref/blob/9ea33c127d7ff2055c5b1f18266e7ad7904ebb41/runtime/experimental/compiler.Dockerfile#L79

Once we build our own Openssl, anything that relies on SSL communications can get funky, because we have two versions of Openssl on the system, and we have out environment setup to look for our own newly compiled libraries/executables first, then look for the older system libraries.

I ran into a handful of situations that got complicated because the old system curl got picked up, but couldn't find the the old openssl libraries (because of the way *LD_LIBRARY_PATH* is setup for development). So instead of relying on `$PATH` to magically find the correct binaries and libraries ... we explicitely use `/usr/bin/curl` for the system curl, prior to building our own curl, then we swap over to `${CURL}`

We use the `${CURL}` variable for our own path because it relies on the `${TARGET}` which could potentially be changed. If `${TARGET}` is modified, then `${CURL}` simply keeps working as advertised.

`$(shell ${TARGS} ${url_php})`
[$(shell ...) ](https://www.gnu.org/software/make/manual/html_node/Shell-Function.html)is a GNU Make Function.

It would resolve into `$(shell  /usr/local/bin/get_tar_args.sh https://github.com/php/php-src/archive/php-${VERSION_PHP}.tar.gz) which would return `xzC` and then make would run `${CURL} -Ls ${url_php} | tar xzC ${build_dir_php} --strip-components=1`

runtime/experimental/helpers/export.sh
mnapoli: I see, let's revisit that later when things settle down ^^

runtime/experimental/makefiles/curl.mk
mnapoli: Awesome thank you

runtime/experimental/makefiles/apcu.mk
mnapoli: Let's revisit that later!

runtime/experimental/compiler.Dockerfile
bubba-h57: Commit https://github.com/mnapoli/bref/pull/132/commits/2a4e4a986ed30c728c4d0dd58593c50abb9f0298 addresses this.

runtime/experimental/compiler.Dockerfile
bubba-h57: Commit https://github.com/mnapoli/bref/pull/132/commits/2a4e4a986ed30c728c4d0dd58593c50abb9f0298 addresses this.

runtime/experimental/compiler.Dockerfile
bubba-h57: Commit https://github.com/mnapoli/bref/pull/132/commits/2a4e4a986ed30c728c4d0dd58593c50abb9f0298 addresses this.

runtime/experimental/compiler.Dockerfile
bubba-h57: Commit https://github.com/mnapoli/bref/pull/132/commits/2a4e4a986ed30c728c4d0dd58593c50abb9f0298 addresses this.

runtime/experimental/Makefile
bubba-h57: Done in commit https://github.com/mnapoli/bref/pull/132/commits/cafe64b1cfc09acf35557e0ddba7838fade54a28

runtime/experimental/versions.ini
bubba-h57: Done in commit https://github.com/mnapoli/bref/pull/132/commits/9a72b62ca1ed3d89d9e367f9982f5d919677574c

runtime/experimental/compiler.Dockerfile
mnapoli: This is awesome thank you!

runtime/experimental/export.Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+RUN rm -rf /opt/bref/data
+
+WORKDIR /opt
+RUN LD_LIBRARY_PATH= yum -y install zip
+RUN export PHP_ZIP_NAME=/export/php-$(php -r '$version = explode(".", phpversion());printf("%d.%d", $version[0], $version[1]);'); \
+ mkdir -p /export; \
+ zip -y -o -9 -r ${PHP_ZIP_NAME}.zip . -x "*php-cgi"; \
+ cp ${PHP_ZIP_NAME}.zip ${PHP_ZIP_NAME}-cli.zip; \
+ zip -d ${PHP_ZIP_NAME}-cli.zip bref/sbin/php-fpm bin/php-fpm; \
+ cp ${PHP_ZIP_NAME}.zip ${PHP_ZIP_NAME}.fpm.zip; \
```
mnapoli: should be `-fpm` instead of `.fpm`

runtime/experimental/export.Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+ zip -d ${PHP_ZIP_NAME}.fpm.zip bref/bin/php /bin/php
```
mnapoli: The files exported are:

- `php-7.2.zip` -> ✅ PHP CLI layer
- `php-7.2.fpm.zip` -> ✅ FPM layer
- `php-7.2-cli.zip` -> is it a duplicate of the first one?

bubba-h57: **php-7.2.zip** - This contains everything we could possibly build.
**php-7.2.fpm.zip** - Copy of  **php-7.2.zip**, removes the _CLI SAPI_, only provides the _FPM SAPI_.
**php-72-cli.zip** - Copy of  **php-7.2.zip**, removes the _FPM SAPI_, only provides the _CLI SAPI_.

sandrokeil:
What is the intention of pthreads/zts in a Lambda environment?

bubba-h57:
I do a quite a bit of multi-threading in our Lambda Jobs. Particularly for pooling a large number of tasks that can be done asynchronously/in parallel, yet are too shortlived for it to be cost effective to send each of them off to their own lambda function. There have been occasions where it has been beneficial for the sake of collecting all the outputs from the threads as well, without relying on a db or other network caching to report to and then collect from.

Due to the lack of writeable disk space, I also find myself doing a lot of work in memory that would traditionally be done by reading/writing to disk. For example, you can not download a >550GB file to a lambda function, writing it to `/tmp` and then doing anything with it. So I often download it to memory and work with it from there. 

PHP garbage collection is fairly lazy, it eventually happens, but it is lazy. By doing work in a thread, I ensure that when the work is done, all memory is immediately freed when the thread dies. The only thing I want back, if anything, is the artifact the thread generated. That way I ensure I do not run into an OOM issue when tossing around huge globs in RAM.

Back in the day, I would do that sort of work with PCNTL. But that was just clunky and prone to bugs. It also incurred a lot of support issues down the road when others attempted to support the scripts. As far as I am concerned, pthreads is a no brainer over PCNTL for the majority of use cases.

So, when I was looking at writing my bootstrap. I made sense to me that the Developers Lambda Function should execute in its own thread. There is no need for the overhead of spinning up a new php environment for it with `exec`, and no need to copy the entire environment with `pnctl` into a child process. Simply loading their function in a thread and catching the output ensures that nothing they do leaks into or breaks the bootstrap code. Furthermore, because it is a different memory space that dies when the function completes, the developer can not inadvertently create a memory leak by leaving artifacts laying around in memory that never get cleaned up between invocations.

That is my intention anyway.  Do you have reservations? Haven't had cause to leverage multi-threading in Lambda? Something else that peaked your curiosity?

sandrokeil:
@bubba-h57 Thank you for your explanation. I just wanted to know your use case.

PR #135, opened 2019-01-05
Reorder docs to match each variable for the ARN

deleugpn:
While reading I found it a bit confusing what the layer-version meaning was until I actually found it all the way down to the page. Perhaps it would be nice if the explanation were to follow the order of the variables.

### DIFF
--- a/docs/runtimes/README.md
+++ b/docs/runtimes/README.md
@@ -15,10 +15,21 @@ This page is an introduction to the runtimes. The next sections (PHP functions, 
 The name of the runtimes follow this pattern:
 
 ```
-arn:aws:lambda:<region>:416566615250:layer:<layer-name>:<layer-version>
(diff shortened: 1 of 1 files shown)


mnapoli:
That's a good idea thanks!

docs/runtimes/README.md
```diff
@@ -15,10 +15,21 @@ This page is an introduction to the runtimes. The next sections (PHP functions,
...
+### Supported regions (`<region>`)
+
+- `us-east-2`
+- other regions will be supported soon
+
+Remember to use the `<region>` that matches your application, else Lambda will not find the layer.
+
+### Account
+
+Bref Account id is: `416566615250`
```
mnapoli: I think having a placeholder + section for that is not necessary, it is simpler without it. Could you re-insert the account ID in the pattern above?

deleugpn: Done

mnapoli:
Thanks!

PR #140, opened 2019-01-05
updating sam deploy url

scottfoster:


docs/deploy.md
```diff
@@ -121,4 +121,4 @@ Remember to also delete the S3 bucket that was created to hold the application's
 
 ## Learn more
 
-Read more about `sam deploy` in [the official documentation](https://github.com/awslabs/aws-sam-cli/blob/develop/docs/deploying_serverless_applications.md).
```
mnapoli: The current URL seems more informative than the one you are suggesting though?

scottfoster: Oh weird. When I tried it earlier, it gave me a 404 page. That's why I updated it.

PR #141, opened 2019-01-05
keeping things consistent

scottfoster:


docs/local-development.md
```diff
@@ -12,8 +12,8 @@ The `sam local invoke` command invokes your function in a Docker container. You
 For example, given this function:
 
 ```php
-λ(function ($event) {
-    return 'Hello ' . ($event['name'] ?? 'world!');
+lambda(function (array $event) {
+    return 'Hello ' . ($event['name'] ?? 'world');
```
mnapoli: Could you update the example below as well (remove `!`)

docs/local-development.md
```diff
@@ -12,8 +12,8 @@ The `sam local invoke` command invokes your function in a Docker container. You
 For example, given this function:
 
 ```php
-λ(function ($event) {
-    return 'Hello ' . ($event['name'] ?? 'world!');
+lambda(function (array $event) {
```
mnapoli: Good point!

However you shouldn't type-hint with `array` since the variable can be anything, including `null` which is what we use in the example below.

mnapoli: My bad the event is always an array :) (I just realized that all the examples are like this.

PR #147, opened 2019-01-09
Documentation: add a "First steps" guide and links to next/previous articles

mnapoli:
The new "First steps" guide helps getting started with a simple PHP function after having installed Bref.

I think this is better to have this because it serves as a transition before getting started in all the articles, the runtimes, etc. Now there's a simple guide to follow to discover Bref from scratch.

This is based on feedback I got from proofreaders.

### DIFF
--- a/couscous.yml
+++ b/couscous.yml
@@ -27,6 +27,9 @@ menu:
             installation:
                 text: Installation
                 url: /docs/installation.html
+            first-steps:
+                text: First steps
+                url: /docs/first-steps.html
     -   section: Runtimes
         items:
             runtimes-introduction:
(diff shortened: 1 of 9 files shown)


mnapoli:
Also sorry I mixed the links "previous/next" in this PR as well, I should have separated the 2 but I started both at once…

nealio82:
Great - this is what I was getting at with my comment on https://github.com/mnapoli/bref/pull/128#issuecomment-451416134 :)

mnapoli:
@nealio82 oh that makes more sense now :)

I was reviewing the website with someone completely new to the project and it became obvious to me only at that moment that this was needed ^^

PR #150, opened 2019-01-11
Publish the runtimes in all regions

### LINKED ISSUE #152: Region support in 0.3
I see that in the new docs, for 0.3, https://bref.sh/docs/runtimes/ lists only _us-east-2_ and says that runtime region needs to match where the application is.

However on https://bref.sh/docs/installation.html it says while _us-east-1_ is the default, you can pick another region, like _eu-west-1_ to host the application closer to visitors ...

If the runtime is only available in us-east-2, does that mean the application needs to be there too? Or can that runtime be moved to e.g. eu-west-1 so that everything can be hosted there (aplication, runtime, DB etc)? Or can the runtime be in a different region?

Thanks!

mnapoli:
pinging @bubba-h57 for review

This PR is based on #149, it should be merged first.

Closes #152


runtime/experimental/Makefile
```diff
@@ -1,46 +1,23 @@
 SHELL := /bin/bash
 
-REGION ?= us-east-2
-BREF_VERSION ?= 0.3.0
-PHP_VERSION_SHORT ?= 7.2
-S3_BUCKET ?= bref-php-${REGION}
-PHP_VERSION_SHORT_UNDERSCORE = $(subst .,_,${PHP_VERSION_SHORT})
```
mnapoli: Those were not necessary anymore.

runtime/experimental/Makefile
```diff
@@ -1,46 +1,23 @@
...
 publish: distribution
-	$(eval LAYER_NAME = bref-php-default-${PHP_VERSION_SHORT_UNDERSCORE})
-	$(eval S3_KEY = ${BREF_VERSION}/php-default/${PHP_VERSION_SHORT}/php.zip)
-
-	# Upload to S3
-	#aws s3 cp export/php-${PHP_VERSION_SHORT}-cli.zip s3://${S3_BUCKET}/${S3_KEY}
-	#sleep 15
-
-	# Publish the layer
-	$(eval LAYER_VERSION = $(shell aws lambda publish-layer-version --region=${REGION} --layer-name ${LAYER_NAME} --description "bref PHP ${PHP_VERSION} Default" --license-info "MIT" --zip-file fileb://export/php-${PHP_VERSION_SHORT}-cli.zip --compatible-runtimes provided --output text --query Version))
```
mnapoli: Kept the `--zip-file` option (in the new script) which avoids going through S3 🎉

runtime/experimental/export.Dockerfile
```diff
@@ -20,15 +20,17 @@ RUN rm -rf /opt/bref/data
...
  zip -d ${PHP_ZIP_NAME}-cli.zip bref/sbin/php-fpm bin/php-fpm; \
  cp /tmp/cli.bootstrap /bootstrap; \
  chmod 755 /bootstrap; \
- zip -u ${PHP_ZIP_NAME}-cli.zip /bootstrap; \
- cp ${PHP_ZIP_NAME}.zip ${PHP_ZIP_NAME}.fpm.zip; \
- zip -d ${PHP_ZIP_NAME}.fpm.zip bref/bin/php /bin/php
+ zip -u ${PHP_ZIP_NAME}-cli.zip /bootstrap;
+
+# Create the PHP FPM layer
+RUN zip --quiet --recurse-paths ${PHP_ZIP_NAME}-fpm.zip . -x "*php-cgi"
```
mnapoli: I separated the two just for the sake of readability.

Also I fixed `php-7.2.fpm` into `php-7.2-fpm`.

Note that the FPM runtimes also contains the `php` CLI binary. I originally did that on purpose thinking that there may be scenarios where one would want to run CLI script from a PHP-FPM process. Now that I'm writing this paragraph I'm starting to doubt that 🤔 That wouldn't make sense on lambda.

I'm not considering remove the binary… Thoughts?

runtime/experimental/export.Dockerfile
```diff
@@ -20,15 +20,17 @@ RUN rm -rf /opt/bref/data
 
 WORKDIR /opt
 COPY bootstraps/cli.bootstrap /tmp/cli.bootstrap
-RUN export PHP_ZIP_NAME=/export/php-$(php -r '$version = explode(".", phpversion());printf("%d.%d", $version[0], $version[1]);'); \
- mkdir -p /export; \
+RUN export PHP_ZIP_NAME=/export/php-$(php -r '$version = explode(".", phpversion());printf("%d.%d", $version[0], $version[1]);');
+RUN mkdir -p /export; \
  rm -rf /export/*; \
- rm -rf /bootstrap; \
- zip -qr ${PHP_ZIP_NAME}.zip . -x "*php-cgi"; \
```
mnapoli: Removed the `php-7.2.zip`: wasn't used afterwards (and it was confusing having that on disk), it just means we zip the files twice but I reckon it doesn't really matter.

runtime/experimental/helpers/publish.sh
```diff
@@ -0,0 +1,40 @@
...
+    --query Version)
+
+# Add layer permissions
+aws lambda add-layer-version-permission \
+    --region=$REGION \
+    --layer-name $LAYER_NAME \
+    --version-number $LAYER_VERSION \
+    --statement-id=public \
+    --action lambda:GetLayerVersion \
+    --principal '*'
```
mnapoli: I reckon this little script could also be merged in the parent PHP script (just below) in the future…

runtime/experimental/helpers/export.sh
```diff
@@ -1,3 +0,0 @@
-#!/bin/bash
-cp /tmp/php-$(php -r 'echo phpversion();').zip /export/php-fpm-$(php -r 'echo phpversion();').zip
-cp /tmp/php-cli-$(php -r 'echo phpversion();').zip /export/php-default-$(php -r 'echo phpversion();').zip
```
mnapoli: This file wasn't used anymore, removed it.

runtime/experimental/export.Dockerfile
mnapoli: Just removed it, we can add it again if we see a good use case pop up.

mnapoli:
And now it's finally working :)

runtime/experimental/Makefile
bubba-h57: I think that is a good idea. I also noticed that seemed a lot faster on uploads. But maybe that was just wishful thinking. :-)

runtime/experimental/export.Dockerfile
bubba-h57: Aye. There might be a use case for it, but I can't think of it. I believe that in a perfect world, if http needs to run cli, it would do it by calling a different lambda job. That could either be synchronously in a blocking call or asynchronously and send it to a queue.

runtime/experimental/helpers/publish.sh
bubba-h57: I concur.

runtime/experimental/helpers/export.sh
bubba-h57: Delete all the unused things! ;-)

bubba-h57:
Looks good to me.

PR #159, opened 2019-01-11
Precise the only authorized region in deployment commands

bertrandjamin:
cf https://github.com/mnapoli/bref/issues/156#issuecomment-453672510


bertrandjamin:
I close this PR. 

The layers will be available in several regions (cf https://github.com/mnapoli/bref/pull/150).

However I still think that it should be useful to precise that the S3 bucket and the cloud formation must be in the same region.

PR #165, opened 2019-01-13
clarify the install docs

swiffer:
running `aws configure` ist the quick configuration according to the offical aws cli docs

https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html


mnapoli:
Awesome thank you!

PR #170, opened 2019-01-14
Fix typo in Symfony doc

ludofleury:


mnapoli:
Thanks!

Just a quick FYI: the symfony docs haven't been updated at all for Bref 0.3 yet. Some of the information (env variables, etc.) are still true but some are obsolete (e.g. anything related to `bref.php`, `serverless.yml`, etc.).

PR #178, opened 2019-01-17
Automatically set the Content-Length header on POST requests

### LINKED ISSUE #162: Cannot read body content of POST request
*v0.3*

I try to retrieve the body content of a POST request with an HTTP application configuration

I use this simple handler :

```php 
# index.php
<?php 

echo file_get_contents('php://input');
```
---

When testing my lambda via Api Gateway everything seems to work fine (when I manually specify the content-length) : 

image

But with a « real » HTTP call I cannot read any content from the body of my request...

image

---

*this issue come  from this comment : https://github.com/mnapoli/bref/issues/116#issuecomment-453745811*

mnapoli:
When a POST request is made without a Content-Length header the content of the request body is dropped by PHP-FPM.

This auto-adds the header when it's not set in POST requests.

Fixes #162

### DIFF
--- a/src/Runtime/PhpFpm.php
+++ b/src/Runtime/PhpFpm.php
@@ -178,6 +178,11 @@ class PhpFpm
         if (isset($headers['content-type'])) {
             $requestHeaders['CONTENT_TYPE'] = $headers['content-type'];
         }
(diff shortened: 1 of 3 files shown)


mnapoli:
Confirmed in #162, merging!

PR #179, opened 2019-01-17
remove 0.2-specific Symfony documentation & update for 0.3/SAM local

nealio82:
Updated Symfony docs to remove `.bref.yml` & `serverless.yml`, and include `template.yaml` instead


mnapoli:
Thanks!

mnapoli:
(in the future we'll try again to compile the cache before deploying, but at least here we are moving to something usable)

nealio82:
> (in the future we'll try again to compile the cache before deploying, but at least here we are moving to something usable)

yup, that's exactly what I was thinking too

Issue #161, opened 2019-01-12
Allow environment variables to be accessible in PHP-FPM processes


mnapoli:
Follow up of https://github.com/mnapoli/bref/issues/116#issuecomment-453740463 by @julienmortuaire

The `clear_env` option is enabled by default in php-fpm:

> Clear environment in FPM workers. Prevents arbitrary environment variables from reaching FPM worker processes by clearing the environment in workers before env vars specified in this pool configuration are added.

Default value seems to be yes, aka the environment var is "cleaned" by default.

We should disable this behavior to allow environment variables to be accessible in PHP-FPM processes.

mnapoli:
Fixed.

PR #181, opened 2019-01-17
update if(...) condition for determining log/cache dir in Symfony kernel

nealio82:
Forgot to make this change in the previous docs PR! Without changing the `if(...)` condition, Symfony doesn't realise it's running in Lambda or SAM Local :O

### DIFF
--- a/docs/frameworks/symfony.md
+++ b/docs/frameworks/symfony.md
@@ -81,7 +81,7 @@ The filesystem is readonly on lambdas except for `/tmp`. Because of that you nee
 public function getLogDir()
 {
     // When on the lambda only /tmp is writeable
-    if (getenv('LAMBDA_TASK_ROOT') !== false) {
+    if (getenv('SERVER_SOFTWARE') === 'bref') {
         return '/tmp/log/';
     }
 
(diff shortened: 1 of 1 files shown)


mnapoli:
I've never seen this environment variable before, I don't think it will be defined.

There is (I think) `AWS_SAM_LOCAL` to know if it's running locally in Docker or not though.

nealio82:
Oh, really!? It was definitely there when I dumped the environment. I
presumed you’d added it in as an extra touch!

mnapoli:
Oh wait I actually did 😄 

https://github.com/mnapoli/bref/blob/731c54fad810541eaad5ca80791f896ff7134b32/src/Runtime/PhpFpm.php#L164

However it is set only in PHP-FPM, and it's not an environment variable it's a special variable that PHP seems to expect: http://php.net/manual/reserved.variables.server.php

So I'm not sure it's usable here.

Are you sure about `LAMBDA_TASK_ROOT`? Because it's a vital variable, it is used in bootstrap files. Oh wait maybe you were affected by #161 so you didn't see that variable at the time? It's been fixed now.

nealio82:
Yup, looks like I might have been experiencing #161; I've just tested and I have the `LAMBDA_TASK_ROOT` available in SAM Local again.

PR #186, opened 2019-01-19
Change some of the documentation wording

deleugpn:


mnapoli:
Thank you.

Could you explain the motivation behind these changes?

mnapoli:
Thanks. The thing is that this new version looks much more complex to me. For example it includes new concepts such as:

- AWS resources
- resource definitions
- RDS/DynamoDB/VPC/Load Balancer/API Gateway
- CloudFormation changeset

The vocabulary and details are also more precise but more complex, for example for comparison:

> All of this is great except CloudFormation configuration is complex. This is where SAM helps.

compared to the proposed replacement:

> CloudFormation primary goal is to provide one unified structure to provision any AWS resource on your account. SAM on the other hand is a wraper around CloudFormation dedicated to facilitating a Lambda function deployment.

Since all of these concepts will be new to most PHP developers, and some will not be native english spearkers, I want to make it as accessible as possible.

You version is more exact and I definitely acknowledge that I take some shortcuts ^^ But this is by design.

I'd be interested though if you see things that can be simplified further (while still trying to be "correct").

In the meantime I'm sorry but I'll hold off merging it. However there are interesting parts, I'll mark them in the diff. If you have time to update the PR (or make another) to keep only those changes I'll merge it gladly!

docs/deploy.md
```diff
@@ -13,45 +13,49 @@ Let's look at how serverless applications are deployed on AWS with SAM.
...
 
-All of this is great except CloudFormation configuration is complex. This is where SAM helps.
+CloudFormation primary goal is to provide one unified structure to provision any AWS resource on your account. SAM on the other hand is a wraper around CloudFormation dedicated to facilitating a Lambda function deployment.
 
 ### SAM
 
 SAM offers a simpler configuration format. This is what you are using if you use Bref (the `template.yaml` file).
 
-The deployment process with SAM works like this:
+The deployment process with SAM is divided into three steps:
```
mnapoli: 👍 This change is great because it makes it more explicit that these steps follow each other.

We could still make it simpler I think, for example:

> Deploying with SAM is done in 3 steps:

docs/deploy.md
```diff
@@ -13,45 +13,49 @@ Let's look at how serverless applications are deployed on AWS with SAM.
...
+AWS Lambda requires your code to be stored on a S3 bucket.
 
 That means you must **create a S3 bucket** to host your application code:
 
 ```sh
 aws s3 mb s3://<bucket-name>
 ```
 
-The content of this bucket will be managed by AWS SAM. Do not use it in your application to store things like assets or uploaded files.
+Note: The content of this bucket will be managed by AWS SAM. Do not use it in your application to store things like assets or uploaded files.
```
mnapoli: :+1: let's keep this

docs/deploy.md
```diff
@@ -13,45 +13,49 @@ Let's look at how serverless applications are deployed on AWS with SAM.
...
 
 ### Deployment
 
-Before deploying make sure your code is ready to be deployed. For example remove any development files from the project and install Composer dependencies optimized for production:
+Before deploying, make sure your code is ready to be deployed. For example remove any development files from the project and install Composer dependencies optimized for production:
```
mnapoli: :+1:

docs/deploy.md
```diff
@@ -13,45 +13,49 @@ Let's look at how serverless applications are deployed on AWS with SAM.
...
+- Deploy the CloudFormation stack
 
 ## Deploying with SAM
 
 ### Setup
 
 > This step must be done only once per application.
 
-To be deployed into AWS Lambda, your code must be uploaded on AWS S3.
+AWS Lambda requires your code to be stored on a S3 bucket.
```
mnapoli: :+1:

Let's introduce the `bucket` concept in the next sentence, we can make it simpler here:

> AWS Lambda requires your code to be stored on AWS S3.

Issue #184, opened 2019-01-19
Logging to Cloudwatch from PHP-FPM


mnapoli:
We should find a solution (and document it) to log to Cloudwatch from PHP-FPM processes.

Ideally it should be without going through a file (e.g. in `/tmp`). Maybe with `error_log`? Gotta try this out.

sandrokeil:
Maybe this works like in Docker.

`php-fpm.conf`
```
[global]

error_log = /proc/self/fd/2
```

`pool.conf`
```
[www]

; Ensure worker stdout and stderr are sent to the main error log.
catch_workers_output = yes
```

mnapoli:
Lambda isn't running Docker though, they are running their own implementation called Firecracker.

mnapoli:
So far I think I have found 2 solutions.

First in both cases we need to redirect PHP-FPM's output to stderr:

```diff
-        $this->fpm = new Process(['php-fpm', '--nodaemonize', '--fpm-config', $this->configFile]);
+        $this->fpm = new Process(['php-fpm', '--nodaemonize', '--force-stderr', '--fpm-config', $this->configFile]);
        $this->fpm->setTimeout(null);
        $this->fpm->start(function ($type, $output): void {
-            if ($type === Process::ERR) {
-                echo $output;
-                exit(1);
-            }
+            echo $output;
        });
```

### Log in `stderr` through PHP-FPM

In `php-fpm.conf`:

```ini
catch_workers_output = yes
```

Then in PHP scripts we can write to `stderr`:

```php
$f = fopen('php://stderr', 'a+');
fwrite($f, 'My log');
fclose($f);
```

However this shows up like this in Cloudwatch:

```
[19-Jan-2019 15:41:24] NOTICE: fpm is running, pid 14
[19-Jan-2019 15:41:24] NOTICE: ready to handle connections
[19-Jan-2019 15:41:25] WARNING: [pool default] child 15 said into stderr: "My log"
```

That works but the log line is pretty long. Imagine a Monolog line with the default config, it would look like this in cloudwatch:

```
2019-01-19T13:52:57.167Z	c423d3a3-e975-484f-bbcb-f0d0c6865ee9	[STDERR] [19-Jan-2019 15:41:25] WARNING: [pool default] child 15 said into stderr: [2018-07-05 11:18:29] default.INFO: My log
```

That could probably be simplified however.

## Log using `error_log`

In PHP:

```php

```

[...]

PR #193, opened 2019-01-21
Build PHP 7.3 runtimes

mnapoli:
Following #184 I gave a try to compiling PHP 7.3 and it just worked, and it was pretty simple to re-add to the build.

If the published runtimes work correctly then let's support that version as well.


mnapoli:
Works fine for me 🎉 

I've added all that to the documentation and updated the templates.

Check out the runtimes here: https://runtimes.bref.sh/

PR #196, opened 2019-01-21
Update deploy.md

prolic:
rename .stack.yml => template.yaml

### DIFF
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -63,7 +63,7 @@ composer install --optimize-autoloader --no-dev
 
 ```bash
 sam package \
-    --output-template-file .stack.yaml \
+    --output-template-file template.yaml \
     --s3-bucket <bucket-name>
 ```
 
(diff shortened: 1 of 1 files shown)


mnapoli:
I think you missed an important part in the pull request template ;)

> Please explain the motivation behind the changes.
>
> In other words, explain **WHY** instead of **WHAT**.

I don't understand why you would rename files like this, it seems it's the output of `sam package` so it doesn't make sense to me.

prolic:
The `bref init` command creates a `template.yaml` file and not a `.stack.yml` file.

mnapoli:
Yes but what you changed is the `sam package` command, you changed the output file. The output file is not template.yaml, it's a different file.

nealio82:
From what I understand the `template.yaml` is _your_ domain to configure how you want things, and `.stack.yaml` is what SAM actually creates. Think of `template.yaml` in the same way as `composer.json` and `.stack.yaml` as `composer.lock`.

mnapoli:
You are mixing both commands, see https://github.com/mnapoli/bref/blob/e7f7cb7f8158bc435dc284ced9d644efe8115e9c/docs/deploy.md#deployment (copy/paste them from the documentation if you still get the error)

I'll be closing, let's reopen if we find an issue with the docs.

prolic:
thanks

Issue #160, opened 2019-01-12
Enable opcache by default


mnapoli:
Let's enable opcache by default as it brings performance improvements at no cost.

sandrokeil:
> Not that I can think of why anyone would want to disable opcache

Maybe if Lambda memory hits the limit. But it depends on running application e.g. full stack framework vs. simple cron worker. What for default `opcache.memory_consumption` should we use?

Have we experience/benchmarks how opcache works with Lambda? If the Lambda is warm, is opcache shared between requests?

bubba-h57:
Good questions. I have no experience with it (in lambda) myself, and now that you ask the question, it does seem to present some challenges. Especially given that the cache is in shared memory, we would need to do some testing to ensure that the cache persists from one request to the next. If we are caching, and then not actually using the cache, it would just be overhead.

Default `memory_consumption` is another excellent question. I think we would have to set that to a very conservative amount, assuming the function runs with the smallest amount of ram possible. Then, the developer would be free to change that setting.

It appears that testing is in order.

mnapoli:
Opcache on lambda was introduced in #14, I did some testing with API Platform (on Symfony 4).

Here is the optimized `php.ini` with the values explained in comments: https://github.com/mnapoli/bref/blob/32aac1b8f1f488a998f45a0f0d0544b8a505f51b/template/php.ini

In Bref 0.2 PHP was only really usable with 1Gb Lambdas and up so this is why I didn't care that much about the numbers. But maybe we could tone down the memory limit.

Also I've updated those configs in #155 to differentiate between a FPM-optimized config and a CLI one.

sandrokeil:
Alright, using opcache filecache is the key. :+1:

mnapoli:
Opcache is enabled, this is tested and now documented (in #206).

🎉

PR #197, opened 2019-01-21
opcache is a zend_extension

prolic:
also add the ".so" suffix


mnapoli:
Thanks!

FYI the `.so` extension is not necessary since PHP 7.2: https://wiki.php.net/rfc/load-ext-by-name, could you update the PR to remove that?

prolic:
Oh cool, I didn't recognize that. As of my latest testing, the opcache extension is always enabled, so should I change that in the docs?

mnapoli:
#160 should take care of that then, we'll straighten things out when fixing that issue.

PR #202, opened 2019-01-22
secret environment variables passed through in ci/cd pipeline

swiffer:
extend docs to include instructions on how to configure secret env vars in automated deployment

### DIFF
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -110,7 +110,7 @@ sam package ...
 sam deploy ...
 # Configure (e.g. passing secret env vars - existing env vars defined in template.yaml will be replaced!)
 aws lambda update-function-configuration
-  --function-name cake
+  --function-name <function-name>
   --environment '{"Variables":{
     "SECRET_ENV_VAR":"'"$SECRET_ENV_VAR_VALUE_FROM_CICD"'",
   }}' > /dev/null
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -108,6 +108,12 @@ composer install --optimize-autoloader --no-dev
 sam package ...
 # Deploy
 sam deploy ...
+# Configure (e.g. passing secret env vars - existing env vars from template.yaml will be replaced!)
+aws lambda update-function-configuration
+  --function-name cake
+  --environment '{"Variables":{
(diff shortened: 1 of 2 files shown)


docs/deploy.md
```diff
@@ -108,6 +108,12 @@ composer install --optimize-autoloader --no-dev
 sam package ...
 # Deploy
 sam deploy ...
+# Configure (e.g. passing secret env vars - existing env vars defined in template.yaml will be replaced!)
+aws lambda update-function-configuration
+  --function-name <function-name>
+  --environment '{"Variables":{
+    "SECRET_ENV_VAR":"'"$SECRET_ENV_VAR_VALUE_FROM_CICD"'",
+  }}' > /dev/null
```
mnapoli: This is interesting, however could you move this section to https://github.com/mnapoli/bref/blob/master/docs/environment/variables.md instead? That way the deployment documentation stays as simple as possible, and whenever you wonder about environment variables all the information will be in one place.

mrthehud: If it moves, perhaps it's worth adding a brief link to that page nearby?

swiffer: Unsure about moving this .. seeing it as a typical use Case for automated ci/cd pipelines (e.g. review apps with database connection etc.)

I can do that nevertheless

mnapoli: Right but dealing with secret environment variables should go in the "environment variables" documentation.

The deploy guide is linked from all over the place, this is a guide that most people will read when getting started with Bref. It has to cover the most basic needs.

When users want to define environment variables (public or secret) they will need to go see the "Environment variables" documentation.

If we don't do that then there are many many things we could put here, e.g. anything related to Symfony's CI/CD, Laravel as well, PHP configuration as well, etc.

mnapoli:
Thanks!

mnapoli:
@swiffer I just realized that the code example actually replaces all the existing environment variables.

Additionally I'm not sure it's a good idea at all because it is not an atomic deployment: first cloudformation will deploy the function and then (seconds later) it defines the environment variables. For a short period the variables will not exist.

Until we document a proper solution (see #7) I think it's best to revert this bit because it will create confusion and frustration for users. Sorry for not catching that sooner!

Issue #173, opened 2019-01-16
Include and enable the zip extension


mnapoli:
Issue forked from #143 

At the moment the `zlib` extension is installed and enabled but not the `zip` extension.

Explanations of the difference by @bubba-h57:

> The [zlib](http://php.net/manual/en/book.zlib.php) and the [zip](http://php.net/manual/en/book.zip.php) extensions are different, but the differences are subtle.
>
> At a high level, think of **zlib** as handling `gzip` files and **zip** handling `zip` files. 
>
> It can get confusing because you will run across documentation that purports **zlib** can handle `zip` files. **zlib** can indeed handle `zip` files if, and only if, the `zip` file uses `deflate` compression.
>
> However, while **zlib** has stuck to the `deflate` algorithm, `zip` files can support a much wider range of compression algorithms, even mixing compression in the same `zip` file, and it is not reliable to expect **zlib** to open a `zip` file.

Another issue is that `zlib` does not include the `ZipArchive` class.

@bubba-h57 if `zip` has advantages over `zlib` are there any reason we shouldn't use zip instead?

pmayet:
Thank's, I have build my own layer with extension `zip`, It works !

mnapoli:
Right, I still think for the project if there are no downsides we should probably install zip instead?

bubba-h57:
I know a lot of libraries bind to **zlib**, so we have to build it anyway. I believe pecl requires`zlib`? I reckon the easiest thing is to simply include the zip extension along with it.

mnapoli:
OK thanks for the update, let's include both then 💪

mnapoli:
@fredericgboutin-yapla note that this is a very old issue, I'm not 100% sure it's still accurate with the current Bref versions (I don't remember right now)

But if it is, happy to merge a PR indeed!

Also I'm interested to learn more how you use Bref at your company, feel free to send me an email :) (en français 👍)

aromka:
It would be good to have this included in the docs since there is no information on this available, and it's not part of extended extensions, but it looks like libzip is indeed included in the build - https://github.com/brefphp/aws-lambda-layers/blob/main/php-84/Dockerfile#L293

mnapoli:
@aromka it's there: https://bref.sh/docs/environment/php#built-in-extensions

Issue #166, opened 2019-01-13
Strip unneeded binaries from the layers


mnapoli:
@bubba-h57 I've done a `ls -la /opt/bin` in the layers and this is the result:

```
-rwx------ 1 sbx_user1051  495     6198 Jan 13 21:38 c_rehash
-rwx------ 1 sbx_user1051  495   359936 Jan 13 21:38 curl
-rwx------ 1 sbx_user1051  495     6088 Jan 13 21:38 curl-config
-rwx------ 1 sbx_user1051  495   642600 Jan 13 21:38 openssl
-rwx------ 1 sbx_user1051  495      812 Jan 13 21:38 pear
-rwx------ 1 sbx_user1051  495      833 Jan 13 21:38 peardev
-rwx------ 1 sbx_user1051  495      746 Jan 13 21:38 pecl
-rwx------ 1 sbx_user1051  495    27552 Jan 13 21:38 pg_config
-rwx------ 1 sbx_user1051  495    14816 Jan 13 21:38 phar
-rwx------ 1 sbx_user1051  495    14816 Jan 13 21:38 phar.phar
-rwx------ 1 sbx_user1051  495     2935 Jan 13 21:38 php-config
-rwx------ 1 sbx_user1051  495 12748024 Jan 13 21:38 php-fpm
-rwx------ 1 sbx_user1051  495     4531 Jan 13 21:38 phpize
-rwx------ 1 sbx_user1051  495     1695 Jan 13 21:38 xml2-config
-rwx------ 1 sbx_user1051  495    17008 Jan 13 21:38 xmlcatalog
-rwx------ 1 sbx_user1051  495    71920 Jan 13 21:38 xmllint
```

I guess we can remove everything except `php`/`php-fpm` right?

bubba-h57:
Sure.

viezel:
`curl` and `openssl` is often used by php applications. Are you sure you would be able to strip those without side effects?

mnapoli:
If we remove the binaries that shouldn't affect the libraries and the PHP extensions, right?

viezel:
hmm will `exec('curl ...')` still work without binaries?

mnapoli:
@viezel indeed that will not work. But the same could be said about any binary out there. I don't see a reason to include `curl` over any other Linux binary.

viezel:
@mnapoli agreed. Its fine for now. We can always add it again.

nealio82:
[...]

Issue #191, opened 2019-01-20
The default php.ini is incomplete


mnapoli:
The default php.ini is copied from the production one here:

https://github.com/mnapoli/bref/blob/6ce991e6f4a851d8aeb8527f9b2363486a0c3c85/runtime/php/php.Dockerfile#L389

However it is later overridden with our own php.ini that only set specific flags:

https://github.com/mnapoli/bref/blob/41cd70c3253aa8eaed58944af66e3f0e71e91d47/runtime/php/export.sh#L33

That means all the recommended defaults for production are lost. This should be fixed, maybe by appending our overridden parameters to the default file?

mnapoli:
A suggestion from #203:

> `/etc/bref/php/php.ini` -> default production php.ini from the PHP distribution
> `/etc/bref/php/conf.d/ext-*.ini` -> default config files from extensions (actually I don't know if there are any here)
> `/etc/bref/php/conf.d/bref.ini` -> Bref's optimized config for Lambda (which is currently here https://github.com/mnapoli/bref/blob/0.3/runtime/php/layers/function/php.ini)

mnapoli:
Fixed in #207, runtimes are updated.

PR #203, opened 2019-01-24
Enable Opcache and Zip. Remove unwanted binaries.

bubba-h57:
Address #160 #166 & #173

### DIFF
--- a/runtime/php/php.Dockerfile
+++ b/runtime/php/php.Dockerfile
@@ -246,6 +246,36 @@ RUN set -xe; \
  && cp xml2-config ${INSTALL_DIR}/bin/xml2-config
 
 ###############################################################################
+# LIBZIP Build
+# https://github.com/nih-at/libzip/releases
+# Needed by:
+#   - php
+ARG libzip
+ENV VERSION_ZIP=${libzip}
+ENV ZIP_BUILD_DIR=${BUILD_DIR}/zip
+
+RUN set -xe; \
+    mkdir -p ${ZIP_BUILD_DIR}/bin/; \
+# Download and upack the source code
+    curl -Ls https://github.com/nih-at/libzip/archive/rel-${VERSION_ZIP//./-}.tar.gz \
+  | tar xzC ${ZIP_BUILD_DIR} --strip-components=1
+
+# Move into the unpackaged code directory
+WORKDIR  ${ZIP_BUILD_DIR}/bin/
+
+# Configure the build
+RUN set -xe; \
+    CFLAGS="" \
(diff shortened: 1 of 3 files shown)


runtime/php/php.Dockerfile
```diff
@@ -374,10 +404,12 @@ RUN set -xe \
         --with-curl \
         --enable-exif \
         --enable-ftp \
+        --enable-opcache-file \
```
mnapoli: This line exists below already

runtime/php/php.Dockerfile
```diff
@@ -413,10 +445,22 @@ RUN set -xe; \
...
 
+RUN set -xe; \
+    mkdir -p /opt/bref/etc/php/config.d \
+ && echo  zend_extension=opcache.so > /opt/bref/etc/php/config.d/opcache.ini \
+ && echo display_errors=0 >> /opt/bref/etc/php/config.d/opcache.ini \
+ && echo dopcache.enable=1 >> /opt/bref/etc/php/config.d/opcache.ini \
+ && echo dopcache.validate_permission=0 >> /opt/bref/etc/php/config.d/opcache.ini \
+ && echo dopcache.validate_timestamps=0 >> /opt/bref/etc/php/config.d/opcache.ini \
+ && echo dpcache.memory_consumption=128 >> /opt/bref/etc/php/config.d/opcache.ini \
+ && echo dopcache.max_accelerated_files=10000 >> /opt/bref/etc/php/config.d/opcache.ini
```
mnapoli: These flags are already customized in [our php.ini](https://github.com/mnapoli/bref/blob/0.3/runtime/php/layers/function/php.ini) right? (by the way there are typos I think: `dopcache` -> `opcache`)

I think it's best to keep Bref's customization/optimization config separately (in the existing layer files). For example it's different between PHP CLI and FPM. And at least Bref's optimizations are easily identified.

bubba-h57: You want to just pull that out then?

mnapoli: Yep :+1:

In any case this will change with #191, probably the best would be to have:

/etc/bref/php/php.ini -> default production php.ini from the PHP distribution
/etc/bref/php/conf.d/ext-*.ini -> default config files from extensions (actually I don't know if there are any here)
/etc/bref/php/conf.d/bref.ini -> Bref's optimized config for Lambda (which is currently here https://github.com/mnapoli/bref/blob/0.3/runtime/php/layers/function/php.ini)

That way it's more explicit.

mnapoli:
🚀

PR #204, opened 2019-01-25
The default php.ini is incomplete

daim2k5:
Fix for Bug #191
We append now the content from our own php.ini and don't override the hole file.


mnapoli:
Thanks, unfortunately this comes with a few problems (it misses the fact that there's the FPM build right below for example).

Also we might go a different way, see https://github.com/mnapoli/bref/pull/203#discussion_r251000093

Issue #214, opened 2019-01-30
Symfony: the terminate event must run synchronously


mnapoli:
By default Symfony runs the terminate event asynchronously when using PHP-FPM.

That will not work as expected here because when PHP-FPM sends the response, Bref sends it to Lambda and the lambda instance is then freezed. The terminate event will then probably run on the next invocation.

We need to document how to disable this behavior.

j-tec:
i can just provide the logs i see in CloudWatch 

`Fatal error: Uncaught Hoa\Socket\Exception\Exception: Client returns an error (number 111): Connection refused while trying to join unix:///tmp/.bref/php-fpm.sock. in /var/task/vendor/hoa/socket/Client.php:191
Stack trace:
#0 /var/task/vendor/hoa/stream/Stream.php(219): Hoa\Socket\Client->_open('unix:///tmp/.br...', NULL)
#1 /var/task/vendor/hoa/stream/Stream.php(297): Hoa\Stream\Stream::_getStream('unix:///tmp/.br...', Object(Hoa\Socket\Client), NULL)
#2 /var/task/vendor/hoa/stream/Stream.php(178): Hoa\Stream\Stream->open()
#3 /var/task/vendor/hoa/socket/Connection/Connection.php(197): Hoa\Stream\Stream->__construct('unix:///tmp/.br...', NULL)
#4 /var/task/vendor/hoa/fastcgi/Responder.php(183): Hoa\Socket\Connection\Connection->connect()
#5 /var/task/vendor/mnapoli/bref/src/Runtime/PhpFpm.php(103): Hoa\Fastcgi\Responder->send(Array, '{\n\t"DoctrineEntity...')
#6 /opt/bootstrap(26): Bref\Runtime\PhpFpm->proxy(Array)
#7 /var/task/vendor/mnapoli/bref/src/Runtime/LambdaRuntime.php(58): {closure}(Array)
#8 /opt/bootstrap(27): Bref\Runtime\LambdaRuntime->processNextEvent(Object(Closure))`

j-tec:
The previous connection timed out, this was also the response the lambda returned. 
`REPORT RequestId: afeb2cb2-3334-45cd-bb9b-199765833828 Duration: 30005.81 ms Billed Duration: 30000 ms Memory Size: 128 MB Max Memory Used: 82 MB `

mnapoli:
OK thank you, that makes sense then. We should modify the PHP max execution time to 29 seconds then.

[...]

PR #215, opened 2019-01-30
Update the Symfony documentation for 0.3

mnapoli:
I've splitted #214 in a separate issue.


mnapoli:
The Symfony demo is updated as well: https://github.com/mnapoli/bref-symfony-demo

PR #216, opened 2019-01-30
Add tests to cover a bug where logs are forwarded in HTTP response headers

mnapoli:
PHP-FPM sends logs in the FastCGI output under the key `PHP message:`. Those lines end up in the HTTP headers. They are parsed by Hoa/FastCGI as response headers.

In the end those logs are considered as part as the HTTP response headers but they are not.

PHP-FPM should not write those in the FastCGI connection. I cannot understand why this is happening and how to fix this.

Example of a raw HTTP response sent by PHP-FPM over FastCGI:

```
PHP Message: This is a test log from error_logX-Powered-By: PHP/7.3.1
Content-Type: text/html; charset=UTF-8
Content-Length: 12
Server: Werkzeug/0.14.1 Python/3.7.2
Date: Wed, 30 Jan 2019 21:29:43 GMT

Hello world!
```

As you can see in the first line the first HTTP header (`X-Powered-By`) is mangled with the log line.

See the tests for a scenario to reproduce it.

I tried several options (not daemonizing FPM, disabling logging in stderr, etc.) but no success.

### DIFF
--- a/src/Runtime/PhpFpm.php
+++ b/src/Runtime/PhpFpm.php
@@ -106,6 +106,22 @@ class PhpFpm
 
         $responseHeaders = array_change_key_case($responseHeaders, CASE_LOWER);
 
+        /**
+         * In some cases PHP-FPM sends logs into the FastCGI connection. That isn't valid and isn't supported
+         * by our FastCGI client library, so those logs end up mangled in the HTTP headers.
+         * We can't return them to the client as:
+         * - the response is broken
+         * - it contains internal logs
+         * Until we find a better solution we simply return an empty 500 response, because since there were logs sent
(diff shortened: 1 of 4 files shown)


mnapoli:
I have pushed a temporary solution:

> In some cases PHP-FPM sends logs into the FastCGI connection. That isn't valid and isn't supported by our FastCGI client library, so those logs end up mangled in the HTTP headers.
>
> We can't return them to the client as:
> - the response is broken
> - it contains internal logs
>
> Until we find a better solution we simply return an empty 500 response, because since there were logs sent through FastCGI something wrong must have happened anyway.
>
> Those logs will also be sent to CloudWatch so debugging is possible.

mnapoli:
https://github.com/hoaproject/Fastcgi/pull/19 will help solve this properly.

PR #219, opened 2019-01-31
Add PHP 7.2 notice into documentation

claudiunicolaa:
Nice to be mentioned PHP 7.2 as a must into documentation.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -46,6 +46,7 @@ Install Bref in your project using [Composer](https://getcomposer.org/):
 ```
 composer require mnapoli/bref
 ```
+> Notice: For running the latest version of Bref, you must have installed PHP 7.2.
 
 The `bref` command line tool can now be used by running `vendor/bin/bref` in your project.
 


mnapoli:
I'm not sure I understand why recommend 7.2? The 7.3 version is good as well

claudiunicolaa:
Maybe to be recommended at least 7.2? ` For running the latest version of Bref, you must have at least version 7.2 of PHP installed.`. 

My version of PHP was 7.1 and yesterday when I tried to install Bref (following the docs)  I spent some time to understand why at `init` my generated files was `bref.php` and `serverless.yml` 😄

mnapoli:
But it shouldn't be possible to install Bref with PHP 7.1 because of the Composer requirement:

https://github.com/mnapoli/bref/blob/de82a731b8699c0178d2406c8d5949d6e7363d13/composer.json#L25

Are you sure you aren't using Bref 0.2? Maybe you installed Bref just before the 0.3.0 release was made?

mnapoli:
(in 0.3.0 there shouldn't be bref.php and serverless.yml)

mnapoli:
😢 oh you are right! I didn't think of that…

Sorry @claudiunicolaa I didn't read the diff carefully enough, I thought you were recommending 7.2 over 7.3… My bad! So yes this PR is a good change!

PR #225, opened 2019-02-04
Improve copying template files using init option

jenschude:
When trying to create a library layer it makes a lot of sense to be able to override the templates as one may want to init the lambda function with a custom bootstrap or template.yaml which already includes the library layer additionally.

Also it could be cumbersome in the future to maintain additional template files. So it copies all files from the specified template directory.

### DIFF
--- a/bref
+++ b/bref
@@ -20,8 +20,8 @@ if (file_exists(__DIR__ . '/vendor/autoload.php')) {
 
 $app = new Silly\Application('Deploy serverless PHP applications');
 
-$app->command('init [--baseTemplateDir=]', function (?string $baseTemplateDir, SymfonyStyle $io) {
-    $baseTemplateDir = $baseTemplateDir ?? __DIR__ . "/template";
+$app->command('init', function (SymfonyStyle $io) {
+    $baseTemplateDir = __DIR__ . "/template";
     $exeFinder = new ExecutableFinder();
     if (! $exeFinder->find('sam')) {
         $io->error(
--- a/bref
+++ b/bref
@@ -69,12 +69,12 @@ $app->command('init', function (SymfonyStyle $io) {
     $fs = new Filesystem;
     $rootPath = "$baseTemplateDir/$templateDirectory";
     $files = new \RecursiveIteratorIterator(new \RecursiveDirectoryIterator($rootPath));
-    /**
-     * @var RecursiveDirectoryIterator $files
-     * @var SplFileInfo $file
-     */
     foreach ($files as $file) {
-        $fileName = $files->getSubPathname();
+        /**
+         * @var RecursiveDirectoryIterator $dir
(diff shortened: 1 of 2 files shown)


mnapoli:
Hi, thank you for the pull request.

I'm not sure I understand the added value of the new option. What `bref init` does is basically copy 2 files from a template directory. If you have your own template files then it's probably better to distribute them via another way?

I don't see a use case for this (I don't mean to say it doesn't exist though), I don't know what those other templates could be and customizing the bootstrap is not something we advertise at the moment (there are no docs about it). This is why I'm not very keen on adding such an option at the moment.

Maybe you could explain what you are trying to achieve? What custom templates do you want to publish?

jenschude:
The idea is as a library author you may want to provide a way to use the library in a simple way also at AWS Lambda with more sophisticated templates, which go beyond "Hello world". Also the library may be provided a a own lambda layer.

So it would be possible to either create own commands for copying templates or to just hook up in the bref init method. No matter what I could also split this PR to the template directory override and the more generic copying of the template files.

jenschude:
As a small example how it could be used:

```
composer require <vendorName>/<lambda-library> (will also require bref)
vendor/bin/bref init --baseTemplateDir vendor/<vendorName>/lambda-templates
```

Now you would have a function with direct support and usage of the library

mnapoli:
I have not seen any library needing that at the moment.

And since we are talking about copying 1 or 2 files, I'm not sure what the added value is to support that (and create extensions point) in Bref.

Unless I see a lot of libraries needing this behavior that seems like a "what if" feature to me.

jenschude:
Okay.

I removed the init option from the PR and just left the generic approach of copying files as this may help maintaining template files in the future.

bref
```diff
@@ -66,20 +68,28 @@ $app->command('init', function (SymfonyStyle $io) {
...
+    $finder = new Finder();
+    $finder->files()->in("$baseTemplateDir/$templateDirectory");
+
+    foreach ($finder as $file) {
+        $fileName = $file->getRelativePathname();
+        $srcFile = "$baseTemplateDir/$templateDirectory/$fileName";
+        if ($fs->exists($srcFile)) {
+            $io->writeln("Creating $fileName");
+            $fs->copy($srcFile, $fileName);
+        }
```
mnapoli: Thanks that's a good solution I think!

I started looking into whether we could avoid using the Finder, since it's a new dependency in `composer.json` now.

Something like `$files = scandir("$baseTemplateDir/$templateDirectory");` should be a simple equivalent? Then you can use `$files` in the `foreach`.

jenschude: I switched to a RecursiveDirectoryIterator to avoid the Finder component as dependency

mnapoli:
Thanks!

Issue #226, opened 2019-02-05
Allow customization of vendor dir


faizanakram99:
I get this error when I try local deployment of a Symfony app with bref.

> Warning: require(/var/task/vendor/autoload.php): failed to open stream: No such file or directory in /opt/bootstrap on line 12
> PHP Fatal error:  require(): Failed opening required '/var/task/vendor/autoload.php' (include_path='.:/opt/bref/lib/php') in /opt/bootstrap on line 12

As such I can't even bootstrap my app

This error is due to the fact it tries to search for vendor directory in RootDir/vendor, in my app the vendor dir is located in other location RootDir/symfony/vendor and it is specified properly in composer.json

vendor dir is customizable (https://getcomposer.org/doc/06-config.md#vendor-dir) and bref should also fetch vendor dir from composer.json file (or environment variable)

roenschg:
I have the same problem and what you proposed @atrope unfortunatly does not help in this case. 

The problem lies in the bootstrap file of bref in the following line: https://github.com/brefphp/bref/blob/master/runtime/layers/fpm/bootstrap#L12

It tries to include the autoload.php in the root directory.

If I have the following structure in my project it will work fine:
`
composer.json
vendor/
  ...
  autoload.php
src/
  lambda1/
    index.php
  lambda2/
    index.php
`

But if I have a structure like the following:
`
src/
  lambda1/
    composer.json
    vendor/
      ...
      autoload.php
    index.php
  lambda2/
    composer.json
    vendor/
      ...
      autoload.php
    index.php
`

It will fail because it cannot find the autoload.php file.

I see the second code structure as a common case if my two lambdas are using different dependencies and I want every lambda to be as small as possible.

jenschude:
What you always can come up with is a modified bootstrap file in the root of your project.

mnapoli:
This has been fixed in [v0.5.15](https://github.com/brefphp/bref/releases/tag/0.5.15).

sc0ttdav3y:
Although this works, there's no documentation I could find that describes how to use it, so it took me about a day to understand how to get it going in the real world.  Here's what worked for me in case it helps others.

### 1. Folder structure

My folder structure includes a `services/` directory with a heterogeneous list of lambda functions using JS and PHP. 

```
services/
  test-php-api/
    handler.php
    composer.json
    lib/
      ...
    vendor/
      autoload.php
      ...
serverless.yml
```

### 2. serverless.yml

In `serverless.yml` I have the following:

```
[...]

PR #227, opened 2019-02-05
Add a self contained fpm layer

jenschude:
The FPM layer needs a lambda function which has a installed composer autoloader. It's not possible to just write a small function inside the AWS console itself. A self contained FPM layer can be bootstrapped without any autoloading mechanism inside the function itself.

I opened this PR as a discussion base as it's not yet complete. It doesn't have the functionality to publish the layer. It's maybe also better to move this in the FPM layer itself.


mnapoli:
Thank you for the pull request, this is something that we need to consider indeed.

I don't think however that we need to create a separate FPM layer for this. We could update the existing `php-fpm` layer directly.

runtime/php/export.sh
```diff
@@ -33,3 +33,23 @@ zip --quiet --recurse-paths /export/php-${PHP_SHORT_VERSION}-fpm.zip . --exclude
...
 rm bref/etc/php/conf.d/bref.ini
 rm bref/etc/php-fpm.conf
+
+# Create the PHP FPM self contained layer
+# Add files specific to this layer
+cp /layers/fpm-self/bootstrap bootstrap
+chmod 755 bootstrap
+cp /layers/fpm-self/php.ini bref/etc/php/conf.d/bref.ini
+cp /layers/fpm-self/php-fpm.conf bref/etc/php-fpm.conf
+cp /layers/fpm-self/bref/composer.json bref/composer.json
```
mnapoli: Maybe we could directly copy the `composer.json` that exists at the root of the project? That would avoid to duplicate it in the layers directory? (I'm afraid it will not up be up to date at some point)

mnapoli:
I checked and we have the same problem/need in the `console` layer.

One disadvantage of this approach as well is that it will increase the size of the layer (in short Bref and its dependencies will be installed twice on the Lambda).

Another solution could be to introduce an environment variable, e.g. `BREF_AUTOLOAD_FILE`, that users could define and point it to their custom `vendor/autoload.php` file. By default it would look in `/var/task`, but if you put your dependencies in a custom directory it could be customized with this variable.

The related issue is #226 (ping @faizanakram99 who reported it)

runtime/php/export.sh
jenschude: I decided to use a minimal composer.json in order to avoid unnecessary dependencies.

jenschude:
The size is increased at least for the fpm layer by about 2mb. As it was before above 30mb this adds not that much.

Will change the PR to directly use the FPM layer

jenschude:
> I checked and we have the same problem/need in the `console` layer.
> 
> One disadvantage of this approach as well is that it will increase the size of the layer (in short Bref and its dependencies will be installed twice on the Lambda).
> 
> Another solution could be to introduce an environment variable, e.g. `BREF_AUTOLOAD_FILE`, that users could define and point it to their custom `vendor/autoload.php` file. By default it would look in `/var/task`, but if you put your dependencies in a custom directory it could be customized with this variable.
> 
> The related issue is #226 (ping @faizanakram99 who reported it)

See #232 as a possible solution to this which avoids completely the autoloading issue

mnapoli:
Thank you @jenschude for working on this.

With #521 encapsulating runtimes will be even harder. Also, since now runtime versions are synchronized with Bref versions, and since we have a better https://runtimes.bref.sh/ (that lets you select Bref versions), I'd rather close this.

Thanks again and sorry for closing.

PR #229, opened 2019-02-07
Cleanup Bref 0.2 classes

mnapoli:
Those classes are not used anymore in Bref 0.3. Since they are not maintained and they induce dependencies we can remove them.


nealio82:
:tada:

PR #230, opened 2019-02-08
Fix typo in CLIBootstrap property name

akrabat:
Minor typo fix in CLIBootstrap: `$rumtimeAPI` => `$runtimeAPI`


mnapoli:
:+1: thanks

Let's merge this but I'll check afterwards because I think this file is actually not used at the moment, it's an experimental bootstrap (should eventually go in https://github.com/mnapoli/bref-bootstrap-benchmarks for example so that we can compare it to the others).

PR #232, opened 2019-02-08
Make console layer self contained

jenschude:
As like as the FPM layer, the console layer needs a autoloader in order to be able to instantiate the LambdaRuntime. In order to remove this need this PR copies over the LambdaRuntime inside the Layer and directly requires it. As the LambdaRuntime has no other dependencies we can completely get rid of the autoloader here which opens a lot of possibilties for console applications without actually requiring bref itself.

### DIFF
--- a/runtime/Makefile
+++ b/runtime/Makefile
@@ -10,7 +10,7 @@ export/php-%.zip:
 	cd php && make distribution
 
 # The console runtime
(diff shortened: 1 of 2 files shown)


mnapoli:
Closing per https://github.com/brefphp/bref/pull/227#issuecomment-579203765

PR #233, opened 2019-02-09
Allow to use Diactoros 2

bakura10:
Update Diactoros to allow to use version 2 (currently it is not possible to install this package with latest version of Expressive due to dependency constraints).


mnapoli:
Thanks!

PR #236, opened 2019-02-11
LambdaRuntime test

jenschude:
When trying to refactor the LambdaRuntime to reuse curl handles I realised that there is no unit test for this class. So I added a small node server which is able to return canned responses and created unit tests for the LambdaRuntime which covers a lot of edge cases.

### DIFF
--- a/src/Runtime/LambdaRuntime.php
+++ b/src/Runtime/LambdaRuntime.php
@@ -166,6 +166,7 @@ class LambdaRuntime
         $handler = curl_init($url);
         curl_setopt($handler, CURLOPT_CUSTOMREQUEST, 'POST');
         curl_setopt($handler, CURLOPT_RETURNTRANSFER, true);
+        curl_setopt($handler, CURLOPT_FAILONERROR, true);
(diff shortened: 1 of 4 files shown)


src/Runtime/LambdaRuntime.php
```diff
@@ -166,6 +166,7 @@ private function postJson(string $url, $data): void
         $handler = curl_init($url);
         curl_setopt($handler, CURLOPT_CUSTOMREQUEST, 'POST');
         curl_setopt($handler, CURLOPT_RETURNTRANSFER, true);
+        curl_setopt($handler, CURLOPT_FAILONERROR, true);
```
staabm: This constant defaults to true, why do we need it here?

jenschude: Yes it should default to true. But it didn't throw an error when returning with a 4xx status code. So I forced the fail on error flag.

See also https://github.com/mnapoli/bref/blob/master/src/Runtime/LambdaRuntime.php#L75

mnapoli:
Thanks!

Issue #118, opened 2018-12-19
Laravel queue driver


barryvdh:
So this might be out-of-scope, but with the bootstrap method etc, I was thinking of a Laravel queue driver. Where each job is just the event basically (as jobs are stored as json mostly anyways)

 - Bootstrap start the queue worker (`passthru("cd app && /opt/bin/php -c/opt/php.ini artisan queue:work lambda");`)
- In the application code, jobs are dispatched by invoking the function with the payload in the event (eg `dispatch(new SomeJob)->onQueue('lambda');`), using the aws php API.
 - Queue worker on Lambda is responsible for popping jobs, which is just calling the Lambda API endpoint to get a new event
- After job is done, output is sent to the Success/Fail endpoints

So the queue worker should always be idling until a job is incoming, which should work fast (eg. no need to Bootstrap laravel/start CLI command). And Laravel has methods for checking for memory etc + easy dispatch.

mnapoli:
Would that make more sense to use SQS as a queue driver? That way we have durability, dead letter queues, logging, monitoring, etc.

pmayet:
+1 @mnapoli with the trigger SQS/Lambda

barryvdh:
But that would trigger a new process, not using the same Laravel queue loop right?

pmayet:
Exactly !

For SQS with JSON payload : https://github.com/dusterio/laravel-plain-sqs

mnapoli:
And that makes much more sense to me. Maybe there is some misunderstanding here. Here is how it works with SQS (though I haven't tried it yet so…):

- Laravel HTTP app ("function A") sends job message to SQS
- SQS triggers another lambda ("function B") that receives the message in the `$event` variable -> it should process it

The Laravel queue loop doesn't make sense on Lambda since the code is executed on demand (when there is a message in SQS).

deleugpn:
I think mnapoli suggestion is on the right track here. AFAIK, you cannot rely on the event loop anymore. Even if you manage to get it working, it will probably be expensive as you wouldn't pay in batches of 100ms and it would still shutdown when the lambda max time reached.
Having an SQS triggering a Lambda that runs the code would be highly scalable and easy to manage. Each job get it's own instance of the lambda and finishes. 
One thing I thought that could work with bref 0.2 is this:
- Web App pushes to SQS
- SQS triggers lambda but not as a worker, just as an event. The message is still in the queue.
- Lambda runs `php artisan queue:work --once`
[...]

PR #246, opened 2019-02-14
Introduce Event object for late success or failure

deleugpn:
This Pull Request introduces an `Event` object which contains the event data, a fulfill and a reject method.
The goal is to allow developers to decide when to consider an event fulfilled or rejected without having to rely on exceptions.

For me, the primary motivation for this PR is the Laravel Worker system. I'm basically trying to achieve #118, but I don't want to reinvent an entire worker system. The strongest reason are Laravel internal events. If I don't use the internal Worker, I won't get any of the Laravel events dispatched.

The way I intend to achieve this is the following: 

- Write a custom Queue Driver for Laravel with a LambdaConnector, a LambdaQueue and a LambdaJob.
- Call the `fetchNextEvent` from LambdaQueue and build a LambdaJob and return it.
- When the Worker finishes the job, it will dispatch a `JobProcessed` event, which I can then invoke `fulfill`,
- If the Worker fails to process the job, it will dispatch a `JobFailed` event, which I can then invoke the `reject` and make the message go back to the Queue.

The main reason why this cannot be achieved with the current state of Bref is because Laravel expects the event data to be fetched and returned to the Worker class, which then handles exceptions. In other words, doing `lambda(function ($event) {...})` would never throw an exception, which means all jobs would be considered successfully processed. The job isn't really being processed inside the handle. It is just gathering the event data and returning it.

With these changes, it becomes easy to make a Lambda function that gets triggered every time a new SQS Message comes and simply runs the standard Laravel worker through `php artisan queue:work --once`. By having a Service Provider that binds the `JobProcessed` and `JobFailed` events to the `fulfill` and `reject` methods, the job will be cleared or returned to the queue accordingly.

### DIFF
--- a/src/Runtime/Client.php
+++ b/src/Runtime/Client.php
@@ -1,0 +1,132 @@
+<?php declare(strict_types=1);
+
+namespace Bref\Runtime;
+
+class Client
+{
+    /** @var string */
+    private $apiUrl;
+
+    public function __construct(string $apiUrl)
+    {
+        $this->apiUrl = $apiUrl;
+    }
+
+    /**
+     * Wait for the next lambda invocation and retrieve its data.
+     *
+     * This call is blocking because the Lambda runtime API is blocking.
+     *
+     * @see https://docs.aws.amazon.com/lambda/latest/dg/runtimes-api.html#runtimes-api-next
+     */
+    public function waitNextInvocation(): array
+    {
+        $handler = curl_init("http://{$this->apiUrl}/2018-06-01/runtime/invocation/next");
+        curl_setopt($handler, CURLOPT_FOLLOWLOCATION, true);
+        curl_setopt($handler, CURLOPT_FAILONERROR, true);
+
+        // Retrieve invocation ID
+        $invocationId = '';
+        curl_setopt($handler, CURLOPT_HEADERFUNCTION, function ($ch, $header) use (&$invocationId) {
+            if (! preg_match('/:\s*/', $header)) {
+                return strlen($header);
+            }
+            [$name, $value] = preg_split('/:\s*/', $header, 2);
+            if (strtolower($name) === 'lambda-runtime-aws-request-id') {
+                $invocationId = trim($value);
+            }
+
+            return strlen($header);
+        });
+
+        // Retrieve body
+        $body = '';
+        curl_setopt($handler, CURLOPT_WRITEFUNCTION, function ($ch, $chunk) use (&$body) {
+            $body .= $chunk;
+
+            return strlen($chunk);
+        });
+
+        curl_exec($handler);
+        if (curl_error($handler)) {
+            throw new \Exception('Failed to fetch next Lambda invocation: ' . curl_error($handler));
+        }
+        if ($invocationId === '') {
+            throw new \Exception('Failed to determine the Lambda invocation ID');
+        }
+        if ($body === '') {
+            throw new \Exception('Empty Lambda runtime API response');
+        }
+        curl_close($handler);
+
+        $event = json_decode($body, true);
+
+        return [$invocationId, $event];
+    }
+
+    /**
+     * @param mixed $responseData
+     *
+     * @see https://docs.aws.amazon.com/lambda/latest/dg/runtimes-api.html#runtimes-api-response
+     */
+    public function sendResponse(string $invocationId, $responseData): void
+    {
+        $url = "http://{$this->apiUrl}/2018-06-01/runtime/invocation/$invocationId/response";
+        $this->postJson($url, $responseData);
+    }
+
+    /**
+     * @see https://docs.aws.amazon.com/lambda/latest/dg/runtimes-api.html#runtimes-api-invokeerror
+     */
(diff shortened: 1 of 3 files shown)


mnapoli:
Thanks for the PR!

Can you confirm that what you suggests means writing a custom `bootstrap`, and as such a custom layer?

Also let's discuss 2 options to see if we can simplify this further:

- instead of adding the new class `Runtime\Client` would it help if all the methods of the `LambdaRuntime` class were public? (those you moved to `Client`)?

- your plan is to boot a new `php artisan queue:work --once` process for every invocation of the lambda right? What about instead writing a custom queue `Worker` class?

It seems the current Laravel `Worker` class was written completely for scenarios where a deamon would poll/fetch events from queue drivers. Here it's different (lambda is invoked by the queue).

For example isn't this what we need in the end:

https://github.com/laravel/framework/blob/d4f89f9ca33c70672b0c5eeb0f9523a557841e69/src/Illuminate/Queue/Worker.php#L312-L337

If we take that and rethrow exceptions in the `catch` block then it should be possible to connect that with the lambda's approach where failure is signaled via an exception.

I'm trying to preserve that behavior because AFAIK this is how lambda works in other languages (and because it is simple).

deleugpn:
> Can you confirm that what you suggests means writing a custom `bootstrap`, and as such a custom layer?

I'm not sure what this means, I haven't gotten my head wrapped around bootstrap and layers yet. What it feels for me is that with these changes, I can have my lambda's handler be `artisan queue:work --once`, which so far I **assumed** the current bootstrap would be able to handle.

> instead of adding the new class `Runtime\Client` would it help if all the methods of the `LambdaRuntime` class were public? (those you moved to `Client`)?

Yes, as long as when we instantiate the `Event` class, we send in the LambdaRuntime object in. Something like `new Event($invocationId, $data, $this);`. I thought having a dedicated class like `Client` was nicer, but that's just an implementation detail and irrelevant to the end goal.

> your plan is to boot a new `php artisan queue:work --once` process for every invocation of the lambda right? What about instead writing a custom queue `Worker` class?

I've been doing some crazy stuff with the Worker / Queue system of Laravel, and to be honest I don't think it's simpler to maintain a custom worker. My goal was to simply have a custom driver that fetches the job differently. It would be much simpler for me to maintain a fork of bref than a custom worker.
That being said, I hadn't noticed that the `process` method was public and the `handleJobException` already rethrow the exception. So what I'm thinking now is to try to simply write a custom command like `php artisan queue:lambda`. Inside this Command I can do something like 

```
lambda(function ($event)) {
    $job = new LambdaJob($event);

    $this->worker->process('lambda', $job, $options);
}
```

It looks like all I have to do to make this work is to write a `LambdaJob implements Job`, which I had to do anyway and write the `QueueLambdaCommand`.

Thank you very much for your time and insight. I'll try to set this up this weekend.

mnapoli:
Oh nice!

By the way if what you will do is this:

```php
lambda(function ($event)) {
    $job = new LambdaJob($event);

    $this->worker->process('lambda', $job, $options);
}
```

then you don't need to write a custom Artisan command. You can write the code above and run it with the [function runtime](https://bref.sh/docs/runtimes/function.html) and it should be enough?

deleugpn:
I need Laravel to bootstrap, initialize the container and Instantiate the Worker class for me. The best way to do that is to put that code inside a custom artisan command.

mnapoli:
```php
$app = require_once __DIR__ . '/bootstrap/app.php';

$kernel = $app->make(\App\Console\Kernel::class);
$kernel->bootstrap();

$worker = $app->make(Worker::class);
```

something like this maybe?

Anyway I understand if you'd rather rely on a subprocess.

deleugpn:
Laravel either starts from the `public/index.php` or `artisan` file. I didn't want to pile on to more and more starting points. Unfortunately, it seem like I'll be forced to do it the way you described. I tried to setup a Lambda function like the following:

```
  Worker:
    Properties:
      Role: !ImportValue LambdaExecutionRoleArn
      CodeUri: .
      Handler: 'artisan queue:lambda'
      Timeout: 300
      MemorySize: 1024
      Runtime: provided
      Layers:
        - arn:aws:lambda:eu-west-1:209497400698:layer:php-72:1
        - arn:aws:lambda:eu-west-1:209497400698:layer:console:1
      VpcConfig:
        SecurityGroupIds: [!ImportValue AllowAllAddressesContainerSecurityGroup]
        SubnetIds: !Split [',', !ImportValue PrivateSubnets]
```

But I got 

```
1 validation error detected: Value 'artisan queue:lambda' at 'handler' failed to satisfy constraint: Member must satisfy regular expression pattern: [^\s]+ (Service: AWSLambdaInternal; Status Code: 400; Error Code: ValidationException; Request ID: xxxxxxx)
```

I cannot have `artisan my:command` as a Handler probably because of the space character. The only way to achieve that would be having a `bootstrap` file that could parse some pre-defined character. Something like `artisan.my:command` and then the `bootstrap` file would split the `_HANDLER` variable and make `/opt/bin/php $LAMBDA_TASK_ROOT/$PREFIX $SUFFIX 2>&1`.

PR #248, opened 2019-02-18
Publish runtime layer in eu-north-1 (Stockholm)

hmazter:
eu-north-1 (Stockholm) was missing from the list of regions where the runtime layer was available, and to be able to use bref in that region I added the region to the publish script.


mnapoli:
Thank you

mnapoli:
I published all the layers (and took the opportunity to update the PHP versions to the latest patch versions).

PR #251, opened 2019-02-19
Fix #214 by disabling the `fastcgi_finish_request()` function

### LINKED ISSUE #214: Symfony: the terminate event must run synchronously

mnapoli:
This function is incompatible with Lambda's environment because when PHP-FPM sends the response, Bref sends it to Lambda and the lambda instance is then frozen. The terminate event will then probably run on the next invocation.

In Symfony this will force the `terminate` to run synchronously.

### DIFF
--- a/runtime/php/layers/fpm/php.ini
+++ b/runtime/php/layers/fpm/php.ini
@@ -18,3 +18,7 @@ opcache.memory_consumption=128
 opcache.max_accelerated_files=10000
 
 zend_extension=opcache.so
+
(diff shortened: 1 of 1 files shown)


mnapoli:
New runtimes are published.

PR #253, opened 2019-02-24
Fix #216 with a proper fix by upgrading hoa/fastcgi

mnapoli:
This is roughly a revert of the original quickfix of #216 + an upgrade of hoa/fastcgi. This upgrade fixes the root cause of the bug (in https://github.com/hoaproject/Fastcgi/pull/19), which was then released (https://github.com/hoaproject/Fastcgi/issues/23).

### DIFF
--- a/src/Runtime/PhpFpm.php
+++ b/src/Runtime/PhpFpm.php
@@ -105,22 +105,6 @@ class PhpFpm
 
         $responseHeaders = array_change_key_case($responseHeaders, CASE_LOWER);
 
-        /**
-         * In some cases PHP-FPM sends logs into the FastCGI connection. That isn't valid and isn't supported
-         * by our FastCGI client library, so those logs end up mangled in the HTTP headers.
-         * We can't return them to the client as:
-         * - the response is broken
-         * - it contains internal logs
-         * Until we find a better solution we simply return an empty 500 response, because since there were logs sent
-         * through FastCGI something wrong must have happened anyway. Those logs will also be sent to CloudWatch
-         * so debugging is possible.
-         *
-         * @see https://github.com/mnapoli/bref/pull/216
-         */
-        if (isset($responseHeaders['php message'])) {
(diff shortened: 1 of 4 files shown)


src/Runtime/PhpFpm.php
```diff
@@ -129,7 +113,7 @@ public function proxy($event): LambdaResponse
         }
         unset($responseHeaders['status']);
 
-        $responseBody = $responder->getResponseContent();
+        $responseBody = (string) $responder->getResponseContent();
```
mnapoli: This is just to make sure we have a string, the getter can return `null`.

tests/Sam/PhpFpmRuntimeTest.php
```diff
@@ -84,10 +84,9 @@ public function test warnings are logged()
     {
         $response = $this->invoke('/?warning=1');
 
-        self::assertSame(500, $response->getStatusCode(), $this->logs);
-        // Unfortunately with the temporary fix in https://github.com/mnapoli/bref/pull/216
-        // we must settle for an empty 500 response for now
-//        self::assertEquals('Hello world!', $this->getBody($response), $this->logs);
```
mnapoli: This is good: that means that a PHP warning will no longer return a blank 500 page (that was our previous workaround).

PHP now behaves just as usual.

tests/Sam/PhpFpmRuntimeTest.php
```diff
@@ -47,7 +47,7 @@ public function test error_log function()
     {
         $response = $this->invoke('/?error_log=1');
 
-        self::assertSame(500, $response->getStatusCode(), $this->logs);
+        $this->assertResponseSuccessful($response);
```
mnapoli: This is good: using `error_log()` will no longer crash the response by forcing a 500 blank page.

mnapoli:
By the way thanks @luxifer and @vdechenaux ;)

PR #257, opened 2019-03-06
Try to fix #220: recover from timeouts

### LINKED ISSUE #220: Connection refused while trying to join unix:///tmp/.bref/php-fpm.sock
I'm using Slim framework with bref v0.3 and mysql extension loaded to query on Aurora Serverless
Layer: arn:aws:lambda:eu-central-1:416566615250:layer:php-72-fpm:4
Some endpoints using mysql query are returning 502 with this error:
**Fatal error: Uncaught Hoa\Socket\Exception\Exception: Client returns an error (number 111): Connection refused while trying to join unix:///tmp/.bref/php-fpm.sock. in /var/task/vendor/hoa/socket/Client.php:191**
How i can solve this error??

mnapoli:
Recap from the beginning: with PHP-FPM if a script times out (e.g. trying to connect to an unreachable database #220 or with a `sleep(60)`) the PHP-FPM is left in a broken state.

I am trying to recover from that by restarting PHP-FPM. Restarting works correctly, and the socket file is correctly cleared.

However any new request going through the new socket file to the new PHP-FPM process end in error:

```
Hoa\Socket\Exception\BrokenPipe: Pipe is broken, cannot write data. in /var/task/vendor/hoa/socket/Connection/Connection.php:853
Stack trace:
#0 /var/task/vendor/hoa/socket/Connection/Connection.php(961): Hoa\Socket\Connection\Connection->write('\x01\x01\x00\x01\x00\x08\x00\x00\x00\x01\x00\x00\x00\x00\x00...', 1353)
#1 /var/task/vendor/hoa/fastcgi/Responder.php(228): Hoa\Socket\Connection\Connection->writeAll('\x01\x01\x00\x01\x00\x08\x00\x00\x00\x01\x00\x00\x00\x00\x00...')
#2 /var/task/src/Runtime/PhpFpm.php(113): Hoa\Fastcgi\Responder->send(Array, '')
#3 /opt/bootstrap(26): Bref\Runtime\PhpFpm->proxy(Array)
...
```

My only guess for now is that some data is left in the PHP-FPM socket… I don't know if it is possible, but to be clear:

- this is a new PHP-FPM process
- the `/tmp/.bref/php-fpm.sock` file was recreated when restarting PHP-FPM

I don't understand why the communication still fails.

Any help is welcome.

### DIFF
--- a/demo/http.php
+++ b/demo/http.php
@@ -5,10 +5,6 @@ use Psr\Http\Message\ServerRequestInterface;
 
 require __DIR__ . '/../vendor/autoload.php';
 
-/** @noinspection PhpComposerExtensionStubsInspection */
-$pdo = new \PDO('mysql:host=example.com;dbname=db');
-$pdo->exec('SELECT * FROM foo');
-
 $slim = new Slim\App;
 
 $slim->get('/', function (ServerRequestInterface $request, ResponseInterface $response) {
--- a/demo/http.php
+++ b/demo/http.php
@@ -5,6 +5,10 @@ use Psr\Http\Message\ServerRequestInterface;
 
 require __DIR__ . '/../vendor/autoload.php';
 
+/** @noinspection PhpComposerExtensionStubsInspection */
+$pdo = new \PDO('mysql:host=example.com;dbname=db');
+$pdo->exec('SELECT * FROM foo');
+
 $slim = new Slim\App;
 
 $slim->get('/', function (ServerRequestInterface $request, ResponseInterface $response) {
--- a/src/Runtime/PhpFpm.php
+++ b/src/Runtime/PhpFpm.php
@@ -118,7 +118,7 @@ class PhpFpm
             // Once the socket gets broken every following request is broken. We need to reconnect.
             $this->reconnect();
             throw new FastCgiCommunicationFailed(sprintf(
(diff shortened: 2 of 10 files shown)


demo/http.php
```diff
@@ -5,6 +5,10 @@
 
 require __DIR__ . '/../vendor/autoload.php';
 
+/** @noinspection PhpComposerExtensionStubsInspection */
+$pdo = new \PDO('mysql:host=example.com;dbname=db');
+$pdo->exec('SELECT * FROM foo');
```
mnapoli: This is a temporary test to cause a timeout and test on Lambda. To be removed.

src/Runtime/PhpFpm.php
```diff
@@ -22,7 +24,7 @@
 class PhpFpm
 {
     private const SOCKET = '/tmp/.bref/php-fpm.sock';
-    private const CONFIG = '/opt/bref/etc/php-fpm.conf';
+    private const CONFIG = '/var/task/php/conf.d/php-fpm.conf';
```
mnapoli: Temporary test, to be reverted.

mnapoli:
Wait I think I found it! I need to disconnect Hoa and recreate a client and reconnect to PHP-FPM: c2cffda3de7e8a165ba18c51d5a7906365c9eb79

Seems to work!

mnapoli:
In the end no need to restart PHP-FPM, we simply need to reconnect to it! 🎉

PR #258, opened 2019-03-08
Add a "maturity matrix" in the documentation

mnapoli:
This PR adds a new section in the introduction of the documentation: the maturity matrix.

The goal of this section is to help anyone understand what to expect for each kind of PHP application regarding:

- the simplicity of configuration and deployment
- the performances
- the confidence/reliability in production

**Review and feedback is highly appreciated**, if you have questions or want to challenge a specific rate please go ahead. I think this is an important piece of documentation to get right because it sets expectations for future users.

Click the image to see [the full preview](https://5c8237d7236d845763b1c9b3--bref-website.netlify.com/docs/#maturity-matrix):

[capture d ecran 2019-03-08 a 10 38 48](https://5c8237d7236d845763b1c9b3--bref-website.netlify.com/docs/#maturity-matrix)


gonzalovilaseca:
`API with MySQL/PostgreSQL` having a red might scare potential users, we are going to use it for api/mysql as there are a few workarounds until aws fixes the issue: eg pinging every 10 min etc.

mnapoli:
Taking the discussion from Slack here so that we have one place to discuss this:

Right I'm on the fence about this, and MySQL APIs definitely work in some cases. But at the same time it's a very strong limitation if you want to host an API… That doesn't sound too exaggerated to me to say that this scenario has "strong limitations"? Red doesn't mean that it's impossible. Here is what I say in the explanations regarding this:

> More importantly, using a VPC means cold starts get much worse: around 5 seconds. While this can be acceptable for some scenarios, for most APIs this is a deal breaker. AWS is planning on removing VPC in 2019, which will very probably change the game here.

Maybe it could be this instead:

> More importantly, using a VPC means cold starts get much worse: around 5 seconds. While this can be acceptable for some scenarios, for most APIs this is a deal breaker. *This is why we will rate it "red" for now as this is a "strong limitation".* AWS is planning on removing VPC in 2019, which will very probably change the game here.

(emphasis on the sentence I added)

mnapoli:
I added the sentence I mentioned above.

Issue #282, opened 2019-03-25
Allow user to specify region when using bref init


ngdo-pro:
When I use `vendor/bin/bref init`, I'll get a template with `us-east-1` region by default.

It was quite difficult for me to find out I could modify it in the `Layers` string in `template.yaml`.

Maybe Bref should give an opportunity to define it as a command option ?

Something like:

`vendor/bin/bref init --region=eu-west-1`

I tried to do something like that: 

```php
    if ($region !== 'us-east-1') {
        file_put_contents('template.yaml', implode('',
            array_map(function($data) use ($region) {
                return stristr($data,"                - 'arn:aws:lambda:us-east-1:209497400698:layer:php-73-fpm:1'") ? "                - 'arn:aws:lambda:".$region.":209497400698:layer:php-73-fpm:1'\n" : $data;
            }, file('template.yaml'))
        ));
    }
```

It works for HTTP, but would not for Console for example. 
I'm not great at modifying files :stuck_out_tongue:

mnapoli:
👍 	

I think it can be OK to blindly replace any `us-east-1` string with the target region (it's a very specific string, I don't see a problem happening soon).

For example: `str_replace('us-east-1', $region, file_get_contents(...))`

mnapoli:
But I think also the region could be asked by `bref init` (with the goal that `bref init` is a complete "assistant"), for example:

```
$ vendor/bin/bref init
In which AWS region do you want to deploy your application? [us-east-1]
```

We could even repeat the note added in #280 to encourage users to use `us-east-1` for their first project.

PR #285, opened 2019-03-26
add region choice during init

ngdo-pro:
Since it's not "easy" to find where we can change the region, I created an issue about making the choice during the init command.

This PR follows the recommandations made in the issue.

Related to #282


mnapoli:
The phpstan failure exists on master, I'll fix that.

Thanks!

Issue #290, opened 2019-04-01
$_ENV is not populated in PHP-FPM


lhilton:
I started by using the docs, and leaned into this for the Laravel specifics:

[https://bref.sh/docs/frameworks/laravel.html](https://bref.sh/docs/frameworks/laravel.html)

I have successfully setup bref+laravel using a hard-coded string value in place of the `$_ENV['APP_STORAGE]` value normally added to `/bootstrap/app.php` (see below for contents).

It appears that my php-fpm process is not getting the environment propped. In my `template.yaml` I am using the latest layers. Specifically for fpm `arn:aws:lambda:us-west-2:209497400698:layer:php-73-fpm:3`

It looks like this was a problem back in #161 but that looked to be resolved and closed.

To best illustrate the problem and solution, this is the top part of my `/bootstrap/app.php` with comments on the `$app->useStoragePath(...)` patch the docs have me apply:

```php
<?php

/*
|--------------------------------------------------------------------------
| Create The Application
|--------------------------------------------------------------------------
|
| The first thing we will do is create a new Laravel application instance
| which serves as the "glue" for all the components of Laravel, and is
| the IoC container for the system binding all of the various parts.
|
*/

$app = new Illuminate\Foundation\Application(
    $_ENV['APP_BASE_PATH'] ?? dirname(__DIR__)
);

/*
 * Allow overriding the storage path in production using an environment variable.
 */
// Per the docs (this does not work):
// $app->useStoragePath($_ENV['APP_STORAGE'] ?? $app->storagePath());
// When I dd($_ENV['APP_STORAGE']) I get a null, when I dd($_ENV) I get an empty []

// My temporary fix that I enable when I'm ready to deploy to AWS
$app->useStoragePath('/tmp' ?? $app->storagePath());
```

Just in case it is needed, this is my `template.yaml`

```yaml
AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31

Globals:
    Function:
        Environment:
            Variables:
                # Laravel environment variables
                APP_STORAGE: '/tmp'

Resources:
    Website:
        Type: AWS::Serverless::Function
        Properties:
            FunctionName: 'laravel-website'
            CodeUri: .
            Handler: public/index.php
            Timeout: 30 # in seconds (API Gateway has a timeout of 30 seconds)
            Runtime: provided
            Layers:
                - 'arn:aws:lambda:us-west-2:209497400698:layer:php-73-fpm:3'
            Events:
[...]

PR #291, opened 2019-04-01
Reproduce #290: `$_ENV` is not populated in PHP and PHP-FPM

mnapoli:
See #290

`$_ENV` is not populated in the PHP-FPM runtime.

Fortunately thanks to @matt-allan's [awesome article](https://mattallan.me/posts/how-php-environment-variables-actually-work/) this might be why:

> There is also an `$_ENV` superglobal.  Just like `$_SERVER` it can be disabled by removing `E` from the [`variables_order` directive](https://secure.php.net/manual/en/ini.core.php#ini.variables-order).  The default value for [development](https://github.com/php/php-src/blob/master/php.ini-development#L627) and [production](https://github.com/php/php-src/blob/b167dbe3ae9201e725d9f02817849e65bbb50c02/php.ini-production#L627) is `GPCS`, meaning `$_ENV` is most likely empty on your server.

I am a bit confused and unsure what to do: `$_ENV` is not populated per PHP's default configuration, yet it is relied upon by some frameworks, especially Laravel (see #290).

Should we populate `$_ENV` by overloading PHP's default configuration?

### DIFF
--- a/runtime/php/layers/fpm/php.ini
+++ b/runtime/php/layers/fpm/php.ini
@@ -19,6 +19,13 @@ opcache.max_accelerated_files=10000
 
 zend_extension=opcache.so
 
+; This directive determines which super global arrays are registered when PHP
+; starts up. G,P,C,E & S are abbreviations for the following respective super
+; globals: GET, POST, COOKIE, ENV and SERVER.
+; We explicitly populate all variables else ENV is not populated by default.
+; See https://github.com/mnapoli/bref/pull/291
+variables_order="EGPCS"
+
 ; The lambda environment is not compatible with fastcgi_finish_request
 ; See https://github.com/mnapoli/bref/issues/214
 disable_functions=fastcgi_finish_request
--- a/runtime/php/layers/function/php.ini
+++ b/runtime/php/layers/function/php.ini
@@ -24,3 +24,10 @@ opcache.memory_consumption=128
 opcache.max_accelerated_files=10000
 
 zend_extension=opcache.so
+
+; This directive determines which super global arrays are registered when PHP
+; starts up. G,P,C,E & S are abbreviations for the following respective super
+; globals: GET, POST, COOKIE, ENV and SERVER.
+; We explicitly populate all variables else ENV is not populated by default.
+; See https://github.com/mnapoli/bref/pull/291
+variables_order="EGPCS"
--- a/tests/Sam/Php/function.php
+++ b/tests/Sam/Php/function.php
@@ -11,6 +11,14 @@ lambda(function (array $event) {
         return ini_get_all(null, false);
     }
 
+    if ($event['env'] ?? false) {
+        return [
+            '$_ENV' => $_ENV['FOO'] ?? null,
+            '$_SERVER' => $_SERVER['FOO'] ?? null,
+            'getenv' => getenv('FOO'),
+        ];
+    }
+
     if ($event['stdout'] ?? false) {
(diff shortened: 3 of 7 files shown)


nealio82:
Considering that one of Bref's goals is to 'provide simple and familiar solutions' and that the majority of Bref's users will be using one framework or another, I think we probably have a strong argument to populate `$_ENV` (provided it doesn't have a negative effect on working with other frameworks' env handling).

Is this related to the Laravel `5.8.5` thing where they changed their approach to environment loading?

mnapoli:
👍 	

> Is this related to the Laravel 5.8.5 thing where they changed their approach to environment loading?

No I think that's a separate problem.

nealio82:
If it's a different problem, I'm confused. How come the documented method worked before but doesn't work now?

mnapoli:
That's a good point, the [demo](https://github.com/mnapoli/bref-laravel-demo) references master so I guess the layers changed in-between.

Anyway it makes sense to enable this. I'll get on this.

lhilton:
> yet it is relied upon by some frameworks, especially Laravel

If I understand your assumption, you believe that Laravel (and others) require E to be loaded? This isn't quite correct. It will integrate values stored in $_ENV (if any) along with other configuration data, but I run my local machines without it.

The challenge here is that Laravel's preferred way to handle this environment configuration is with the `.env` file using the package [PHP dotenv](https://github.com/vlucas/phpdotenv). This is done much later in Laravel's boot cycle than the processing done in `bootstrap/app.php`. That is why the superglobal is such a reasonable solution as it allows this to be configured very high up in the execution lifecycle. This is how I traditionally handle other issues like this in the Docker / Laravel world.

My vote would be to overload the config and enable the feature.

mnapoli:
Laravel uses `$_ENV` directly in a few places that are not related to the `env()` helper. The original bug was related to [this](https://github.com/laravel/laravel/blob/ae44ee0059f6218aab26de5551b19f70329f8832/bootstrap/app.php#L15)

```php
$app = new Illuminate\Foundation\Application(
    $_ENV['APP_BASE_PATH'] ?? dirname(__DIR__)
);
```

After searching a bit I can only find this place 🤔 Maybe it's simply something to fix in Laravel.

lhilton:
> Maybe it's simply something to fix in Laravel.

Or even a different way of hooking this configuration change into Laravel. I'll dig into the load order and see if I can suggest a different approach for the documentation that perhaps depends on `.env` or something similar.

matt-allan:
Laravel might accept a PR to use getenv instead.  The Dotenv component uses getenv now so there isn’t much to gain by avoiding it here anymore.

deleugpn:
Although I'm a Laravel fan, I really don't have enough knowledge to give a good opinion here. I just want to add this information to the table: 
https://github.com/laravel/framework/blob/5.8/src/Illuminate/Foundation/Application.php#L886

When I ran Laravel on Lambda, the `/bootstrap/cache.php` need to be writable, which is read-only on Lambda. I had to edit the original `artisan` file and write `$_ENV['APP_SERVICES_CACHE'] = '/tmp/services.php`.

mnapoli:
@deleugpn good point. It doesn't seem like a huge burden to populate `$_ENV` honestly, and that will cover it for all frameworks out there.

mnapoli:
The build is failing but this is correct: the runtime tests are still using the current runtimes that include the bug.

I need to merge this, build new releases and update the tests.

Issue #292, opened 2019-04-03
Enable bcmath extension


tiagobrito:
`bcmath` extension is currently being used in important libraries such as `google/protobuf` (directly) and indirectly by other relevant libraries eg: `google/cloud-vision` and more.

@mnapoli What you think about adding it to the list of enabled extensions?

deleugpn:
`laravel/telescope` also uses it.

mnapoli:
New runtimes include bcmath by default!

PR #289, opened 2019-04-01
Enable SOAP

jsamouh:
issue #283 

enabled soap by default

PR #293, opened 2019-04-03
#292 Enable bcmath extension

tiagobrito:
More context here #292


mnapoli:
Thanks, just like in #289 could you try to compile (`cd runtime && make layers`) and tell us the size of the layers/zip files that are created?

tiagobrito:
Size | File name
------------ | -------------
700B | console.zip
31M | php-72-fpm.zip
24M | php-72.zip
31M | php-73-fpm.zip
24M | php-73.zip

PR #297, opened 2019-04-05
Add content length header for more http methods

timoschaefer:
**What**
Allow more than just POST requests to automatically add the Content-Length header within the PHP-FPM runtime. As per the HTTP spec, this should include at least `POST`, `PUT`, and `PATCH` - semantics for all other HTTP methods are undefined except for `TRACE`, which explicitly forbids the usage of a request body.

**Why**
Content-Length auto-detection appears to be a common feature at least in NGINX, which is the web server I'm used to in an environment with a server. It's simply better for DX and allows us to write RESTful services using Bref.

@mnapoli One question regarding the tests: The interface currently hardcodes POST in a few method names, what would be your preferred approach to add the other newly supported http methods? I'll add the tests to cover all methods and cases (Content-Length detection, base64 decoding) after that question was answered. :)

TODO:
- [x] Implement changes in PhpFpm runtime
- [x] Get feedback on how tests should be implemented
- [x] Implement tests

### DIFF
--- a/src/Runtime/PhpFpm.php
+++ b/src/Runtime/PhpFpm.php
@@ -200,8 +200,6 @@ class PhpFpm
         $headers = $event['headers'] ?? [];
         $headers = array_change_key_case($headers, CASE_LOWER);
 
-        $serverName = $headers['host'] ?? 'localhost';
-
         $requestHeaders = [
             'GATEWAY_INTERFACE' => 'FastCGI/1.0',
             'REQUEST_METHOD' => $event['httpMethod'],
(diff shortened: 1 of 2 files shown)


mnapoli:
Awesome, thanks for that!

Regarding the tests, you could do the same thing as what was done in #178. You could even keep the existing tests (there are 2) and run them against `POST`, `PUT` and `PATCH` using a PHPUnit dataprovider ?

deleugpn:
Is this what will fix PUT requests being received with empty body at the moment?

mnapoli:
@deleugpn it should yes.

timoschaefer:
Please ignore the amount of commits for a pull request this simple, I'm having issues with my local environment right now. 🙈

mnapoli:
No worries! Thank you!

deleugpn:
@mnapoli do I just have to update composer or does this require a new layer?

timoschaefer:
`composer update` should be sufficient, since your Lambda uses your vendor folder for the FPM runtime as of right now.

deleugpn:
Thank you @timoschaefer. This worked great!

image

PR #298, opened 2019-04-05
reuse curl handler

jenschude:
**What**

Reuses the curl connection handle in the lambda runtime.

**Why**

curl_init creates always a new connection. As we are constantly talking to the same server this could be avoided. This will save some milliseconds on each recurring invocation.


mnapoli:
Thanks!

PR #308, opened 2019-04-16
Add a new `bref deployment` command to debug failed deployments

mnapoli:
If an error occurs when deploying with SAM we need to look into the *Events* tab of the stack in the CloudFormation dashboard.

This works but isn't friendly and practical. This PR introduces the following Bref command to list the logs from the latest deployments and understand what went wrong:

```bash
vendor/bin/bref deployment <stack-name>
```

Only events from the last 24 hours are displayed. We could make it configurable in the future (PRs welcome).

Preview of the output:

Capture d’écran 2019-04-16 à 20 47 21

### DIFF
--- a/bref
+++ b/bref
@@ -2,6 +2,7 @@
 <?php
 declare(strict_types=1);
 
+use Aws\CloudFormation\CloudFormationClient;
 use Bref\Lambda\InvocationFailed;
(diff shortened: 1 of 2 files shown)


nealio82:
:heart:

PR #311, opened 2019-04-18
Pass the Lambda context to handlers

amne:
We would like access to the aws-request-id header that is returned by the runtime API.

Our specific use case is that we use the request id to correlate log entries generated by this lambda invocation. Initially we generated our own correlation id but we noticed we can't link them to the cloudwatch entries but we can link them if we use the lambda request id as the correlation id.


mnapoli:
That's really interesting!

I think we can do without the `ContextAwareHandler` interface. Indeed it is possible to invoke a function with "extra" arguments. If the function doesn't use them or even declare them they will simply be ignored.

src/Runtime/Context.php
```diff
@@ -0,0 +1,62 @@
...
+    public function getAwsRequestId(): string
+    {
+        return $this->awsRequestId;
+    }
+
+    public function setAwsRequestId(string $awsRequestId): Context
+    {
+        $this->awsRequestId = $awsRequestId;
+        return $this;
+    }
```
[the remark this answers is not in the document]
ubermuda: To encapsulate the complexity of collecting the context data and then passing it to a constructor, what do you think of using a `ContextBuilder` to collect on the go and then use it to build a valid `Context`?

The code in `LambdaRuntime` could then look something like:

```php
        $contextBuilder = new ContextBuilder();
        curl_setopt($this->handler, CURLOPT_HEADERFUNCTION, function ($ch, $header) use ($contextBuilder) {
            $contextBuilder->addFromHeader($header);
        });

        $context = $contextBuilder->build();
```

(not sure where it's best to put the header parsing, but you get the idea)

src/Runtime/Context.php
```diff
@@ -0,0 +1,62 @@
...
+
+    public function getDeadlineMs(): int
```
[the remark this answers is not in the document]
amne: It looks like all documented context objects implement getRemainingTimeInMillis() instead of just providing the deadline timestamp. We should follow that.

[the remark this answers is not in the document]
amne: Strike that. All except Go of course.
Regardless, I did implement getRemainingTimeInMillis in the end as there is no "clean" way to get a milliseconds timestamp in PHP. I chose the simpler intval(microtime(true)*1000) and substract it from AWS's deadline.

src/Runtime/LambdaRuntime.php
```diff
@@ -131,16 +145,19 @@ private function waitNextInvocation(): array
...
-        if ($invocationId === '') {
-            throw new \Exception('Failed to determine the Lambda invocation ID');
-        }
         if ($body === '') {
             throw new \Exception('Empty Lambda runtime API response');
         }
 
+        $context = $contextBuilder->buildContext();
+
+        if ($context->getAwsRequestId() === '') {
```
ubermuda: Is a Context with an empty aws request id valid? Maybe this check should be moved in the ContextBuilder? (same question for the other properties of the context, what is the minimal set of data that constitutes a valid context?)

src/Runtime/LambdaRuntime.php
```diff
@@ -104,14 +109,23 @@ private function waitNextInvocation(): array
...
-        curl_setopt($this->handler, CURLOPT_HEADERFUNCTION, function ($ch, $header) use (&$invocationId) {
+        $contextBuilder = new ContextBuilder;
+        curl_setopt($this->handler, CURLOPT_HEADERFUNCTION, function ($ch, $header) use ($contextBuilder) {
             if (! preg_match('/:\s*/', $header)) {
                 return strlen($header);
             }
             [$name, $value] = preg_split('/:\s*/', $header, 2);
             if (strtolower($name) === 'lambda-runtime-aws-request-id') {
-                $invocationId = trim($value);
+                $contextBuilder->setAwsRequestId(trim($value));
```
jsamouh: Maybe do a switch/case to prevent unnecessary condition call or if/elseif.

if you don't do that, at least, resolve strtolower($name) and trim($value) just one time.

src/Runtime/LambdaRuntime.php
amne: context can be completely empty but that piece of code expects a non-empty invocation id. It's just a coincidence I'm refetching it from the context instead of keeping around in an $invocationId variable. I could do revert back to using a $invocationId variable in that piece of code and build the context last just before return

src/Runtime/LambdaRuntime.php
amne: they're not nested ifs but the strtolower/trim is good catch. too much copy/pasta going on

src/Runtime/LambdaRuntime.php
mnapoli: It might ideally be better indeed to forbid having an invalid Context around, but honestly this should never happen and an exception will be thrown if it ever happens. Let's not block this PR longer than necessary.

Thank you for updating your PR @amne.

PR #318, opened 2019-04-25
#316 Add a test to cover a bug with large responses 

### LINKED ISSUE #316: Random FastCgiCommunicationFailed errors when using php-fpm layer
Hi @mnapoli !

We tried to use bref and php-fpm layer but we find it very unstable. We tried to make it working with Symfony4, ApiPlatform as well as bare minimal code that sends one header echoes a response. In any case (running locally using sam and in aws environment) we observed very strange and non deterministic behaviours. Instead of getting a real response we were getting 502 response code from Sam with a response of:

```json
{
"message": "Internal server error"
}
```

The console running lambda using Sam was printing the following output:

```
2019-04-24 12:35:12 Invoking index.php (provided)
2019-04-24 12:35:12 arn:aws:lambda:eu-west-1:209497400698:layer:php-73-fpm:5 is already cached. Skipping download
2019-04-24 12:35:12 Requested to skip pulling images ...

2019-04-24 12:35:12 Mounting /Users/kwnuk/Repositories/thirdbridge/bref as /var/task:ro,delegated inside runtime container
START RequestId: 52fdfc07-2182-154f-163f-5f0f9a621d72 Version: $LATEST
Fatal error: Uncaught Bref\Runtime\FastCgiCommunicationFailed: Error communicating with PHP-FPM to read the HTTP response. A common root cause of this can be that the Lambda (or PHP) timed out, for example when trying to connect to a remote API or database, if this happens continuously check for those! Bref will reconnect to PHP-FPM to clean things up. Original exception message: Hoa\Fastcgi\Exception\Exception Bad request (not a well-formed FastCGI request). in /var/task/vendor/mnapoli/bref/src/Runtime/PhpFpm.php:122
Stack trace:
#0 /opt/bootstrap(30): Bref\Runtime\PhpFpm->proxy(Array)
#1 /var/task/vendor/mnapoli/bref/src/Runtime/LambdaRuntime.php(85): {closure}(Array)
#2 /opt/bootstrap(31): Bref\Runtime\LambdaRuntime->processNextEvent(Object(Closure))
#3 {main}END RequestId: 52fdfc07-2182-154f-163f-5f0f9a621d72
REPORT RequestId: 52fdfc07-2182-154f-163f-5f0f9a621d72	Init Duration: 166.87 ms	Duration: 46.38 ms	Billed Duration: 100 ms	Memory Size: 1024 MB	Max Memory Used: 42 MB
{
  "errorType": "Bref\\Runtime\\FastCgiCommunicationFailed",
  "errorMessage": "Error communicating with PHP-FPM to read the HTTP response. A common root cause of this can be that the Lambda (or PHP) timed out, for example when trying to connect to a remote API or database, if this happens continuously check for those! Bref will reconnect to PHP-FPM to clean things up. Original exception message: Hoa\\Fastcgi\\Exception\\Exception Bad request (not a well-formed FastCGI request).",
  "stackTrace": [
    "#0 /opt/bootstrap(30): Bref\\Runtime\\PhpFpm-\u003eproxy(Array)",
    "#1 /var/task/vendor/mnapoli/bref/src/Runtime/LambdaRuntime.php(85): {closure}(Array)",
    "#2 /opt/bootstrap(31): Bref\\Runtime\\LambdaRuntime-\u003eprocessNextEvent(Object(Closure))",
    "#3 {main}"
  ]
}
2019-04-24 12:35:14 Function returned an invalid response (must include one of: body, headers or statusCode in the response object). Response received:
2019-04-24 12:35:14 127.0.0.1 - - [24/Apr/2019 12:35:14] "GET / HTTP/1.1" 502 -
```

Another weird behaviour we observed was cutting the json response in a middle and adding `[] }` at the end of it (sill not being able to produce correct json object). An example output looked like that:

```
{"foo": "bar", "fizz": "bizz", "someth []}
```

These were responses that were echoed directly. We didn't use database, or anything that could potentially delay the response or cause any timeout. We were getting errored responses within <1sec.

The minimal repo that allows to reproduce an issue is available here:

https://github.com/third-bridge/bref-example

Do you have any clue why is it failing? It's almost a default configuration generated by bref. We tried with different layers (v4 and v5, php-fpm 7.2 and 7.3) still getting the same results.

mnapoli:
The goal of this PR is to fix #316 

This PR currently contains only a test that reproduces the bug.

### DIFF
--- a/tests/Runtime/PhpFpm/big-json.json
+++ b/tests/Runtime/PhpFpm/big-json.json
@@ -1,0 +1,10982 @@
+[
+    {
+        "_id": "5cc0264e5af745cd7357963b",
+        "index": 0,
+        "guid": "6186252d-99d1-4a14-85d6-582f8722acd4",
+        "isActive": true,
+        "balance": "$3,251.06",
+        "picture": "http://placehold.it/32x32",
+        "age": 29,
+        "eyeColor": "green",
+        "name": "Hodges Norris",
+        "gender": "male",
+        "company": "KINETICUT",
+        "email": "hodgesnorris@kineticut.com",
+        "phone": "+1 (905) 562-2098",
+        "address": "315 Schroeders Avenue, Itmann, Nevada, 1831",
+        "about": "Duis consequat in magna irure irure pariatur fugiat nostrud proident occaecat Lorem. Aute id qui sunt cupidatat incididunt sit est amet deserunt ad non in. Dolor magna pariatur commodo aliqua nisi laboris nulla esse amet do sint.\r\n",
+        "registered": "2016-05-11T04:46:37 -02:00",
+        "latitude": 4.670759,
+        "longitude": 161.850372,
+        "tags": [
+            "est",
+            "eiusmod",
+            "proident",
+            "qui",
+            "deserunt",
+            "reprehenderit",
+            "qui"
+        ],
+        "friends": [
+            {
+                "id": 0,
+                "name": "Sherman Mccoy"
+            },
+            {
+                "id": 1,
+                "name": "Miles Delaney"
+            },
+            {
+                "id": 2,
+                "name": "Myrtle Lara"
+            }
+        ],
+        "greeting": "Hello, Hodges Norris! You have 2 unread messages.",
+        "favoriteFruit": "strawberry"
+    },
+    {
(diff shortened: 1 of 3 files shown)


mnapoli:
🤔 the build shouldn't be green.

Issue #274, opened 2019-03-18
Unrecoverable state on Socket Exception


atrope:
When Lambda enters in the catch section of the responder exception, it will never  recover until AWS decides to kill it.

```
        try {
            $responder->send($requestHeaders, $requestBody);
        } catch (HoaFastCgiException|HoaSocketException $e) {
}
```

I am investigating what we can do in order to fix this issue.

I Already tried:

1 - killExistingFpm - it does kill, but the process stays as defunct and waitUntilStopped always return timeout error exception.
2 - killExistingFpm removing the waitUntilStopped, this "works" as fpm again loads, but the socket never attach to it..
3 - used shell_exec to kill the process and results the same as 1

Will update here when I have  more results.. Feel free to drop your penny here.

mnapoli:
Thanks for the report. FYI there were 2 recent releases trying to address timeouts like these (and recovering from them): https://github.com/mnapoli/bref/releases It might be worth to check those pull requests.

Could you make sure you are using the latest Bref version as well as the latest runtimes? (https://runtimes.bref.sh/)

atrope:
Hi @mnapoli  i did, `killExistingFpm` was added in #272.
The runtime i am using ours(we need custom extensions), but with the latest code

mnapoli:
👍 	

> The runtime i am using ours(we need custom extensions), but with the latest code

FYI you can use Bref's official layers and only add your compiled extension via an additional layers (but maybe that's already what you are doing).

atrope:
@mnapoli most extensions yes, but i need to have PHP compiled with SOAP support (google ads sdk) and redis compiled with igbinary support also, with this i need to have my own runtime running ( i could only overwrite those files, but i think its easier having everything in a single layer)

atrope:
```
Fatal error: Uncaught Bref\Runtime\FastCgiCommunicationFailed: Error communicating with PHP-FPM to read the HTTP response. A common root cause of this can be that the Lambda (or PHP) timed out, for example when trying to connect to a remote API or database, if this happens continuously check for those! Bref will reconnect to PHP-FPM to clean things up. Original exception message: Hoa\Fastcgi\Exception\Exception Bad request (not a well-formed FastCGI request). in /opt/var/task/app/library/vendor/mnapoli/bref/src/Runtime/PhpFpm.php:120
Stack trace:
#0 /opt/bootstrap(33): Bref\Runtime\PhpFpm->proxy(Array)
[...]

PR #319, opened 2019-04-25
Migrate from hoa/fastcgi to hollodotme/fast-cgi-client

### LINKED ISSUE #316: Random FastCgiCommunicationFailed errors when using php-fpm layer

mnapoli:
The goal is to fix #316 and changing the underlying library seems to solve the problem. This PR includes #318.

https://github.com/hollodotme/fast-cgi-client seems to be more maintained and used as well.

This may also solve #274 🤞 

ping @hollodotme for information :)

### DIFF
--- a/tests/Runtime/PhpFpmTest.php
+++ b/tests/Runtime/PhpFpmTest.php
@@ -3,7 +3,7 @@
 namespace Bref\Test\Runtime;
 
 use Bref\Http\LambdaResponse;
-use Bref\Runtime\FastCgiCommunicationFailed;
+use Bref\Runtime\FastCgi\FastCgiCommunicationFailed;
 use Bref\Runtime\PhpFpm;
 use Bref\Test\HttpRequestProxyTest;
 use PHPUnit\Framework\TestCase;
@@ -39,6 +39,8 @@ class PhpFpmTest extends TestCase implements HttpRequestProxyTest
                 'PATH_INFO' => '/hello',
                 'REQUEST_METHOD' => 'GET',
                 'QUERY_STRING' => '',
+                'CONTENT_LENGTH' => '0',
+                'CONTENT_TYPE' => 'application/x-www-form-urlencoded',
             ],
             'HTTP_RAW_BODY' => '',
         ]);
@@ -72,6 +74,8 @@ class PhpFpmTest extends TestCase implements HttpRequestProxyTest
                 'PATH_INFO' => '/hello',
                 'REQUEST_METHOD' => 'GET',
                 'QUERY_STRING' => 'foo=bar&bim=baz',
+                'CONTENT_LENGTH' => '0',
+                'CONTENT_TYPE' => 'application/x-www-form-urlencoded',
             ],
             'HTTP_RAW_BODY' => '',
         ]);
@@ -108,6 +112,8 @@ class PhpFpmTest extends TestCase implements HttpRequestProxyTest
                 'PATH_INFO' => '/',
                 'REQUEST_METHOD' => 'GET',
                 'QUERY_STRING' => 'vars%5Bval1%5D=foo&vars%5Bval2%5D%5B0%5D=bar',
+                'CONTENT_LENGTH' => '0',
+                'CONTENT_TYPE' => 'application/x-www-form-urlencoded',
             ],
             'HTTP_RAW_BODY' => '',
         ]);
@@ -135,6 +141,8 @@ class PhpFpmTest extends TestCase implements HttpRequestProxyTest
                 'REQUEST_METHOD' => 'GET',
                 'QUERY_STRING' => '',
                 'HTTP_X_MY_HEADER' => 'Hello world',
+                'CONTENT_LENGTH' => '0',
+                'CONTENT_TYPE' => 'application/x-www-form-urlencoded',
             ],
             'HTTP_RAW_BODY' => '',
         ]);
@@ -452,6 +460,8 @@ Content-Disposition: form-data; name=\"delete[categories][]\"\r
                 'REQUEST_METHOD' => 'GET',
                 'QUERY_STRING' => '',
                 'HTTP_COOKIE' => 'tz=Europe%2FParis; four=two+%2B+2; theme=light',
+                'CONTENT_LENGTH' => '0',
+                'CONTENT_TYPE' => 'application/x-www-form-urlencoded',
             ],
             'HTTP_RAW_BODY' => '',
         ]);
@@ -577,6 +587,8 @@ Year,Make,Model
                 'QUERY_STRING' => '',
                 'SERVER_NAME' => 'www.example.com',
                 'HTTP_HOST' => 'www.example.com',
+                'CONTENT_LENGTH' => '0',
+                'CONTENT_TYPE' => 'application/x-www-form-urlencoded',
             ],
             'HTTP_RAW_BODY' => '',
         ]);
@@ -599,6 +611,8 @@ Year,Make,Model
                 'PATH_INFO' => '/',
                 'REQUEST_METHOD' => 'PUT',
                 'QUERY_STRING' => '',
+                'CONTENT_LENGTH' => '0',
+                'CONTENT_TYPE' => 'application/x-www-form-urlencoded',
             ],
             'HTTP_RAW_BODY' => '',
         ]);
@@ -621,6 +635,8 @@ Year,Make,Model
                 'PATH_INFO' => '/',
                 'REQUEST_METHOD' => 'PATCH',
                 'QUERY_STRING' => '',
+                'CONTENT_LENGTH' => '0',
+                'CONTENT_TYPE' => 'application/x-www-form-urlencoded',
             ],
             'HTTP_RAW_BODY' => '',
         ]);
@@ -643,6 +659,8 @@ Year,Make,Model
                 'PATH_INFO' => '/',
                 'REQUEST_METHOD' => 'DELETE',
                 'QUERY_STRING' => '',
+                'CONTENT_LENGTH' => '0',
+                'CONTENT_TYPE' => 'application/x-www-form-urlencoded',
             ],
             'HTTP_RAW_BODY' => '',
         ]);
@@ -665,6 +683,8 @@ Year,Make,Model
                 'PATH_INFO' => '/',
                 'REQUEST_METHOD' => 'OPTIONS',
                 'QUERY_STRING' => '',
+                'CONTENT_LENGTH' => '0',
+                'CONTENT_TYPE' => 'application/x-www-form-urlencoded',
             ],
             'HTTP_RAW_BODY' => '',
         ]);
--- a/tests/Runtime/PhpFpmTest.php
+++ b/tests/Runtime/PhpFpmTest.php
@@ -745,6 +745,19 @@ Year,Make,Model
         }
     }
 
+    /**
+     * @see https://github.com/mnapoli/bref/issues/316
+     */
+    public function test large response()
+    {
+        // Repeat the test 5 times because some errors are random
+        for ($i = 0; $i < 5; $i++) {
+            $response = $this->get('large-response.php')->toApiGatewayFormat();
+
+            self::assertStringEqualsFile(__DIR__ . '/PhpFpm/big-json.json', $response['body']);
+        }
+    }
+
     private function assertGlobalVariables(array $event, array $expectedGlobalVariables): void
     {
         $this->startFpm(__DIR__ . '/PhpFpm/request.php');
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
         "symfony/process": "^3.3|^4.0",
         "aws/aws-sdk-php": "^3.44",
         "psr/http-message": "^1.0",
-        "hollodotme/fast-cgi-client": "^2.6"
+        "hollodotme/fast-cgi-client": "^2.7"
     },
     "require-dev": {
         "phpunit/phpunit": "^6.5",
--- a/composer.json
+++ b/composer.json
@@ -29,9 +29,8 @@
         "symfony/filesystem": "^3.1|^4.0",
         "symfony/process": "^3.3|^4.0",
         "aws/aws-sdk-php": "^3.44",
-        "hoa/fastcgi": "^3.19.02.19",
-        "hoa/socket": "1.17.05.16",
-        "psr/http-message": "^1.0"
+        "psr/http-message": "^1.0",
+        "hollodotme/fast-cgi-client": "^2.6"
     },
     "require-dev": {
         "phpunit/phpunit": "^6.5",
--- a/src/Runtime/FastCgi/FastCgiCommunicationFailed.php
+++ b/src/Runtime/FastCgi/FastCgiCommunicationFailed.php
@@ -1,6 +1,6 @@
 <?php declare(strict_types=1);
 
-namespace Bref\Runtime;
+namespace Bref\Runtime\FastCgi;
 
 /**
  * There was an error while communicating with FastCGI.
--- a/src/Runtime/FastCgi/FastCgiRequest.php
+++ b/src/Runtime/FastCgi/FastCgiRequest.php
@@ -1,0 +1,27 @@
+<?php declare(strict_types=1);
+
+namespace Bref\Runtime\FastCgi;
+
+use hollodotme\FastCGI\Requests\AbstractRequest;
+
+class FastCgiRequest extends AbstractRequest
+{
+    /** @var string */
+    private $method;
+
+    public function __construct(string $method, string $scriptFilename, string $content)
+    {
+        $this->method = $method;
+        parent::__construct($scriptFilename, $content);
+    }
+
+    public function getRequestMethod(): string
+    {
+        return $this->method;
+    }
+
+    public function getServerSoftware(): string
+    {
+        return 'bref';
+    }
+}
--- a/src/Runtime/PhpFpm.php
+++ b/src/Runtime/PhpFpm.php
@@ -3,10 +3,11 @@
 namespace Bref\Runtime;
 
 use Bref\Http\LambdaResponse;
-use Hoa\Fastcgi\Exception\Exception as HoaFastCgiException;
-use Hoa\Fastcgi\Responder;
-use Hoa\Socket\Client;
-use Hoa\Socket\Exception\Exception as HoaSocketException;
+use Bref\Runtime\FastCgi\FastCgiCommunicationFailed;
+use Bref\Runtime\FastCgi\FastCgiRequest;
+use hollodotme\FastCGI\Client;
+use hollodotme\FastCGI\Interfaces\ProvidesRequestData;
+use hollodotme\FastCGI\SocketConnections\UnixDomainSocket;
 use Symfony\Component\Process\Process;
 
 /**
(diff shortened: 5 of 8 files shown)


atrope:
@mnapoli one important metric that we should be considering is performance, here my test:

https://prnt.sc/ngwe5q

The first part is using  hollodotme and the faster one is Hoa.

mnapoli:
@atrope thank you for testing, it is indeed something I wanted to test. I have the feeling that hollodotme's library is reconnecting to PHP-FPM every time. I need to dig into this 🤔

tests/Runtime/PhpFpmTest.php
```diff
@@ -745,6 +765,19 @@ public function test FPM timeouts are recovered from()
         }
     }
 
+    /**
+     * @see https://github.com/mnapoli/bref/issues/316
+     */
+    public function test large response()
```
kwn: Sorry - out of curiosity. How come php doesn't fail with syntax error? Is it some UTF-8 character that "looks" like a space?

mnapoli: :trollface: see https://mnapoli.fr/using-non-breakable-spaces-in-test-method-names/ for more details but it is valid PHP code as those are non-breaking spaces. This is an alternative to `snake_case`.

hollodotme:
@mnapoli Hi, thanks for considering my library for integration!

Correct, the lib currently opens a new connection to php-fpm for each request and disconnects as soon as the response was fetched. TBH, I never considered re-using the sockets for succeeding requests and now I'm asking myself why I didn't. :-/

However I quickly prototyped the re-using of sockets into the lib and that seems to actually work. :) (https://github.com/hollodotme/fast-cgi-client/pull/33) 

Looks like I would need some days to finish a new release with proper implementation and tests.

mnapoli:
@hollodotme awesome!

mnapoli:
@hollodotme sorry for the double ping I just thought about that:

https://github.com/mnapoli/bref/pull/319/files#diff-08c4870f917ac32beb81b794eb61b760L120

When communication via the socket fails we needed to explicitly reconnect with Hoa. I'm guessing that's something that will be necessary as well for you if you make connections persistent, so you may want to expose a public `disconnect()` or `reconnect()` method.

hollodotme:
@mnapoli There are several checks in the library if the socket is broken. In those cases the `Client` class (which you use) would/should create a new connection anyway. The lib tries to avoid that the user has to deal with the socket or connections directly. 

Is there any specific kind of "socket fails" that I can test against?

mnapoli:
Nice!

> Is there any specific kind of "socket fails" that I can test against?

What we test in Bref are timeouts: https://github.com/mnapoli/bref/blob/883d74ab410b99b585afd2109fa0dbedec985414/tests/Runtime/PhpFpmTest.php#L724-L746 (the file that is invoked simply does a `sleep()` until a timeout occurs).

There is also https://github.com/mnapoli/bref/issues/274#issuecomment-487029070 which is a separate issue (we haven't found the root cause), but I'm not sure how it relates to this discussion TBH.

tests/Runtime/PhpFpmTest.php
kwn: Yup, nice one. Thanks 👍

hollodotme:
@mnapoli I just [released `v2.7.0`](https://github.com/hollodotme/fast-cgi-client/releases) of the library that now automatically re-uses idle socket connections on successive requests.

In the following cases a new connection will be established:

- There is no connected socket
- All connected sockets are either busy or not usable anymore, which means 
  - their responses were not ready or received, yet (= busy)
  - the stream meta data indicates a time out, an EOF or unread bytes (= not usable)
  - the stream was (externally) closed or terminated (= not usable)

If non of this is true, the first idle connection from the internal collection is re-used for the next request.

In practice that means that all successive **synchronous** requests will use the same connection as long as the stream is usable.

For successive **asynchronous** requests it depends on the fact when a socket becomes idle again to be re-used (and how the triggering of requests is organized in call-flow).

I hope this helps to improve performance and reduce resource allocation.

mnapoli:
@hollodotme that is awesome! I have updated the required version in this PR. I'll try to do some performance tests (@atrope if you are able to test as well that would be great).

mnapoli:
I did some performance tests with https://github.com/brefphp/bref-benchmark and numbers are looking good!

It even seem to drop from 2ms to 1ms for HTTP response time of a `echo "Hello world"` application. I'm not sure if it is actually significant or not since we are alreay close to very low latencies, but it could be a performance improvement!

Hopefully @atrope and others will be able to test this and confirm that so we can be extra confident.

atrope:
Hey @mnapoli and @hollodotme I Updated to 2.7 and tested in production for 10 minutes.
It did not solved #274 and the performance was worse than before :-/
https://prnt.sc/ni0mpt

@mnapoli Do we need to enable something in phpFPM file? I Just Updated to 2.7 and used the PR's file (adapted to use ALB).

taylorotwell:
Does @hollodotme's implementation actually handle multiple headers with the same name like `Set-Cookie` correctly? Hoa's library does not and multiple cookies can't be returned in a response without overriding the library in a specific spot I added in a recent PR.

mnapoli:
@atrope thank you for testing. I wonder how we get such a huge difference (+90ms) on FPM communication 🤔 There must be something else going on. I'll try with a real application on my side as well.

@taylorotwell AFAIK no, @atrope was mentioning that as well and if I recall correctly it would be easier to implement with the new library.

hollodotme:
@taylorotwell, I fear the answer is no for the [current version](https://github.com/hollodotme/fast-cgi-client/blob/v2.7.0/src/Responses/Response.php#L77), as long as you don't want to parse the raw response (which is accessible and includes the header lines) yourself. 

I'll see what I can do in the next days. But it'll most likely be a BC break. 🙃

taylorotwell:
Has the issue been raised with the Hoa/FastCGI maintainer? Perhaps it can be fixed on that end without swapping to an entirely new library?

atrope:
@taylorotwell With Hollodotme it is doable in fpm file in our side. 
If @hollodotme Wants i can easily make a PR to include in his library also.

We just need a flag to enable Multi Header and if it is enabled return each Header as an array.

I Did it in my end:
```

    /**
     * Return an array of the response headers.
     *
     * @param ProvidesResponseData $response
     * @param boolean $isMultiHeader
     */
    private function getHeaders($response,$isMultiHeader): array
    {
      if ($isMultiHeader){
        $responseHeaders = [];
        $lines  = explode( PHP_EOL, $response->getRawResponse() );
        foreach ($lines as $i => $line ){
          if (preg_match('#^([^\:]+):(.*)$#', $line, $matches )){
            $key = trim( $matches[1]);
            if (!array_key_exists($key, $responseHeaders)) $responseHeaders[$key]= [];
            $responseHeaders[$key][] = trim($matches[2]);
            continue;
          }
          break;
        }
      }
      else {
        $responseHeaders = $response->getHeaders();
      }
      return array_change_key_case($responseHeaders, CASE_LOWER);
    }

```

atrope:
@taylorotwell and yes, i 've made a PR there:
https://github.com/hoaproject/Fastcgi/pull/28

atrope:
In Hollodotme's library just need to edit `parseHeadersAndBody` and add isMultiHeader variable to constructor, default as false, and a way to change it.

```
if ( preg_match( self::HEADER_PATTERN, $line, $matches ) )
			{
				$offset                               = $i;
				$name = trim( $matches[1] );
				$value = trim( $matches[2] );
				if ($this->isMultiHeader){ 
					if (!array_key_exists($key, $this->headers)){ 
						$this->headers[$key]= [];
					}
					$this->headers[$key][] = $value;
				}
				else {
					$this->headers[ $name ] = $value;	
				}
				continue;
			}
```

mnapoli:
I am trying to dig into performances, here is a simple script I'm trying:

```php
require_once __DIR__ . '/vendor/autoload.php';

$fpm = new Bref\Runtime\PhpFpm(__DIR__ . '/tests/Runtime/PhpFpm/request.php', __DIR__ . '/tests/Runtime/PhpFpm/php-fpm.conf');
$fpm->start();

for ($i = 0; $i <= 1000; $i++) {
    $response = $fpm->proxy([
        'httpMethod' => 'GET',
        'path' => '/hello',
        'requestContext' => [
            'protocol' => 'HTTP/1.1',
        ],
    ]);
}
```

Here is the Blackfire diff between master and this PR:

https://blackfire.io/profiles/compare/529fc635-88eb-4616-902b-75314b334fb2/graph

As you can see this PR runs 40% faster.

One of the explanation is because Hoa seems to be disconnecting/reconnecting every time (which surprises me):

Capture d’écran 2019-04-29 à 10 15 53

But beyond that Hoa is also slightly slower at reading the response. @atrope can you confirm that your screenshot wasn't all cold starts?

@taylorotwell we could indeed open an issue. TBH this new library seems to be more maintained (https://github.com/hoaproject/Fastcgi/graphs/contributors) and used, Hoa is dragging behind in new releases and the last security issue wasn't fixed upstream (https://github.com/hoaproject/Fastcgi/issues). There is also the Hoa magic around autoloading that makes it not so nice to use. In any case we could stick with Hoa if this issue is fixed upstream, nothing against it. I just don't want to hold my breath if something better is right there ^^ TBH I started this branch not expecting it to work, but since it does I've opened the PR.

mnapoli:
Just to be safe I tried with a larger JSON response and got the same results.

hollodotme:
@atrope, @taylorotwell  I filed an issue (hollodotme/fast-cgi-client#35) for me to add the multi-headers to the library. It will come in `v3.0.0` as it is a BC break. I think I can finish this release until the end of the week.

mnapoli:
@atrope I just thought about https://github.com/mnapoli/bref/issues/274#issuecomment-486800345: could it be that response times are worst because of the problem you are facing in #274? I.e. responses are broken, there is a lot of socket reconnect going on, etc.

atrope:
@mnapoli Yes i can assure it was not 10 minutes of Cold Starts :)

And while i cannot confirm or deny that the worst performance is caused by #274 , i think the evidence indicates that it's not that, because the only thing that changes from both performance tests is the underlying library. 

So Hoa's is also facing that problem but performance is still twice better than Hollodotme (I do prefer Hollodotme library, as i stated, that's the one i used before, and as i also stated before, When migrated to bref it came with a performance benefit that we are know finding out was the library).

Checking your tests I Still think that it could be a simple configuration and i will dig further today.

DO You know any way i can debug on prod so we can see where the latency is coming from?

mnapoli:
> DO You know any way i can debug on prod so we can see where the latency is coming from?

Not really. You could edit the PhpFpm class and log custom CloudWatch metrics, it would be a little bit of effort though.

Or you could also try to run a benchmark locally on your PC, without SAM (run it with PHP-FPM) and use https://blackfire.io/ to measure the performance differences. You can use `blackfire run php` instead of `php` and that's it. For my tests above I ran `blackfire run php perf.php`.

hollodotme:
I just [released `v2.7.1`](https://github.com/hollodotme/fast-cgi-client/blob/v2.7.1/CHANGELOG.md) with a bugfix related to #274.

bubba-h57:
@atrope I can create a debug layer/function for you. What tools or functionality would you like in it?

gonzalovilaseca:
Is this performance issue a blocker for this PR? (Just wondering if it will be merged soon)

mnapoli:
@gonzalovilaseca I would really like to figure out that performance problem before merging :/

gonzalovilaseca:
Ok, np. If I can help just let me know what I can do :-)

atrope:
@bubba-h57 anything that would let me see where is the added latency because the code is the same only bref's phpfpm file changes.

mnapoli:
If you are able to run your app locally you could profile the difference (xdebug/blackfire/etc.). But if the difference is because of something that happens on Lambda (or at scale) then that won't help ^^

atrope:
Ok now that my Environment is fixed we can see a better performance from hollodotme.

WHat was happening is the contrary, because lots of requests were failing the lambda time were less than 50ms making the average looks better.

Now when i changed from Hoa to hollodotme i saw a 10ms increase but i think it would be wise for us to make a one-day test for each library. i will test Hoa tomorrow without the errors, and the day after tomorrow i will put hollodotme 2.7.1 for one whole day also, that way we will know for sure if we have or not performance problems.

What do you say @mnapoli  are you in a hurry to merge this PR or i can make this Test( I Will do anyway)?

mnapoli:
OK that's good to know!

Take your time to test tomorrow and let us know. If all goes well I can try to do a release wednesday or thursday.

hollodotme:
I released 3.0.0-alpha of the library containing the new method and behavior for multi-value headers and the removal of the previously deprecated `Response#getRawResponse()` method in favor of `Response#getOutput()`. 

* [BC breaks are documented here](https://github.com/hollodotme/fast-cgi-client/blob/a21c50434acee44731e41b34f5cb2199ce175aa9/CHANGELOG.md#backwards-incompatible-changes-bc-breaks)
* [New header behavior is documented here](https://github.com/hollodotme/fast-cgi-client/blob/v3.0.0-alpha/README.md#responses)

I still want to add some more tests and maybe one or two methods to the request classes for more user convenience. That's why it's an alpha "only".

Hywan:
Hey :-), if the issue comes from `hoa/fastcgi`, it's easily fixable. The size of the HTTP response is hardcoded for obvious reason, but we can increase it easily. Do you need that fix?

Hywan:
Also, about connection and disconnection, if you can provide me a script to replicate, I can fix that quite easily too.

atrope:
@mnapoli  Ok, The performance was pretty much the same between HOA and HOLLODOTME.

:+1:

Hywan:
@atrope @mnapoli Do you need hoa/fastcgi to support connecting/disconnecting once, and not per `send` call? It's easy to integrate.

bubba-h57:
If you want to use the xdebug profiler in Lambda: 

`arn:aws:lambda:us-east-1:965741605173:layer:bref-php-73-xdebug:1` add that layer and activate the extension at `/opt/sts/modules/xdebug.so`

I only have a *US-EAST-1* layer published.

mnapoli:
OK thanks everyone for testing, @Hywan thanks for the heads up.

I think we should move ahead with this change, hopefully that's ok with everyone 🤞 I've been thinking about this these last few days:

- the work here is done already and the bug is fixed, it's ready to merge
- for my own perspective, the hollodotme library is definitely easier for me to dig into in case I need to debug or even maintain (if that's ever necessary), the Hoa code is split over many repositories and uses patterns and PHP practices I'm not familiar with (I'm thinking for example about the custom autoloading)
- better DX with the hollodotme library as a maintainer (that's purely from my personal POV), in Hoa classes are magically resolved and this breaks several features of the IDE
- so far it seems that the hollodotme library has more active users ([issues](https://github.com/hoaproject/Fastcgi/issues?utf8=%E2%9C%93&q=is%3Aissue+sort%3Aupdated-desc+) vs [issues](https://github.com/hollodotme/fast-cgi-client/issues?utf8=%E2%9C%93&q=is%3Aissue+sort%3Aupdated-desc+)) and shorter maintenance cycles ([stats](https://isitmaintained.com/project/hoaproject/Fastcgi) vs [stats](https://isitmaintained.com/project/hollodotme/fast-cgi-client) to illustrate)
- no Composer dependencies in the hollodotme library (a really negligible advantage I agree)

bubba-h57:
*no Composer dependencies in the hollodotme library* <- this is the most awesome advantage evah!

Issue #275, opened 2019-03-18
Bref using all space in /tmp folder


atrope:
Hi, 

I was having some issues with our caching files (it writes to /tmp) , telling me that /tmp was full.

Debugging it with i found out that it was core dump logs that were taking all space:
```

Filesystem     Type      Size  Used Avail Use% Mounted on
/dev/xvda1     ext4       30G  5.4G   24G  19% /
/dev/loop1     ext4      526M  515M     0 100% /tmp
/dev/loop7     squashfs   43M   43M     0 100% /opt
total 514M
-rw------- 1 sbx_user1075 471    0 Mar 15 03:22 core.php-fpm.106
-rw------- 1 sbx_user1075 471    0 Mar 15 03:22 core.php-fpm.108
-rw------- 1 sbx_user1075 471 143M Mar 15 03:21 core.php-fpm.11
-rw------- 1 sbx_user1075 471    0 Mar 15 03:22 core.php-fpm.110
-rw------- 1 sbx_user1075 471    0 Mar 15 03:22 core.php-fpm.116
-rw------- 1 sbx_user1075 471    0 Mar 15 03:22 core.php-fpm.118
-rw------- 1 sbx_user1075 471    0 Mar 15 03:22 core.php-fpm.119
-rw------- 1 sbx_user1075 471    0 Mar 15 03:22 core.php-fpm.120
-rw------- 1 sbx_user1075 471    0 Mar 15 03:22 core.php-fpm.121
-rw------- 1 sbx_user1075 471    0 Mar 15 03:22 core.php-fpm.122
-rw------- 1 sbx_user1075 471    0 Mar 15 03:22 core.php-fpm.125
-rw------- 1 sbx_user1075 471    0 Mar 15 03:22 core.php-fpm.126
-rw------- 1 sbx_user1075 471    0 Mar 15 03:22 core.php-fpm.127
-rw------- 1 sbx_user1075 471 142M Mar 15 03:21 core.php-fpm.13
-rw------- 1 sbx_user1075 471    0 Mar 15 03:22 core.php-fpm.130
-rw------- 1 sbx_user1075 471    0 Mar 15 03:22 core.php-fpm.132
-rw------- 1 sbx_user1075 471    0 Mar 15 03:22 core.php-fpm.133
-rw------- 1 sbx_user1075 471    0 Mar 15 03:22 core.php-fpm.134
-rw------- 1 sbx_user1075 471    0 Mar 15 03:22 core.php-fpm.135
[...]

PR #321, opened 2019-04-30
Avoid dump logs

atrope:
As Explained in #275 this PR adds a limit to dump logs in order to avoid usage of all lambda space in case of FPM dumps.


mnapoli:
Thanks!

PR #323, opened 2019-05-03
tell the user to run aws configure when process exits

nealio82:
When running `aws configure`, if the process throws a `ProcessSignaledException` we should catch it and give the user a bit more information on how to dig into the problem, otherwise they have to follow the stack trace through Bref's code to try to figure out what we were trying to do at the time it borked.

refs: https://github.com/mnapoli/bref/issues/322


mnapoli:
Thanks!

PR #334, opened 2019-05-16
Handling keepWarm events to mitigate cold start

thibaudlemaire:
Hi all !  
Thanks for your work on this awesome project !

We can prevent our Lambda container from dying by calling it periodically.
[Jeremy Daly explains here](https://www.jeremydaly.com/lambda-warmer-optimize-aws-lambda-function-cold-starts/) how he created [Lambda Warmer](https://www.npmjs.com/package/lambda-warmer), a simple Node.js plugin that waits for a special kind of event and replies without running the whole function. 

I tried to generate such events using CloudWatch but I got my logs polluted with Excpetion messages.  
This PR allows us to send keepWarm events using CloudWatch without polluting the logs.

This is a first version as it doesn't manage yet the multiple concurrency warming. But it can be usefull to test on small applications.

### DIFF
--- a/docs/runtimes/http.md
+++ b/docs/runtimes/http.md
@@ -136,3 +136,21 @@ Use `{foo}` as a placeholder for a parameter and `{foo+}` as a parameter that ma
 Lambda and API Gateway are only used for executing code. Serving assets via PHP does not make sense as this would be a waste of resources and money.
 
 Deploying a website and serving assets (e.g. CSS, JavaScript, images) will be covered later in another article.
+
+### Cold start
+
+AWS Lambda automatically destroys Lambda containers that have been unused for 10 to 60 minutes. Warming up a new container can take severals seconds, especially if your package is large or if your Lambda is connected to a VPC. This delay is called [cold start](https://mikhail.io/serverless/coldstarts/aws/).
+
+To mitigate cold start, you can periodically send an event to your Lambda including a `{warmer: true}` key. Bref recognizes this event and immediately responds with a `{status: 100}` without calling your handler file.  
+
+You can generate automatically such events using AWS CloudWatch. For exemple : 
+
+```yaml
+            Events:
+                ...
+                Warmer:
+                    Type: Schedule
+                    Properties:
+                        Schedule: rate(5 minutes)
+                        Input: '{"warmer": true}'
+```
--- a/src/Runtime/PhpFpm.php
+++ b/src/Runtime/PhpFpm.php
@@ -107,6 +107,10 @@ class PhpFpm
      */
     public function proxy($event): LambdaResponse
     {
+        if (isset($event['warmer']) && $event['warmer'] === true) {
+            return new LambdaResponse(100, [], 'Lambda is warm');
+        }
+
         if (! isset($event['httpMethod'])) {
             throw new \Exception('The lambda was not invoked via HTTP through API Gateway: this is not supported by this runtime');
         }


deleugpn:
Mindblowing

cmaerz:
Nice Should be implemented.

mnapoli:
Thank you this is a really good addition.

I am busy these weeks with a few events but we'll definitely want to merge that. A few things to discuss in the meantime:

- are there other libraries/tools that do that in other languages? If so, maybe it's worth reusing the same "event" format as them (just for consistency's and interoperaibility's sake)
- would it be worth implementing with other kind of events? The downside here is that the `keepWarmEvent` key could be conflicting with an actual event (which can't happen with HTTP), so it might be risky.

thibaudlemaire:
Thanks Matthieu for your answer.

- In JS there is the [Lambda Warmer](https://github.com/jeremydaly/lambda-warmer) library in which the event flag is "warmer" by default but configurable. I'll change the event format.  
- Why do you think it might be risky ? An event is dispatched to a specific Lambda only if it's invoked with the corresponding FunctionName right ?

I was thinking of implementing the multiple concurrency mechanism (to warm up many Lambda containers in the same time).  
I can do it either,
- synchronously : the first lambda warms up the second, that warms up the third (etc..) in a recursive way. It's simple and sufficient if you want to warm up not many Lambdas, and takes N*2ms approx to warm up each lambdas.
- asynchronously [(as described here)](https://www.jeremydaly.com/wp-content/uploads/2018/07/lambda-warmer-multiple-concurrency.png) : the first lambda triggers the others asynchronously except the last that is invoked synchronously to release the first one. The drawback of this method is that the lambdas triggered asynchronously must wait a long time (Jeremy Daly recommends 75ms) to be sure that an existing container is not reused. During this "waiting" time, asynchronously triggered lambdas are unavailable for incoming requests (that can be handled the two remaining warm lambdas or a new cold one). 

What do you think about these two strategies ?

mnapoli:
> Why do you think it might be risky ? An event is dispatched to a specific Lambda only if it's invoked with the corresponding FunctionName right ?

What I mean by that is that Bref cannot blindly intercept any event that contains a `warmer` key. Maybe some legitimate events will not be processed because of that.

If the function is a HTTP function that's fine because we can safely say that this is not a HTTP event, so there is no confusion. But with non-HTTP events there can be confusion.

Your PR is doing the safe implementation: it's only about HTTP. That makes sense for a first approach (unless we find a safe solution).

> I was thinking of implementing the multiple concurrency mechanism (to warm up many Lambda containers in the same time).

That's interesting! Let's keep this PR about the first (simpler) version though, so that we can iterate in small changes (keeping up with everything happening in the project is a challenge for me ^^). Feel free to work on a more complex version in a separate PR.

thibaudlemaire:
> What I mean by that is that Bref cannot blindly intercept any event that contains a warmer key. Maybe some legitimate events will not be processed because of that.  

You're right. A simple solution could be to make the key name configurable. 

> Let's keep this PR about the first (simpler) version though  

All right

thibaudlemaire:
I amended my commit to use 'warmer' instead of 'keepWarmEvent'

mnapoli:
:+1: the docs might need a little update, and then we are good to merge!

thibaudlemaire:
Doc's been updated :)

mnapoli:
Thank you!

PR #338, opened 2019-05-27
Follow up on #263: ALB support

### LINKED ISSUE #262: Adding ALB Support
Hello,
I Am working on adding ALB Support (with multi value headers) to bref.
https://aws.amazon.com/pt/blogs/networking-and-content-delivery/lambda-functions-as-targets-for-application-load-balancers/

I Already managed to it without a problem.
The only issue is that we also need to alter Hoa Responder class to process multi headers, how should we do it?
First update there and then here?

Thanks!

mnapoli:
This PR includes #263 to add ALB support and multi-value header support (fix #262).

This PR adds tests, and I refactored some parts to optimize the diff. What I mean by that is that we change quite a few things, which makes the diff huge and hard to review. I reorganized pieces of code from #263 so that it's easier to review for me.

Once merged this will close #263 and mark it as merged (as the commits are included in this PR).


atrope:
@mnapoli  IMHO all good in this PR.

I would just avoid keeping toApiGatewayFormat because it is not only APIGW, but thats just the name.
We could rename it and keep a helper function to call the new one, but i agree this should not matter the most.

mnapoli:
> I would just avoid keeping toApiGatewayFormat because it is not only APIGW, but thats just the name.
We could rename it and keep a helper function to call the new one, but i agree this should not matter the most.

Agreed, but I had to keep the old name to make sure old layer versions still work. We could write an alias function and mark it as deprecated, but honestly I'm not sure it's worth the trouble for now.

My goal is to be able to couple a layer version with a Bref version. That way you simply upgrade Bref with Composer and you don't have to think about layer versions. And we can change both the code and the layers in parallel, which would remove the problem we have right now.

PR #342, opened 2019-06-11
Move the project to the Bref organization on GitHub and Packagist

### LINKED ISSUE #328: Move to the Bref organization on GitHub and Packagist
We have passed 1000 stars on GitHub 🎉 

It's time for Bref to move to its own organization: https://github.com/brefphp

I did it previously for PHP-DI and this is what I remember:

- move the project (do not fork it!) on GitHub: we will keep all issues, history, stars, watchers, etc. The old repository URL will redirect to the new. This works as well for git URLs IIRC.
- change `composer.json` to use the `bref/` vendor name
  - the packagist project will be *created*, not migrated
  - the old package will still exist (its stats will not be migrated to the new one): how can we mark the old one deprecated? should we push a last release on that name?

Users will have to change their composer.json to use `bref/bref` instead of `mnapoli/bref`.

This could be released a `0.4.0` but is it necessary?

mnapoli:
Fixes #328

On GitHub the project will be moved to the https://github.com/brefphp organization. GitHub will redirect the old URL (https://github.com/mnapoli/bref) to the new one (https://github.com/brefphp/bref).

On Packagist the project will be renamed from `mnapoli/bref` to `bref/bref`. The old versions will continue to exist and be installable via `mnapoli/bref`. The old and new versions will be installable via `bref/bref`.

I plan on tagging this as 0.4.0 as well to avoid any confusion. There should be no BC break (but it will be documented in the release notes).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "mnapoli/bref",
+    "name": "bref/bref",
     "license": "MIT",
     "type": "project",
     "description": "Bref is a framework to write and deploy serverless PHP applications on AWS Lambda.",
--- a/README.md
+++ b/README.md
@@ -4,10 +4,10 @@ layout: home
 
 [![Running PHP made simple. Bref provides tools and documentation to easily deploy and run serverless PHP applications. Learn more](docs/readme-screenshot.png)](https://bref.sh/)
 
-[![Build Status](https://travis-ci.com/mnapoli/bref.svg?branch=master)](https://travis-ci.com/mnapoli/bref)
-[![Latest Version](https://img.shields.io/github/release/mnapoli/bref.svg?style=flat-square)](https://packagist.org/packages/mnapoli/bref)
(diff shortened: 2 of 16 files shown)


composer.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "mnapoli/bref",
+    "name": "bref/bref",
```
stloyd: You can add:
```
"replace": {
        "mnapoli/bref": "*"
}
```
https://getcomposer.org/doc/04-schema.md#replace

mnapoli: Thanks, I'm not certain we need this, especially since we are not sure what will happen. Bref is not used much as a library, I'm not sure the gain is worth the risk.

Nyholm: You don't need to rename the package name. 

I would actually suggest to keep it the same.

mnapoli:
Thanks, I think I would prefer to move everything to the `bref` namespace, including Packagist.

We will lose a bit of download history and the transition will require action from users, but the userbase is still small. It's now or never.

Issue #237, opened 2019-02-11
Docker images


deleugpn:
Would it be easy or possible for Bref to provide one Docker image that contains the exact same php binary that the lambda layer has which we could use on our CI / CD build process to install composer and run phpunit? I know I can easily build a container myself to do this, but I would like to make sure that my tests run as close as the production php binary will be (e.g. am I using an extension that the lambda doesn't have, but my unit tests will flag?).
I'm not sure how easy it would be for bref to maintain an image like that on Docker Hub.

mnapoli:
Honestly I'm not sure how much work would that be. It could be a possibility, just to let you know at the moment however I am focusing on other issues and the documentation.

deleugpn:
I'm looking this again and I think I found out how simple it is. I made an organization on Docker Hub called `breftest` (the name `bref` is still available) and created a repository called php (https://hub.docker.com/r/breftest/php). The goal is to have the Docker image with the PHP binary for lambda available (https://github.com/mnapoli/bref/blob/master/runtime/php/php.Dockerfile#L485-L496). If this particular image gets built and pushed to Docker Hub (preferably under the official library), then we can use it to run things like `composer install`, `php artisan route:cache` or any other project-specific setting as well as `./vendor/bin/phpunit`. If all the tests passes using that php binary, we know for a fact that we won't have any problems related to different php version or different extensions between local, test and production.

mnapoli:
Very good points! I am more interested now in this issue as well. I went ahead and secured the "bref" team.

I am not that familiar with Docker Hub: when creating a repository should I create it from scratch, or create it linked to this GitHub repository? (we may need to provide several Docker images over time)

deleugpn:
I don't have experience with Docker Hub either as I'm mostly using AWS ECR. However I feel like the Docker Hub integration with GitHub is just a facilitator tool that will listen to push events and build the images for you automatically. My suggestion would be to just add an automatic `docker push` in your existing build script and only worry about automating it some other time if it makes sense to dig deeper into that.

I just got a huge milestone for a PHP pipeline: I built the images and pushed it to my repo (https://hub.docker.com/r/breftest/php/tags) and then I succeeded by running it like this: 

```
      - docker run --name laravel-app -v $(pwd):/app -w /app -t breftest/php:73 sh -c "php -r \"copy('https://getcomposer.org/installer', 'composer-setup.php');\" && php composer-setup.php && php composer.phar install && /app/vendor/bin/phpunit"
[...]

PR #344, opened 2019-06-12
[WIP] Documentation for PHPUnit

deleugpn:
This Pull Request is an attempt to start the work for #237. I successfully accomplished the desire to run PHPUnit using the php binary provided by the layer using my personal Docker Hub (https://hub.docker.com/r/deleugpn/bref/tags) and I would like to provide as much help as I can to bring this to Bref.

At first, I thought about documenting the entire AWS Code Pipeline process, but maybe that's too much at once and not everyone is using CI/CD. I tried to reduce the scope as much as I could, leading to just documenting how to run phpunit using the same binary. I would be happy to provide a documentation section dedicated for CodePipeline if there is any interest from the community to have this documented (and if Bref wants to provide the documentation).

The reason I mention CI/CD is because for me they're quite intertwined. AWS CodePipeline downloads my source code from GitHub every time I commit and there's no vendor folder, so I had to use 2 Docker images: 1 with PHP Binary + Composer + building instructions and 1 with just PHP Binary. Running composer with a different version of PHP as production might also cause issues.
However, for the sake of brevity, at the moment I'm going to assume this is a deeper problem for another pull request and focus solely on having a Docker image with the binary only that can be used for executing the tests once the source code is ready for it.

Things to consider:

- How to tag the Docker Image? With the same tag as the Layers or with the PHP version?
- Should we focus the documentation on phpunit terms or be more generic about testing frameworks?
- Are there any other considerations I'm missing for this documentation?

I'd love to get feedback on this process as a whole and help bref provide this Docker image officially.

### DIFF
--- a/docs/frameworks/phpunit.md
+++ b/docs/frameworks/phpunit.md
@@ -12,14 +12,15 @@ For this reason, bref provides a Docker image containing the same php binary as 
 
 ## Usage
 
-```
-docker run -v /path/to/my/project:/var/task -t bref/php-73 ./vendor/bin/phpunit
-```
+The following command assumes that you're executing it from the root folder of your project:
 
-TODO: Should we tag the Docker image with the PHP version or with the Layer tag?  
+```
+docker run -v $(pwd):/var/task -w /var/task -t bref/php-73 sh -c "php -r \"copy('https://getcomposer.org/installer', 'composer-setup.php');\" && php composer-setup.php && php composer.phar install && ./vendor/bin/phpunit"
+```
 
 ## How it works?
 
 When building the Lambda Layer, Bref uses a base image provided by Amazon containing the Amazon Linux. This is the same operating system that Lambda is using under the hood.
 The PHP binary is then compiled from source code and packaged into a Lambda layer. 
 Bref pushes the same Docker image used to make up the layer to Docker Hub, which means the exact same binary is being used.
+After starting the container, the command will download and run Composer on the project in order to install all the dependencies, including the development dependencies to bring in the testing framework (e.g. PHPUnit).
--- a/docs/frameworks/phpunit.md
+++ b/docs/frameworks/phpunit.md
@@ -1,0 +1,25 @@
+---
+title: Serverless Testing with PHPUnit
+currentMenu: frameworks
+introduction: Run your unit tests as close to production environment
+---
+
+Running unit tests in an environment that is not close to the production environment might not bring as much confidence as expected from a test suite.
+One way of visualizing this fact is to try and run a test suite using PHP 7.0 against a codebase that makes use of nullable type-hit. 
+Of course, this example is extreme, but the sentiment remains: it is somewhat relevant to execute a test suite using the same php binary that will be running on production to cater for php version as well as required extesions.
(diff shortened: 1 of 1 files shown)


mnapoli:
:+1: that's a good first step.

> How to tag the Docker Image? With the same tag as the Layers or with the PHP version?

I think we have a 3rd option as well:

- use the layer version
- use the PHP version
- use the Bref version

In the future, I would like to avoid having to deal with layer versions and only deal with Bref versions. That could be possible if Bref provides a way to automatically fill in the correct layer version to use.

Then you can simply upgrade with Composer and you know that you use the correct version.

So I think using the Bref version might be what makes more sense.

> Should we focus the documentation on phpunit terms or be more generic about testing frameworks?

We could show an example with PHPUnit. As long as it's easy to replicate with another test framework I think it's fine.

> Are there any other considerations I'm missing for this documentation?

In the future I'd love to provide whatever is necessary to easily test a function (i.e. the `lambda(function () { ... })` thing. But that can be done later.

deleugpn:
I don't know if this is the best approach, but I put the composer install in the sample to show exactly how this is working right now in combination with the `bref/php-73` image. Maybe we should make a `install.sh` inside the `bref/php-73` image that does these steps?
The file could also be overridden by volume if the developer wants.

mnapoli:
Why run `composer install` inside the Docker image and not outside before running PHPUnit?

deleugpn:
I have 3 problems with running it outside, but I guess it could be interpreted as a personal preference.

1- I don't have PHP in my machine. Everything is docker. If I'm working on a project with PHP 7.2, I depend on the 7.2 container to be able to run composer install using a php 7.2 binary to correctly match all dependencies. If I switch to another project on php 7.3 I would need a php 7.3 binary to run composer install as well.

2- When using AWS CodePipeline, the pipeline is fetching the source code form GitHub and it needs to install the php dependencies before running the tests. In this scenario, the AWS CodeBuild machine that I have available does not even know what PHP is. It only has Docker installed. I would have to install PHP outside and try and guarantee extensions and version of the binary matches.

3- PHP Version and Extensions. Running composer install with the exact same binary might fail if the binary doesn't contain an extension that my code needs, which will give me a fast feedback that my project will not work on the lambda. 

I understand these reasoning might be a nitpick and that the docs here might assume that it's the developer's problem to guarantee compatibility between local php and the docker image, but it feels unnecessary divergence. I agree that if I have a test that covers the part where an extension is used, the test will fail. But if for some reason I lack that test, I can still count on `composer install` failing if I'm missing a depended extension.

mnapoli:
Gotcha, that makes sense. It might be worse separating both tasks though (at least for the documentation), in order to make thing simpler.

For example:

```bash
docker run -v $(pwd):/var/task -w /var/task -t bref/php-73 sh -c "php -r \"copy('https://getcomposer.org/installer', 'composer-setup.php');\" && php composer-setup.php && php composer.phar install"

docker run -v $(pwd):/var/task -w /var/task -t bref/php-73 php vendor/bin/phpunit
```

mnapoli:
Closing for now :)

PR #346, opened 2019-06-13
Layer Region note

MaxGiting:
Regions always come back and bite me.

I don't know if this is the best way to alert people to this, but I wanted to try and contribute back instead of just mentioning that I had this issue and possibly other people will too.


mnapoli:
Thank you.

I really wish that we wouldn't have to duplicate that advice into framework-specific documentation. But at the moment those articles are not ideal, so what you suggest makes sense in the short term. In the long term I will probably refactor that documentation differently.

Issue #347, opened 2019-06-14
Enhancement: Output logs from console layers


MaxGiting:
It would be useful if logging could be output from console layers to CloudWatch.

My use case was setting up Events in CloudWatch to trigger commands. So some feedback into CloudWatch would be useful.

@mnapoli has suggested looking into the below line to enable this functionality.
https://github.com/brefphp/bref/blob/3450616d2535b13ff086124a6cca5a9000252902/runtime/console/bootstrap#L31

mnapoli:
:+1:

As you mentioned this feature *may be* as simple as adding `echo $output` in the bootstrap file. Hopefully I'm not missing anything :)

Then `bref cli ...` should display the output, but the output should also be sent to CloudWatch.

MaxGiting:
This has been fixed in #349

PR #349, opened 2019-06-16
Log output to Cloudwatch for console layer

MaxGiting:
Related to issue #347 

This PR allows logging and other output from the console layer to be captured by CloudWatch. I have tested that it works, but don't know if there is any other considerations or tests that should be done.

You can test this by using 
`arn:aws:lambda:eu-west-2:928780349604:layer:consolebref:1`
(I can replicate this layer in another region if requested)

The layer can be download to check its contents by using the below command via the AWS CLI
`aws lambda get-layer-version-by-arn --arn arn:aws:lambda:eu-west-2:928780349604:layer:consolebref:1`

### DIFF
--- a/runtime/console/bootstrap
+++ b/runtime/console/bootstrap
@@ -30,6 +30,9 @@ while (true) {
 
         $output = implode("\n", $output);
 
+        // Echo $output so it is written to CLoudwatch logs as well as terminal connections.
+        echo $output;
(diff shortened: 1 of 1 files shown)


mnapoli:
Thank you!

FYI I'll probably build new runtimes when tagging the next release.

PR #353, opened 2019-06-20
Use Bash to parse the ini file

Nyholm:
Im building images on my CI and I dont have PHP installed on that docker image. This PR will make sure we parse the ini file in bash. 

Im not sure if you will accept this PR. The positive is that we will remove PHP from the build process. The negative is that parsing with bash and regex is not as robust as PHP's `parse_ini_file()`.


mnapoli:
I guess why not, I see the point and maybe we'll run into the same problem when we automate releases.

Thanks!

PR #354, opened 2019-06-20
Move from SAM to Serverless

### LINKED ISSUE #320: SAM vs Serverless: switch to Serverless again?
Yes let's discuss switching back again from SAM to the [serverless framework](https://serverless.com/) ([documentation](https://serverless.com/framework/docs/providers/aws/)) 😄  (see #99)

I open this issue because things have moved forward with the serverless framework. Mainly: it supports layers and it can run things locally with Docker.

But another big reason is that, after trying again serverless, it solves a lot of small pain points with SAM. For example:

- the region is defined in the config, you never have to deal with it in the CLI
- deploy in only one step (instead of 3 with SAM)

and **it is extensible**. That opens very interesting doors, for example to simplify a lot some things like setting layer ARNs:

```yaml
Resources:
    SimpleFunction:
        Properties:
            FunctionName: 'bref-demo-function'
            Handler: function.php
            Runtime: provided
            Layers:
                - 'arn:aws:lambda:us-east-2:209497400698:layer:php-72:4'
```

After with serverless:

```yaml
functions:
    bref-demo-function:
        handler: function.php
        runtime: provided
        layers:
            - ${bref:layer.php-72}
```

What's great with that as well is that runtime versions could be handled by Bref itself, making upgrading Bref as simple as `composer update`.

In short **serverless could help make things simpler**.

## BC break

If we switch we would direct tooling efforts towards serverless.

**However Bref would still work with SAM**.

But since the documentation would be rewritten to target serverless I think it could be considered a BC break (or a major change).

## Full comparison SAM vs Serverless

Here is a complete comparison of features:

https://docs.google.com/spreadsheets/d/1DP2uVbg6TLfBa56XJ1GwlSSKQM1sM87-S3LQ2Pl9sPI/edit?usp=sharing

Please have a look and comment (here or in Google Docs) what is missing and what should be added to the list.

As you will see Serverless has a lot of advantages, the only big question left is local development.

WDYT?

### LINKED ISSUE #351: Ease configuring layers with a Serverless plugin
I open this issue to share a bit my progress around a [Serverless plugin](https://serverless.com/framework/docs/providers/aws/guide/plugins/).

At the moment we use AWS SAM to deploy, but we are considering switching to Serverless (see #320). If we do (which is very likely), we could improve some things via a Serverless plugin.

### Why?

For example, one main improvement I am pursuing is to simplify dealing with layers and layer versions. For example:

```yaml
# Before
functions:
    foo:
        handler: index.php
        layers:
            - 'arn:aws:lambda:us-east-1:209497400698:layer:php-73:7'

# After
plugins:
    - ./vendor/bref/bref
functions:
    foo:
        handler: index.php
        layers:
            - ${bref:layer.php-73}
```

In this example, the plugin is contained in the Bref repository and loaded from the `vendor/` directory. The plugin provides a variable to reference easily Bref layers.

The big benefit is that we don't specify layer versions anymore. Meaning that upgrading layers will be as simple as upgrading Bref via Composer.

We will be able to easily break backward compatibility in layers since they will now be coupled to a specific Bref version. We will also be able to publish "beta" layers when testing something out (at the moment it isn't possible, which I don't really like).

Having a plugin could also provide additional features and helpers, or variables. I have a few ideas, but let's start with layers first :)

### How?

Serverless plugins are written in Javascript. At the moment they are distributed and loaded via NPM (in the `node_modules/` directory).

I am trying to add a new feature to the Serverless framework so that it is possible to load the plugin from the `vendor/` directory: https://github.com/serverless/serverless/issues/6017

Here is the current pull request: https://github.com/serverless/serverless/pull/6261

**If you can help with this please do :)** I am trying my best with NodeJS but I am far from an expert.

mnapoli:
Closes #320, closes #351 

This PR implements the change described in #320: switch from SAM to [Serverless](https://github.com/serverless/serverless) as the recommended deployment tool.

## TODO

- [x] Update the templates
- [x] Update the demo
- [x] Create the serverless plugin
- [x] Store the layer versions in the repository
- [x] Finish updating the documentation
- [x] ~~Update the SAM tests~~ Serverless is slower, I'll keep SAM to run the tests for now
- [x] Remove legacy `template.yaml` files
- [x] ~~Write an update guide~~ I'll write it in the release notes, I want to keep the website as simple as possible
- [x] Local development
- [x] Bref init

## Why?

Serverless lets us write much simpler and smaller config for the same result.

Here is a comparison for a HTTP application:

- with SAM (`template.yaml`):

```yaml
AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31
Description: ''

Resources:
    MyFunction:
        Properties:
            FunctionName: api
            Handler: index.php
            Timeout: 30 # in seconds (API Gateway has a timeout of 30 seconds)
            MemorySize: 1024
            Runtime: provided
            Layers:
                - 'arn:aws:lambda:us-east-1:209497400698:layer:php-73-fpm:7'
            Events:
                HttpRoot:
                    Type: Api
                    Properties:
                        Path: /
                        Method: ANY
                HttpSubPaths:
                    Type: Api
                    Properties:
                        Path: /{proxy+}
                        Method: ANY

# Outputs show up in the CloudFormation dashboard
Outputs:
    DemoHttpApi:
        Description: 'URL of our function in the *Prod* environment'
        Value: !Sub 'https://${ServerlessRestApi}.execute-api.${AWS::Region}.amazonaws.com/Prod/'
```

- with Serverless (`serverless.yml`):

```yaml
service: app

provider:
    name: aws
    runtime: provided

functions:
    api:
        handler: index.php
        timeout: 30 # in seconds (API Gateway has a timeout of 30 seconds)
        layers:
            - 'arn:aws:lambda:us-east-1:209497400698:layer:php-73-fpm:7'
        events:
            -   http: 'ANY /'
            -   http: 'ANY /{proxy+}'
```

Read #320 for more details on what we gain.

**Note that SAM will continue to work with Bref**. You don't _have to_ ditch SAM if you'd rather stick with it. That's fine.

## Serverless plugin

This closes #351 

On top of switching to Serverless, this PR provides a Serverless plugin. For the moment the plugin provides the following variables:

- `bref:layer.php-72`
- `bref:layer.php-73`
- `bref:layer.php-72-fpm`
- `bref:layer.php-73-fpm`
- `bref:layer.console`

This makes it easier to reference Bref layers. Here is a diff to show how to use them:

```diff
+ plugins:
+     - ./vendor/bref/bref

functions:
    api:
        handler: index.php
        layers:
-            - 'arn:aws:lambda:us-east-1:209497400698:layer:php-73-fpm:7'
+            - ${bref:layer.php-73}
        events:
            -   http: 'ANY /'
            -   http: 'ANY /{proxy+}'
```

It is shorter, simpler (no need to care about the region anymore), but it also means we don't have to track and update layer versions anymore. To get the latest versions you just need to update Bref via `composer update`.


mnapoli:
I got a bit carried away, I rewrote a good part of the documentation to get started and install everything. Much clearer now, and I've added screenshots to create the AWS access keys.

I have also finished the plugin and started documenting it. Everything is awesome, I can't wait to release that. The whole developer experience will be much much better I think.

mnapoli:
Preview URL for the documentation: https://5d2303e6a4cbf9110aba55cd--bref-website.netlify.com/docs/environment/serverless-yml.html

mnapoli:
I have deprecated `vendor/bin/bref invoke` in favor of `serverless invoke`:

- simpler to use (picks up the region from serverless.yml for example)
- all commands in one tool (serverless)
- less things to maintain in Bref

The command will still work for now, I haven't removed it.

mnapoli:
Regarding local development, since `serverless-offline` doesn't work with layers yet (https://github.com/dherault/serverless-offline/pull/648), and since it is pretty slow anyway, I've edited the docs to suggest using a classic stack when running locally (for HTTP only).

We are separately working on Docker images in #237, that should help running locally with an environment close to production. And it will be super fast as well. But at least it won't block this pull request.

lmcnearney:
I pulled the `serverless` branch down and was playing around with it. One immediate issue I had was that any serverless command would load the plugin but error out as it couldn't open the `layers.json` file. Looking at some other plugins, they resolve the path using the [__dirname](https://nodejs.org/docs/latest/api/modules.html#modules_dirname) global.

```
const fs = require("fs");
const path = require('path');
const filename = path.resolve(__dirname, 'layers.json');
const layers = JSON.parse(fs.readFileSync(filename));
```

mnapoli:
@lmcnearney thank you for testing, I really appreciate it! Until now I have tested it in the Bref repository, and paths are different here ^^

I'll work on fixing this. Thanks again.

mnapoli:
Thanks for the code example @lmcnearney! That should be fixed.

As soon as the build finishes I'll be merging and tagging an alpha version, that way it will be easier to test locally.

PR #356, opened 2019-06-23
Allow reading AWS credentials from environment

Nyholm:
When I run `vendor/bin/bref cli <function-name>` I get an error saying: 

```
In InstanceProfileProvider.php line 95:
                                                                                                                                               
  Error retrieving credentials from the instance profile metadata server. (cURL error 28: Connection timed out after 1004 milliseconds (see h  
  ttp://curl.haxx.se/libcurl/c/libcurl-errors.html))  
```

That is because we are trying to connect to http://169.254.169.254/latest/meta-data/iam/security-credentials/

(On my CI server I get 404)

I guess the AWS client is checking my default profile in ´~./aws/credentiials` and verify those. 

---------------

This PR allows you to specify environment variables so we never have to go to the IAM service. 

```
AWS_KEY=foo AWS_SECRET=bar vendor/bin/bref cli <function-name>
```

### DIFF
--- a/src/Lambda/SimpleLambdaClient.php
+++ b/src/Lambda/SimpleLambdaClient.php
@@ -21,8 +21,8 @@ final class SimpleLambdaClient
             'region' => $region,
         ];
 
-        $awsKey = getenv('AWS_KEY');
-        $awsSecret = getenv('AWS_SECRET');
+        $awsKey = getenv('AWS_ACCESS_KEY_ID');
+        $awsSecret = getenv('AWS_SECRET_ACCESS_KEY');
         if ($awsKey && $awsSecret) {
             $args['credentials'] = new Credentials($awsKey, $awsSecret);
         }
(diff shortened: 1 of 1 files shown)


mnapoli:
Thanks! That sounds reasonable, however I have some doubts about the name of the environment variables: both the `aws` CLI tool and `serverless` use the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` variables instead.

(see https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html#config-settings-and-precedence and https://serverless.com/framework/docs/providers/aws/guide/credentials/#using-aws-access-keys)

Nyholm:
Thank you. I’ve updated the pr

mnapoli:
Thanks! After checking though it seems those variables are already checked by the PHP SDK ^^ (see https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/guide_credentials_environment.html)

And the environment variables are checked in priority over other authentication methods: https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/guide_credentials.html

So with those variables it should work already?

Nyholm:
Yes, it does! 

I will send a PR to the docs instead. Thanks!

PR #357, opened 2019-06-23
Remove cluster client

Nyholm:
I want to remove this for the following reasons: 

- It does not work when adding `extension=amazon-elasticache-cluster-client.so`
- It does not support PHP 7.2 or 7.3
- It is only required with Memcached but we dont provide the Memcached extensions. 

Users should consider using Redis.


mnapoli:
@bubba-h57 IIRC you added this extension to the layer. What do you think? TBH I've never used or even tested this extension.

mnapoli:
(and by the way I am all for removing things we can do without)

bubba-h57:
I approve of its removal.

mnapoli:
🚀 	thank you both!

Nyholm:
Thank you for merging

PR #359, opened 2019-06-23
Use only `/var/task/php/conf.d`

### LINKED ISSUE #355: Update path for with-config-file-scan-dir
We currently write: 
```   
--with-config-file-scan-dir=${INSTALL_DIR}/etc/php/conf.d:/var/task/php/conf.d 
```

This creates a folder in my docker image as `/opt/bref/etc/php/conf.d:/var/task/php/conf.d`. That is clearly not what we want. I **think** the correct solution is to write: 

```   
--with-config-file-scan-dir=${INSTALL_DIR}/etc/php/conf.d
--with-config-file-scan-dir=/var/task/php/conf.d 
```

But Im not sure how I would verify that.

Nyholm:
This will fix #355

### DIFF
--- a/runtime/php/php.Dockerfile
+++ b/runtime/php/php.Dockerfile
@@ -392,7 +392,7 @@ RUN set -xe \
         --enable-maintainer-zts \
         --enable-sockets \
         --with-config-file-path=${INSTALL_DIR}/etc/php \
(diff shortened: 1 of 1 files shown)


mnapoli:
Closing per https://github.com/brefphp/bref/issues/355#issuecomment-515744182

PR #360, opened 2019-06-27
Document the bref/logger package

mnapoli:
This PR documents and makes official the [`bref/logger`](https://github.com/brefphp/logger) package.

This package is a PSR-3 logger designed for AWS Lambda. In other words, it logs everything into `stderr` so that logs end up in CloudWatch.

This is a lightweight alternative to Monolog, and it is ready to use without having to configure Monolog handlers. It is also a bit more limited as it has no extension points.

### DIFF
--- a/docs/environment/logs.md
+++ b/docs/environment/logs.md
@@ -10,18 +10,24 @@ As explained in the [storage documentation](storage.md), the filesystem on AWS L
 - not shared between lambda instances
 - not persistent
 
-Because of that application logs should not be stored on disk.
+Because of that, logs should not be stored on disk.
 
 ## CloudWatch
 
 The simplest solution is to push logs to AWS CloudWatch, AWS' service for logs.
 
+### PHP errors and warnings
+
+By default, all PHP errors, warnings and notices emitted by PHP will be forwarded into CloudWatch.
+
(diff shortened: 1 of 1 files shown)


docs/environment/logs.md
```diff
@@ -10,18 +10,24 @@ As explained in the [storage documentation](storage.md), the filesystem on AWS L
...
 
 ## CloudWatch
 
 The simplest solution is to push logs to AWS CloudWatch, AWS' service for logs.
 
+### PHP errors and warnings
+
+By default, all PHP errors, warnings and notices emitted by PHP will be forwarded into CloudWatch.
+
+That means that you don't have to configure anything to log errors, warnings or uncaught exceptions.
```
mnapoli: This is a clarification on the current behavior regarding errors.

miketheman:
I like the idea of a Cloudwatch specific logger.
Have you considered making the behavior configurable via env var, like other loggers?
I read https://laravel.com/docs/5.8/logging#configuration and thought it a clean approach for local dev vs deployed environment.

mnapoli:
@miketheman When running locally with SAM or serverless it should work (because stderr is displayed in the console).

When running locally with `php -S` it will also work AFAIK.

So it should be good here!

PR #363, opened 2019-06-29
Updating link to use correct href

nathanczachur:
WHY
Current link skips a page in the "Getting Started" section and takes the user to the runtime functions page.

WHAT
Change link to expected one


mnapoli:
Thanks!

PR #364, opened 2019-07-01
add MyBuilder workers to case studies page

nealio82:
Added MyBuilder's case study, with a brief description of our use-case and a link to our tech article on creating our own layer to use alongside Bref's own layers


mnapoli:
Awesome thanks!

Issue #332, opened 2019-05-14
Upgrade to the new AWS Linux AMI 2018.03


mnapoli:
AWS announced that Lambda is migrating to a new version of their Linux AMI: https://aws.amazon.com/fr/blogs/compute/upcoming-updates-to-the-aws-lambda-execution-environment/

We need to update the layers so that we compile PHP against the 2018.03 AMI.

deleugpn:
Maybe go straight to Amazon Linux 2?

mnapoli:
This is not mentioned in the blog post right? I would make more sense to target the AMI that will actually be used in production.

deleugpn:
You're definitely right. I noticed NodeJS 10.x will be running on Amazon Linux 2 and thought we could do the same, but we don't have control over which OS is being used by the custom Runtime. For some reason I thought that if we compiled PHP using the AL2, the layer would use AL2.

atrope:
@mnapoli do we Must do that? I've just opted in to the new AMI and everything is working as before.

What can we do to test that we must recompile PHP for this AMI? What would be the benefits?

mnapoli:
If everything is working that is perfect!

I would say that if nothing breaks that's a good sign. However we should still update the references to `2017.03` into `2018.03` in the compilation scripts (just to be safe).

mnapoli:
We gave it a try with @Guillaume-Rossignol but the compilation of PHP fails, which is very weird:

```
[...]

PR #367, opened 2019-07-04
Run applications locally with Bref Docker images

Guillaume-Rossignol:
The goal of this pull request is to create Docker images that replicate the AWS Lambda environment locally. That will allow us to have a local development environment that is easier to use and faster than `sam local` or `serverless invoke local`.


runtime/php/Makefile
```diff
@@ -7,7 +7,9 @@ compiler: compiler.Dockerfile
 # Compile PHP and its extensions
 build: compiler
 	docker build -f ${PWD}/php.Dockerfile -t bref/php-72:latest $(shell helpers/docker_args.sh versions.ini php72) .
+	docker build -f ${PWD}/php-dev.Dockerfile -t bref/php-72:dev --build-arg LAYER_IMAGE=bref/php-72:latest .
```
mnapoli: `:dev` is the version (aka [tag name](https://www.freecodecamp.org/news/an-introduction-to-docker-tags-9b5395636c2a/)). I think it should be `bref/php-72-dev` because it's a different image from `bref/php-72`.

runtime/php/readme.md
```diff
@@ -35,3 +34,48 @@ BREF_VERSION ?= 0.3.0
...
+
+
+In your project directory, create a `docker-compose.yml` : 
+```yaml
+web:
+    image: nginx:latest
+    ports:
+        - "8080:80"
+    volumes:
+        - .:/code
```
mnapoli: I think this should be `/var/task` to have the same directory as production.

runtime/php/readme.md
```diff
@@ -35,3 +34,51 @@ BREF_VERSION ?= 0.3.0
...
+            location ~ (/src/index.php|/src/test.php) {
+              fastcgi_split_path_info ^(.+\.php)(/.+)$$;
+              fastcgi_pass php:9000;
+              include fastcgi_params;
+              fastcgi_param SCRIPT_FILENAME $$document_root$$fastcgi_script_name;
+              fastcgi_param PATH_INFO $$fastcgi_path_info;
+            }
+          }
+    command:
+        /bin/sh -c "echo \"$$NGINX_CONFIG\" > /etc/nginx/conf.d/default.conf; nginx -g \"daemon off;\""
```
mnapoli: What we plan to do:

```yaml
http:
    image: bref/php-72-fpm-dev #or bref/php-73-dev:latest
    environment:
        HANDLER: public/index.php
        DOCUMENT_ROOT: public
    volumes:
        - .:/var/task:ro # Read only, like a lambda function
        # - ./var/cache:/var/task/var/cache # You can make subfolder writable, but you should generate cache before uploading on lambda
```

runtime/layers/fpm-dev/Dockerfile
```diff
@@ -0,0 +1,9 @@
+ARG LAYER_IMAGE=bref/php-73:latest
+FROM $LAYER_IMAGE
+
+WORKDIR /opt/bin
+EXPOSE 9000
+
+COPY ./local/php-fpm.conf /opt/bref/etc/php-fpm.conf
+
+CMD [ "php-fpm", "-F", "-y", "/opt/bref/etc/php-fpm.conf" ]
```
mnapoli: Posting here so that we don't forget: we should populate the environment variables available on Lambda: https://docs.aws.amazon.com/lambda/latest/dg/lambda-environment-variables.html

runtime/compiler.Dockerfile
```diff
@@ -8,7 +8,7 @@
 # https://docs.aws.amazon.com/lambda/latest/dg/current-supported-versions.html
 # AWS has kindly provided us with it as a base docker image.
 # https://github.com/aws/amazon-linux-docker-images/tree/2017.03
-FROM amazonlinux:2017.03
+FROM amazonlinux:2
```
mnapoli: Waiting on #332 to use `2018.03` instead.

runtime/layers/fpm-dev/Dockerfile
```diff
@@ -0,0 +1,6 @@
+ARG LAYER_IMAGE
+FROM $LAYER_IMAGE
+
+COPY ./php-fpm.conf /opt/bref/etc/php-fpm.conf
+EXPOSE 9000
+CMD /opt/bin/php-fpm -F -y /opt/bref/etc/php-fpm.conf
```
mnapoli: Could you use the readable version for the options? Like here:

https://github.com/brefphp/bref/blob/7127de2e15abe1f8bd3235adc4ba30b64076172f/src/Runtime/PhpFpm.php#L68

That way we can understand easily what they mean.

mnapoli:
Given this `docker-compose.yml` file:

```yaml
version: "2.1"
services:
    cli:
        image: bref/php-73
        volumes:
            - .:/var/task
        command: 'false'
```

Then:

```bash
$ docker-compose run cli php bin/console
Could not open input file: console

$ docker-compose run cli php /var/task/bin/console
# works
# ...
```

Maybe the default directory should be `/var/task`? (I don't know if it's possible)

mnapoli:
Trying to make externals.io use this: https://github.com/mnapoli/externals/pull/119

PR #373, opened 2019-07-19
Update runtime to Amazon Linux 2018.03

TheNodi:
See https://github.com/brefphp/bref/issues/332#issuecomment-512911000

### DIFF
--- a/runtime/php/compiler.Dockerfile
+++ b/runtime/php/compiler.Dockerfile
@@ -27,7 +27,9 @@ RUN set -xe \
 # Download yum repository data to cache
  && yum makecache \
 # Default Development Tools
- && yum groupinstall -y "Development Tools"  --setopt=group_package_types=mandatory,default
+ && yum groupinstall -y "Development Tools"  --setopt=group_package_types=mandatory,default \
+# Upgrade gcc
+ && yum install -y gcc72 gcc72-c++
 
 
 # CMAKE - cross-platform family of tools designed to build, test and package software. The
--- a/runtime/php/compiler.Dockerfile
+++ b/runtime/php/compiler.Dockerfile
@@ -2,13 +2,13 @@
 # for AWS Lambda. Nothing installed or built in this container should ever
 # be packaged into a Lambda Task nor Layer.
 #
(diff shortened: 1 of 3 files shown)


mnapoli:
Awesome, I have tested and built the layers and it works fine!

runtime/php/compiler.Dockerfile
```diff
@@ -19,15 +19,17 @@ LABEL home="https://github.com/brefphp/bref"
...
 
 
 RUN set -xe \
 # Download yum repository data to cache
  && yum makecache \
 # Default Development Tools
- && yum groupinstall -y "Development Tools"  --setopt=group_package_types=mandatory,default
+ && yum groupinstall -y "Development Tools"  --setopt=group_package_types=mandatory,default \
+# Upgrade gcc
+ && yum install -y gcc72 gcc72-c++
```
mnapoli: I wonder if this could simply be

```suggestion
 && yum install -y gcc72-c++
```

since only `gcc72-c++` is missing.

TheNodi: I though about it too, my main concern is a future change. Since gcc72 is not directly installed by `Development Tools` but pulled in because of `kernel-devel`, if in the future they change something (unluckily) we might end up with the reverse problem of having gcc48 and gcc72-c++. 

This at least sets a "base level" for the compilers, even in case they drop gcc72 dependency, we would still be using 7.2. An upgrade to 8 or 9 would really be unexpected.

At the end, it's in the build process so someone would catch it anyway and it would not break end users features, but he might end wasting time trying to figure out the error from scratch. I think this might be fail-safe(er) and quite self-explanatory in case of future compiler changes.

mnapoli: Understood.

> Since gcc72 is not directly installed by Development Tools but pulled in because of kernel-devel

That's a good point, let's leave it there.

mnapoli:
I have opened https://github.com/TheNodi/bref/pull/1 to add a comment. Let me know if anything I've written in there is wrong. Once merged this PR will be updated with my commit and we'll be good to go.

TheNodi:
Looks clear to me, merged.

PR #377, opened 2019-07-21
Increase function timeout for console applications

### LINKED ISSUE #261: Increase the default timeout for console applications
By default lambdas timeout after 3 seconds.

For the console layer this isn't much. We could set it for example to 30 seconds, just like [in HTTP applications](https://github.com/mnapoli/bref/blob/a22adc8002ac34c2573c1ee070961f6a8fea6f6a/template/http/template.yaml#L13).

The file to edit would be [`template/console/template.yaml`](https://github.com/mnapoli/bref/blob/master/template/console/template.yaml).

Luitame:
The default timeout it's no much for console applications. It was increased until the current [API Gateway max timeout (30s)](https://serverless.com/framework/docs/providers/aws/guide/serverless.yml/). 

This PR will fix #261

### DIFF
--- a/template/console/serverless.yml
+++ b/template/console/serverless.yml
@@ -12,7 +12,7 @@ functions:
     console:
         handler: bin/console # Replace by `artisan` if you are using Laravel
         description: ''
-        timeout: 30 # in seconds (API Gateway has a timeout of 30 seconds)
+        timeout: 30 # in seconds
         layers:
             - ${bref:layer.php-73} # PHP
             - ${bref:layer.console} # The "console" layer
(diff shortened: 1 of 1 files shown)


template/console/serverless.yml
```diff
@@ -12,6 +12,7 @@ functions:
     console:
         handler: bin/console # Replace by `artisan` if you are using Laravel
         description: ''
+        timeout: 30 # in seconds (API Gateway has a timeout of 30 seconds)
```
mnapoli: Since we are running a console command, we aren't using API Gateway. But 30 seconds is a good timeout anyway, we just need to edit the comment, for example:

```suggestion
        timeout: 30 # in seconds
```

PR #378, opened 2019-07-22
Compile php using GCC -O3

TheNodi:
I've looked around too see if there's a reason why we are compiling PHP for size instead of speed, but I haven't found any motivation in the commits/issues so I tested out the options. Especially because compiler optimization can be quite good in an interpreter.

Size-optimized compilation outputs two binaries (`php` and `php-fpm`) of around 13M each. The output directory weights 67M, which compresses down 35M in the layer zip file.
Tier 3 optimization produces 15M binaries, 71M directory which compresses down to 36M.

I tested out the "language" performance using [Zend/bench.php](https://github.com/php/php-src/blob/master/Zend/bench.php) on 10 runs (AMD FX-8350).
Size-optimized compilation executes the benchmark in 0.96, while Tier 3 optimization executes it in 0.86. 

It is a ~10% improvement in interpreter performance, against a ~6% total layer increase.

I know this is quite borderline, and it applies especially to CPU-bound programs, but in a lambda environment where we can run with a fraction of 1 vCPU it might lead to a significant improvement (and cost saving). 
The main example that comes into my mind is hydration of database models (e.g. Laravel Eloquent, Symfony Doctrine), which is basically CPU bound.

Similar consideration can be done for other libraries (we're not setting optimization level for most of them, so they either have a default one in the config or we might be using `-O0`), but I left them out of here for now.

### DIFF
--- a/runtime/php/php.Dockerfile
+++ b/runtime/php/php.Dockerfile
@@ -371,7 +371,7 @@ RUN LD_LIBRARY_PATH= yum install -y readline-devel gettext-devel libicu-devel
 # -fstack-protector-strong : Be paranoid about stack overflows
 # -fpic : Make PHP's main executable position-independent (improves ASLR security mechanism, and has no performance impact on x86_64)
 # -fpie : Support Address Space Layout Randomization (see -fpic)
-# -0s : Optimize for smallest binaries possible.
+# -O3 : Optimize for fastest binaries possible.
 # -I : Add the path to the list of directories to be searched for header files during preprocessing.
 # --enable-option-checking=fatal: make sure invalid --configure-flags are fatal errors instead of just warnings
 # --enable-ftp: because ftp_ssl_connect() needs ftp to be compiled statically (see https://github.com/docker-library/php/issues/236)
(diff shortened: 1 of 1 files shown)


mnapoli:
Wow that's really interesting, thank you for the details and the numbers.

I think this is a tough call: a 10% performance improvement is definitely interesting. At the same time, the layer size impacts the cold starts duration: the larger a layer is, the longer the cold start.

The disk space on Lambda is also limited to 250Mb, so 70Mb is a good share of that. Ideally, the layer should be as small as possible.

But again, speed!

What do others think?

deleugpn:
Interesting topic and contribution.

I guess the obvious optimal approach would be compile both and let people choose, but I think it would be too much of an overhead, specially for the early-stages of Bref. So, going with the assumption that we won't have 2 tags, I'd vote for speed. My reasoning would be:

- 1MB extra download shouldn't be a huge impact on cold start. 
- Majority of invocations are not cold start.
- 4MB extra on disk space wouldn't be a big deal from my perspective. We cannot rely on local disk at all with Lambda.

nealio82:
Well, looking at what @TheNodi found in #380 I think we can have the best of both worlds!

mnapoli:
Oh yeah awesome! I'll be merging and doing some benchmarks ASAP.

mnapoli:
Results are good! See https://github.com/brefphp/benchmarks/commit/a8cdf9c98fe9780791ce867186a484f109a13ec4

Cold starts seem to be very slighly worse, but there is a lot of noise in those benchmarks because they vary a lot. It might not be actually worse.

PHP is indeed 10% faster for CPU-bound operations, and response times are slightly better (hard to make them better still). It might be related to this PR, or to the fact that I've reworked the benchmarks to make them more reliable.

Anyway, everything is great and green! Thank you!

PR #380, opened 2019-07-22
Remove tests and docs from php library directory

TheNodi:
While testing #378 I found out the cleanup script is leaving out extension's tests and docs files because they are stored in `bref/lib/php`.

Resulting layer dropped from 67M to 58M, compressed version dropped from 33M to 31M.
I tested the redis extension and it worked correctly.


mnapoli:
Awesome! I'll merge, recompile layers and benchmark all of that a little later this week!

mnapoli:
Slights cold starts improvements, benchmark here: https://github.com/brefphp/benchmarks/commit/ce5f280047171ce7f8efc8902fba616b9a9d1330

PR #381, opened 2019-07-23
Escape remote CLI arguments

### LINKED ISSUE #379: bref cli fails with quoted arguments
```
bref cli doctrine-cli-develop -- dbal:run-sql 'select * from events'

  Too many arguments, expected arguments "command" "sql".

dbal:run-sql [--depth DEPTH] [--force-fetch] [--] <sql>
```

This happens no matter where I place the `--` and whether or not I use single quotes or double quotes.

shadowhand:
Fixes #379

### DIFF
--- a/bref
+++ b/bref
@@ -90,6 +90,12 @@ $app->command('init', function (SymfonyStyle $io) {
  */
 $app->command('cli function [--region=] [arguments]*', function (string $function, ?string $region, array $arguments, SymfonyStyle $io) {
     $lambda = new SimpleLambdaClient(($region ?: getenv('AWS_DEFAULT_REGION')) ?: 'us-east-1');
+    
+    // Because arguments may contain spaces, and are going to be executed remotely
+    // as a separate process, we need to escape all arguments.
+    $arguments = array_map(static function (string $arg): string {
+        return escapeshellarg($arg);
+    }, $arguments);
 
     try {
         $result = $lambda->invoke($function, json_encode([


bref
```diff
@@ -90,6 +90,12 @@ $app->command('init', function (SymfonyStyle $io) {
  */
 $app->command('cli function [--region=] [arguments]*', function (string $function, ?string $region, array $arguments, SymfonyStyle $io) {
     $lambda = new SimpleLambdaClient(($region ?: getenv('AWS_DEFAULT_REGION')) ?: 'us-east-1');
+    
+    // Because arguments may contain spaces, and are going to be executed remotely
+    // as a separate process, we need to escape all arguments.
+    $arguments = array_map(static function (string $arg): string {
+        return escapeshellarg($arg);
+    }, $arguments);
```
shadowhand: This *should* work in all situations, because the quoting is lost when passed through Symfony's input layer. Since the invocation on the Lambda side does not use Symfony console, this should be correct. And it is definitely more secure.

mnapoli:
Thanks!

shadowhand:
@mnapoli can this be tagged into a release? It is hampering our ability to run `doctrine` console, among other things.

mnapoli:
I've been building layers and running benchmarks all evening yesterday, and I am continuing that as we speak :)

I'm looking into including #378 into the release.

shadowhand:
Ah right, I forgot that `bref` releasing is not as simple as tag and push. 👍

mnapoli:
Well to be fair with all the improvements we've setup lately it's been simpler (I just have one command to run on my laptop), but updating benchmarks is a bit more work ^^

PR #383, opened 2019-07-29
 Run applications locally with Bref Docker images

Guillaume-Rossignol:


Guillaume-Rossignol:
About https://github.com/brefphp/bref/pull/367#issuecomment-515699594

You can try with this docker-compose.yml 
```
version: "2.1"
services:
        cli:
                image: bref/php-73:latest
                volumes:
                        - .:/var/task
                entrypoint: php
```
and execute with `docker-compose run cli bin/console`

mnapoli:
I've updated the documentation, thanks @Guillaume-Rossignol!

Now we're working on auto-publishing those Docker layers so that they can be used. Merging to facilitate our work on that.

PR #387, opened 2019-07-31
Documentation: Performances

mnapoli:
This PR introduces a new documentation page to explain performances.

TODO:

- [x] warm response time (overhead) for HTTP and PHP functions
- [x] lambda memory size impacting the CPU power
- [x] cold starts
- [x] add the page in the menu


docs/environment/performances.md
```diff
@@ -0,0 +1,76 @@
...
+**It is recommended to use 1024M** for PHP applications, or at least to start with that. This is what Serverless configures by default.
+
+To customize the amount of memory, set the `memory` option in `serverless.yml`:
+
+```yaml
+functions:
+    foo:
+        handler: index.php
+        # ...
+        memory: 512
```
staabm: should the code mention the recommended default instead?

```suggestion
        memory: 1024
```

mnapoli: I think it's better to show how to customize the value with something that is not the default, else it isn't really an example of customizing anything, and this might confuse readers (I know I'm easily confused by this).

What I can do is add a comment on the line:

```suggestion
        memory: 512 # set to 512M instead of 1024M (the default)
```

docs/environment/performances.md
```diff
@@ -0,0 +1,76 @@
...
+
+Since this runtime is often used in asynchronous scenarios (processing queue messages for example), it is often negligible.
+
+## CPU power and memory size
+
+The main factor that affects performances is a lambda's memory size. Indeed, [the amount of memory is proportional to the CPU power available](https://docs.aws.amazon.com/lambda/latest/dg/resource-model.html).
+
+In other words, **more memory means a more powerful CPU**. A 1024M lambda has a CPU 2 times more powerful than a 512M lambda.
+
+From 64M to 1,792M, applications run with one CPU (1,792M gives 1 full CPU). From 1,856M to 3,008M, applications run with 2 CPU. However, since PHP is single threaded and one lambda handles only 1 request at a time, using 2 CPU usually does not provide any benefit.
```
Nyholm: This is not super clear to me. I know that 896M gives me 0.5CPU. When I first read this I intrepid it as "everything between 64 and 1792 gives to 1 CPU". 

Maybe write: 

>  From 64M to 1,792M, applications run with up to one CPU (1,792M gives 1 full CPU)

docs/environment/performances.md
```diff
@@ -0,0 +1,76 @@
...
+### Costs
+
+AWS Lambda bills the number of events + the execution time. The more memory configured for a lambda, [the more expensive is the execution time](https://aws.amazon.com/lambda/pricing/).
+
+It might be tempting to lower the memory to save money. However a function might run slower on a smaller lambda, cancelling the costs savings. Here is an example where both lambdas cost the same thing:
+
+- a function runs in 400ms on a 512M lambda
+- the same function runs in 200ms on a 1024M lambda because of the faster CPU
+
+In general, use smaller and slower lambdas only if speed is not important at all.
```
Nyholm: We could also mention that lambda bills per 100ms. 
101ms is twice as expensive as 100ms.

docs/environment/performances.md
mnapoli: :+1: thanks, I struggled with a clear explanation here and I prefer your version.

docs/environment/performances.md
mnapoli: I want to make a separate article to explain pricing, this will be a better fit there. Here what I mention is targeted at performances.

mnapoli:
This PR is finally done!

I will probably be merging and publishing tomorrow.

PR #389, opened 2019-08-02
Run tests with CirceCI

mnapoli:
There are so many problems with Travis, I'm giving this a try.


mnapoli:
That's what we call a failure, I'm wasting too much time on this. I cannot get docker to work correctly inside docker, and given the limitations I've read in the docs I'm not even sure it will work eventually.

PR #392, opened 2019-08-05
Add command to start dashboard

Guillaume-Rossignol:
Simplify access to the dashboard.

### DIFF
--- a/bref
+++ b/bref
@@ -5,6 +5,7 @@ declare(strict_types=1);
 use Aws\CloudFormation\CloudFormationClient;
 use Aws\CloudFormation\Exception\CloudFormationException;
 use Bref\Console\LoadingAnimation;
+use Bref\Console\OpenUrl;
 use Bref\Lambda\InvocationFailed;
 use Bref\Lambda\SimpleLambdaClient;
 use Symfony\Component\Console\Style\SymfonyStyle;
@@ -336,7 +337,9 @@ $app->command('dashboard [--port=]', function (int $port = 8000, SymfonyStyle $i
 
         return 1;
     }
-    $io->writeln("Dashboard started: <fg=green;options=bold,underscore>http://localhost:$port</>");
+    $url = "http://localhost:$port";
+    $io->writeln("Dashboard started: <fg=green;options=bold,underscore>$url</>");
+    OpenUrl::open($url);
     $process->wait(function ($type, $buffer) {
         if (Process::ERR === $type) {
             echo 'ERR > '.$buffer;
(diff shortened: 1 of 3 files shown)


bref
```diff
@@ -248,4 +248,60 @@ $app->command('deployment stack-name [--region=]', function (string $stackName,
...
+        );
+
+        return 1;
+    }
+
+    $servelessInfo = new Process(['serverless', 'info']);
+    $servelessInfo->run();
+
+    if (!$servelessInfo->isSuccessful()) {
+        $io->error('The command `serverless info` failed' . PHP_EOL . $servelessInfo->getErrorOutput());
```
mnapoli: Here we need to `return 1` I think

bref
```diff
@@ -248,4 +248,60 @@ $app->command('deployment stack-name [--region=]', function (string $stackName,
...
+    preg_match('/region: ([a-z0-9-]*)/', $servelessInfoOutput, $region);
+    $region = $region[1];
+
+    $stack = [];
+    preg_match('/stack: ([a-zA-Z0-9-]*)/', $servelessInfoOutput, $stack);
+    $stack = $stack[1];
+
+    $io->writeln("Region : $region, Stack : $stack");
+
+    $process = new Process(['docker', 'run', '--rm', '-p', $port.':8000', '-v', getenv('HOME').'/.aws:/root/.aws:ro', '--env', 'STACKNAME='.$stack, '--env', 'REGION='.$region, 'bref/dashboard']);
```
mnapoli: In the future (https://github.com/docker/cli/pull/1498#issuecomment-505974526) there will be a `docker run --pull` command.

Right now we could do:

- `docker pull`
- `docker run`

to always have the latest version running.

mnapoli:
TODO for me: document what this is :)

bref
```diff
@@ -248,4 +250,105 @@ $app->command('deployment stack-name [--region=]', function (string $stackName,
     return 0;
 })->descriptions('Displays the latest deployment logs from CloudFormation. Only the logs from the last 24 hours are displayed. Use these logs to debug why a deployment failed.');
 
+$app->command('dashboard [--port=]', function (int $port = 8000, SymfonyStyle $io) {
+    if (! file_exists('serverless.yml')) {
```
Guillaume-Rossignol: The `serverless info` at line  289 don't make the job ?

mnapoli: In case someone uses SAM, they probably won't be `serverless` installed. So the error message they will get is: "The `serverless` command is not installed…". If they install it it will still not work.

I was thinking it's better to explain clearly that you need to be using `serverless` entirely, not just have it installed.

Issue #394, opened 2019-08-06
Throw an exception if event is not provided


linuxjuggler:
In the default function file which is generated by bref the code is something like this

```php
<?php declare(strict_types=1);

require __DIR__.'/vendor/autoload.php';

lambda(function (array $event) {
    return 'Hello ' . ($event['name'] ?? 'world');
});

```

and executing the following local command works like a charm 

```shell
serverless invoke local --docker -f function --data '{"name": "Jane"}'
```

but if the user didn't supply the data argument 

```shell
serverless invoke local --docker -f function
```

the system will throw the following exception 

```shell
Fatal error: Argument 1 passed to {closure}() must be of the type array, string given, called in /var/task/vendor/bref/bref/src/Runtime/LambdaRuntime.php on line 92 in /var/task/index.php:7
Stack trace:
#0 /var/task/vendor/bref/bref/src/Runtime/LambdaRuntime.php(92): {closure}('', Object(Bref\Context\Context))
#1 /var/task/vendor/bref/bref/src/functions.php(17): Bref\Runtime\LambdaRuntime->processNextEvent(Object(Closure))
#2 /var/task/index.php(9): lambda(Object(Closure))
#3 {main}
```

To solve the issue we can do either:

1. remove the `$event` from the closure by default and instruct the user to add it when needed, like what is done for the `$context`
2. Validate the data passed to the function in the run time file `https://github.com/brefphp/bref/blob/master/src/Runtime/LambdaRuntime.php#L86`, adding something like this

```php
    public function processNextEvent(callable $handler): void
    {
        /** @var Context $context */
        [$event, $context] = $this->waitNextInvocation();

        if(!is_array($event) && empty($event)) {
            $event = [];
        }

        try {
            $this->sendResponse($context->getAwsRequestId(), $handler($event, $context));
        } catch (\Throwable $e) {
            $this->signalFailure($context->getAwsRequestId(), $e);
        }
    }
```

I am with the second option, and I am happy to do a small PR for any of the suggested solutions.

mnapoli:
Thanks, that is a very good point!

Option 1 would work but it would make the "developer experience" a little bit worse.

Option 2 is interesting, however I would like to keep the behavior predictable. What I mean is that if an empty event is submitted, then it should be `null`, not an empty array (those are different things).

Maybe a 3rd option would be to remove the `array` type-hint from the default function?

[...]

PR #395, opened 2019-08-06
Fix issue #394

linuxjuggler:
As mentioned in https://github.com/brefphp/bref/issues/394, using the array type hint will throw an exception if the user didn't provide any data to the request.

This removes the type hint from the default generated code


mnapoli:
Thank you!

Fixes #394

Issue #408, opened 2019-08-10
Offer PHP 7.4 beta layers


ondrejmirtes:
Hi,
I'd love to have PHP 7.4 beta layers so I could run phpstan.org with them and analyse for example code with native property types. If you don't want to build them yourself, I'd like some pointers how I could do it myself :) Thank you!

mnapoli:
:+1: let's do this! You can find the code related to the runtimes here if you want to look at it: https://github.com/brefphp/bref/tree/master/runtime I'm not sure how much work that entails ^^

ondrejmirtes:
OK, so this is my last checklist item before I'm able to release new PHPStan version with PHP 7.4 support, so I'm looking into it now :)

PR #415, opened 2019-08-13
PHP 7.4 layers

ondrejmirtes:
#408 - this is my current progress. I'm really not familiar with the build systems so it's possible I've been adding the steps in wrong places.

Once I modified Makefile and versions.ini, the build failed on the following errors:

* bison required in version 3.0
* re2c required
* sqlite3 required - not sure why PHP 7.3 works without it, maybe I'm doing something wrong
* oniguruma required

I'm stuck on the last bullet point. No matter what I do - install oniguruma through yum, install it in `/usr/local`, install it in `/opt/bref/lib/pkgconfig` (`PKG_CONFIG_PATH`), the build still fails on this:

```
checking for oniguruma... no
configure: error: Package requirements (oniguruma) were not met:

No package 'oniguruma' found

Consider adjusting the PKG_CONFIG_PATH environment variable if you
installed software in a non-standard prefix.

Alternatively, you may set the environment variables ONIG_CFLAGS
and ONIG_LIBS to avoid the need to call pkg-config.
See the pkg-config man page for more details.
The command '/bin/bash -c set -xe  && ./buildconf --force  && CFLAGS="-fstack-protector-strong -fpic -fpie -O3 -I${INSTALL_DIR}/include -I/usr/include -ffunction-sections -fdata-sections"     CPPFLAGS="-fstack-protector-strong -fpic -fpie -O3 -I${INSTALL_DIR}/include -I/usr/include -ffunction-sections -fdata-sections"     LDFLAGS="-L${INSTALL_DIR}/lib64 -L${INSTALL_DIR}/lib -Wl,-O1 -Wl,--strip-all -Wl,--hash-style=both -pie"     ./configure         --build=x86_64-pc-linux-gnu         --prefix=${INSTALL_DIR}         --enable-option-checking=fatal         --enable-maintainer-zts         --enable-sockets         --with-config-file-path=${INSTALL_DIR}/etc/php         --with-config-file-scan-dir=${INSTALL_DIR}/etc/php/conf.d:/var/task/php/conf.d         --enable-fpm         --disable-cgi         --enable-cli         --disable-phpdbg         --disable-phpdbg-webhelper         --with-sodium         --with-readline         --with-openssl         --with-zlib=${INSTALL_DIR}         --with-zlib-dir=${INSTALL_DIR}         --with-curl         --enable-exif         --enable-ftp         --with-gettext         --enable-mbstring         --with-pdo-mysql=shared,mysqlnd         --with-mysqli         --enable-pcntl         --with-zip         --enable-bcmath         --with-pdo-pgsql=shared,${INSTALL_DIR}         --enable-intl=shared         --enable-soap' returned a non-zero code: 1
make: *** [build] Error 1
```

So I'm not sure what to do here.

More interesting findings:

* I tried libssh2 version 1.9.0 but it failed on some weird error I wasn't able to Google anything about.
* `--enable-zip` was modified to `--with-zip` in PHP 7.4. I don't know how you want to write the condition so that builds for both 7.3 and 7.4 work.
* Same for `--enable-opcache-file` - that one was replaced with opcache.file_cache INI directive.


runtime/php-intermediary.Dockerfile
```diff
@@ -411,11 +411,11 @@ RUN set -xe \
         --with-pdo-mysql=shared,mysqlnd \
         --with-mysqli \
         --enable-pcntl \
-        --enable-zip \
+        --with-zip \
         --enable-bcmath \
         --with-pdo-pgsql=shared,${INSTALL_DIR} \
         --enable-intl=shared \
-        --enable-opcache-file \
+        # --enable-opcache-file \ replaced with opcache.file_cache INI directive
```
nikic: You can just drop this, the option only makes sense in the `--disable` form, as it was enabled by default.

runtime/compiler.Dockerfile
```diff
@@ -52,3 +52,33 @@ RUN  set -xe \
  && ./bootstrap --prefix=/usr/local \
  && make \
  && make install
+
+# BISON parser - 3.0 - required by PHP
```
nikic: Bison and re2c are only required when building from git, not when using release tarballs. Assuming this is not building from git, these *shouldn't* be necessary (though maybe some checks are broken?)

runtime/compiler.Dockerfile
```diff
@@ -52,3 +52,33 @@ RUN  set -xe \
...
+ && make \
+ && make install
+
+# oniguruma - required by PHP
+RUN  set -xe \
+ && mkdir -p /tmp/oniguruma \
+ && cd /tmp/oniguruma \
+ && curl -Ls https://github.com/kkos/oniguruma/releases/download/v6.9.3/onig-6.9.3.tar.gz \
+  | tar xzC /tmp/oniguruma --strip-components=1 \
+ && ./configure --prefix=/opt/bref/lib/pkgconfig \
```
nikic: Using `lib/pkgconfig` as prefix doesn't make a lot of sense. This should be either installed into some standard location that is on the pkg config path (I'd assume using `/usr/local` would work), or else you need to add it to the PKG_CONFIG_PATH.

ondrejmirtes: `/usr/local` didn't work... I'll try to add it to PKG_CONFIG_PATH.

runtime/compiler.Dockerfile
ondrejmirtes: It's built from `https://github.com/php/php-src/archive/php-7.4.0beta2.tar.gz`. And these checks failed, otherwise I wouldn't be adding bison and re2c.

It's also weird to me that previous versions were able to be built without `sqlite-devel` package installed but 7.4 requires it.

Feel free to try it - clone my branch with these changes and run `make build` inside the `runtime/` directory. Remove these lines installing bison and re2c beforehand and you should get to the same error about these dependencies...

ondrejmirtes:
I tried the beta4 and it crashes on this if I don't install bison explicitly:

```
checking for bison version... 2.7
configure: error: bison 3.0.0 or later is required to generate PHP parsers (excluded versions: none).
```

/cc @nikic

ondrejmirtes:
I need help with this, I'm not able to get unstuck on my own.

mnapoli:
After merging master into this and giving it a try, I get:

> configure: error: unrecognized options: --with-gd, --with-png-dir, --with-jpeg-dir

It seems the `--with-gd` option is gone in PHP 7.4? Or am I missing something?

ondrejmirtes:
The UPGRADING file in the php-src repo talks about these changes: https://github.com/php/php-src/blob/PHP-7.4/UPGRADING#L709-L718

```
- GD:
  . --with-gd becomes --enable-gd (whether to enable the extension at all) and
    --with-external-gd (to opt into using an external libgd, rather than the
    bundled one).
  . --with-png-dir has been removed. libpng is required.
  . --with-zlib-dir has been removed. zlib is required.
  . --with-freetype-dir becomes --with-freetype.
  . --with-jpeg-dir becomes --with-jpeg.
  . --with-webp-dir becomes --with-webp.
  . --with-xpm-dir becomes --with-xpm.
```

mnapoli:
Maintaining all versions is going to be fun 😄

daim2k5:
@ondrejmirtes 
i had yesterday the same error "bison .." and change the download from the github archive to https://downloads.php.net/~derick/php-7.4.0RC1.tar.gz then i got an error with "oniguruma not found"

ondrejmirtes:
@mnapoli Does it compile now? 😊

mnapoli:
No 😄 I got distracted with #467

runtime/compiler.Dockerfile
mnapoli: @nikic we are downloading the PHP source from https://github.com/php/php-src/archive/php-${VERSION_PHP}.tar.gz

It is an archive URL (not the git repository), but is that what you meant by "release tarballs"? The official link in the PHP blog is https://downloads.php.net/~derick/php-${VERSION_PHP}.tar.gz which looks weird, but why not (it's a RC after all).

I gave it a try and now I get an error about oniguruma not being found.

FYI this didn't solve the `oniguruma` problem (currently in the PR oniguruma is compiled from source, but still not detected by PHP):

```diff
- RUN LD_LIBRARY_PATH= yum install -y readline-devel gettext-devel libicu-devel libpng-devel libjpeg-devel sqlite-devel
+ RUN LD_LIBRARY_PATH= yum install -y readline-devel gettext-devel libicu-devel libpng-devel libjpeg-devel sqlite-devel oniguruma oniguruma-devel
```

mnapoli: OK I'm making some progress! PHP is compiling, I am just missing `pecl` now, but it's probably not much.

The fix: eddc5deb37f6a7e734c0b253409bf0f7bb898ee7 (took me a while 😅)

mnapoli:
Current status: `pecl: command not found`.

I think this is because of https://github.com/php/php-src/pull/3781

> - PEAR:
>  . Installation of PEAR (including PECL) is no longer enabled by default. It
>    can be explicitly enabled using --with-pear. This option is deprecated and
>    may be removed in the future.

I will try with the `--with-pear` option, I don't know how to install `pecl` on its own.

mnapoli:
I've managed to build the PHP 7.4 layer!

pthreads is out though, but it will never be compatible with PHP 7.4 (https://github.com/krakjoe/pthreads/issues/929) and it's no longer maintained.

I'll leave it as that for now, we now need to find a way to clean all that up and make it work with other PHP versions at the same time (if possible!).

mnapoli:
Update: I have worked on #490 to refactor the build process. This will help make this PR happen, since we need to have different configuration options between PHP 7.3 and 7.4.

I have updated this branch locally, I will push it once #490 is merged. I managed to compile PHP 7.4 properly and make it work. We are close to merging this!

mnapoli:
Everything compiles correctly, tests are green  🎉 

Here is a summary of what changed (since the diff isn't really readable, there is a new Dockerfile for each PHP version):

- We install the oniguruma library: we compile it, because I couldn't make things work with the system one. I think this is a matter of versions: the system one is 5.9.1, the PHP upgrading guide mentions 6.8.0, we compile 6.9.3. All should be good here.
- pthreads is gone (in 7.4 only): there is no 7.4 support, and the extension is dead and deprecated
- The `sqlite-devel` library is installed because PHP 7.4 does not bundle it anymore
- `./configure` options have been updated following the upgrade guide, nothing major here

I think we should be good here! There isn't anything weird anymore, I feel confident :)

ondrejmirtes:
Looking forward to be able to use the 7.4 layer in serverless.yml :)

mnapoli:
Last thing ^^ Blackfire is not compatible with PHP 7.4 yet, I need to find a solution for the dev Docker image.

ondrejmirtes:
I don’t think that should be a blocker, you can release it and update it as soon as Blackfire is updated.

mnapoli:
Yep, but it blocks compilation of docker images right now. It needs to be un-blocked ;)

jaikdean:
Rather than compiling oniguruma from source, you should only need to install the `oniguruma-devel` or `libonig-dev` package. I've just gone through this with our custom layer which is based on the `lambci/lambda:build-provided` Docker image.

mnapoli:
@jaikdean that's interesting. I added [inline in a code comment](https://github.com/brefphp/bref/pull/415/files#diff-2d3e2ebb1bb2e3f2093da485d046033fR23-R24) that I did not manage to compile PHP with the system one. Which one did you install to make things work: oniguruma-devel or libonig-dev?

jaikdean:
I used `oniguruma-devel`. I believe they are the same package under a different name depending on the distro: `libonig-devel` for Debian derivates and `oniguruma-devel` for RedHat derivatives.

mnapoli:
> I used `oniguruma-devel`. I believe they are the same package under a different name depending on the distro: `libonig-dev` for Debian derivates and `oniguruma-devel` for RedHat derivatives.

I gave it a try, but PHP then fails later when called with the following error message:

```
$ php -v
php: error while loading shared libraries: libonig.so.2: cannot open shared object file: No such file or directory
```

In my try I replaced the compilation of oniguruma with:

```Dockerfile
RUN LD_LIBRARY_PATH= yum install -y oniguruma-devel
```

(the `LD_LIBRARY_PATH= yum install` is the pattern used to install all the other system libraries)

koubas:
Hi, thanks for ongoing work! Are there any preliminary public layers I could put to my serverless.yml and try?

mnapoli:
@koubas I updated the PR and built some layers, here they are:

- `arn:aws:lambda:us-east-1:416566615250:layer:php-74:4`
- `arn:aws:lambda:us-east-1:416566615250:layer:php-74-fpm:4`

Blackfire is OK now. The last thing that is a problem is pthreads: because of the way things work, I have to disable it for all layers. I am not sure if it's worth removing pthreads entirely from Bref.

@bubba-h57 do you still use pthreads?

ondrejmirtes:
Thank you, awesome! But I'd need the layers in eu-west-1, any way to get it there?

mnapoli:
@ondrejmirtes here is `eu-west-1`:

- `arn:aws:lambda:eu-west-1:416566615250:layer:php-74:4`
- `arn:aws:lambda:eu-west-1:416566615250:layer:php-74-fpm:4`

ondrejmirtes:
@mnapoli Thank you, works like a charm! (Of course I could have tried changing the region name in the ID, but I thought that different numbers would be involved.)

shadowhand:
@mnapoli is there also a php-74 console layer? Right now I have this, and I am not sure it is working:

```
arn:aws:lambda:us-west-2:416566615250:layer:console:4
```

mnapoli:
@shadowhand The existing layer should not depend on the PHP version, so you don't have to change anything.

shadowhand:
@mnapoli so I should use `arn:aws:lambda:us-west-2:209497400698:layer:console:14` instead?

mnapoli:
@shadowhand yes.

shadowhand:
@mnapoli this is working great for us! When can we expect `bref/php-74-fpm` docker image to be published?

nealio82:
> Blackfire is OK now. The last thing that is a problem is pthreads: because of the way things work, I have to disable it for all layers. I am not sure if it's worth removing pthreads entirely from Bref.

https://twitter.com/krakjoe/status/1206577478493450240

mnapoli:
Merging and building today!

BTW I stand corrected regarding pthreads: there will be no bc-break.

shadowhand:
@mnapoli when can we expect `bref/php-74-fpm` Docker image?

mnapoli:
Is it not published?

Mer 18 déc 2019, à 21:37, Woody Gilk a écrit :

PR #416, opened 2019-08-14
Add --host parameter to dashboard command (default is localhost)

geschke:
As discussed in the dashboard issue, maybe it's useful to add the --host parameter. Default is "localhost", which converts to "127.0.0.1", because the docker run command needs IPs, not hosts (FQDN). 

Kind regards,
   Ralf


mnapoli:
Thanks!

PR #407, opened 2019-08-10
Change API Gateway/FPM timeouts to 28 seconds

linuxjuggler:
REF: https://github.com/brefphp/bref/issues/401

PR #419, opened 2019-08-18
Add new framework setup example using zend expressive

leonhusmann:
I've added a simple setup example for a Zend Expressive 3 application. This example should help Zend Developers to get started with Bref and experience the serverless world.

Feel free to add or edit any information. Also the demo application can be transferred to another GitHub account if you want to.

### DIFF
--- a/docs/frameworks/expressive.md
+++ b/docs/frameworks/expressive.md
@@ -1,0 +1,70 @@
+---
+title: Serverless Zend Expressive applications
+current_menu: expressive
+introduction: Learn how to deploy serverless Zend Expressive applications on AWS Lambda using Bref.
+---
+
+This guide helps you run Zend Expressive applications on AWS Lambda using Bref. These instructions are kept up to date to target the latest Expressive version.
+
+A demo application is available on GitHub at [github.com/leonhusmann/expressive-bref](https://github.com/leonhusmann/expressive-bref).
+
+## Setup
+
+Assuming your are in existing Zend Expressive project, let's install Bref via Composer:
+
+```
+composer require bref/bref
+```
+
+Then let's create a `serverless.yml` configuration file (at the root of the project) optimized for Symfony:
+
+```yaml
+service: bref-demo-expressive
+
+provider:
+  name: aws
+  region: us-east-1
+  runtime: provided
+
+plugins:
+  - ./vendor/bref/bref
+
+functions:
+  api:
+    handler: public/index.php
+    timeout: 30 # in seconds (API Gateway has a timeout of 30 seconds)
+    layers:
+      - ${bref:layer.php-73-fpm}
+    events:
+      -   http: 'ANY /'
+      -   http: 'ANY /{proxy+}'
+```
+
+Now we still have a few modifications to do on the application to make it compatible with AWS Lambda.
+
+Since [the filesystem is readonly](/docs/environment/storage.md) except for `/tmp` we need to customize where the cache and the logs are stored in the `config/config.php` file. This is done by changing this line:
+
+```diff
+diff --git a/config/config.php b/config/config.php
+index 66df694..a22f39a 100644
+--- a/config/config.php
++++ b/config/config.php
+@@ -9,7 +9,7 @@ use Zend\ConfigAggregator\PhpFileProvider;
+ // To enable or disable caching, set the `ConfigAggregator::ENABLE_CACHE` boolean in
+ // `config/autoload/local.php`.
+ $cacheConfig = [
+-    'config_cache_path' => 'data/cache/config-cache.php',
++    'config_cache_path' => 'tmp/cache/config-cache.php',
+ ];
+
+```
+
+## Deploy
+
+Make sure your not in development mode when deploying.
+
+```shell
+$ composer development-disable
+```
+
+Your application is now ready to be deployed. Follow [the deployment guide](/docs/deploy.md).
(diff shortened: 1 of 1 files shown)


docs/frameworks/expressive.md
```diff
@@ -0,0 +1,70 @@
...
+ $cacheConfig = [
+-    'config_cache_path' => 'data/cache/config-cache.php',
++    'config_cache_path' => 'tmp/cache/config-cache.php',
+ ];
+
+```
+
+## Deploy
+
+Make sure your not in development mode when deploying.
```
skyrpex:
```suggestion
Make sure you're not in development mode when deploying.
```

docs/frameworks/expressive.md
```diff
@@ -0,0 +1,70 @@
...
+introduction: Learn how to deploy serverless Zend Expressive applications on AWS Lambda using Bref.
+---
+
+This guide helps you run Zend Expressive applications on AWS Lambda using Bref. These instructions are kept up to date to target the latest Expressive version.
+
+A demo application is available on GitHub at [github.com/leonhusmann/expressive-bref](https://github.com/leonhusmann/expressive-bref).
+
+## Setup
+
+Assuming your are in existing Zend Expressive project, let's install Bref via Composer:
```
AlessandroMinoccheri: typo in "your are" it should be "you are"

docs/frameworks/expressive.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Setup
+
+Assuming you are in existing Zend Expressive project, let's install Bref via Composer:
+
+```
+composer require bref/bref
+```
+
+Then let's create a `serverless.yml` configuration file (at the root of the project) optimized for Symfony:
```
mnapoli:
```suggestion
Then let's create a `serverless.yml` configuration file (at the root of the project) optimized for Zend Expressive:
```

docs/frameworks/expressive.md
```diff
@@ -0,0 +1,70 @@
...
+Since [the filesystem is readonly](/docs/environment/storage.md) except for `/tmp` we need to customize where the cache and the logs are stored in the `config/config.php` file. This is done by changing this line:
+
+```diff
+diff --git a/config/config.php b/config/config.php
+index 66df694..a22f39a 100644
+--- a/config/config.php
++++ b/config/config.php
+@@ -9,7 +9,7 @@ use Zend\ConfigAggregator\PhpFileProvider;
+ // To enable or disable caching, set the `ConfigAggregator::ENABLE_CACHE` boolean in
+ // `config/autoload/local.php`.
```
mnapoli: I think we can get rid of the code above (we don't need a valid `.diff` file, but instead just show the line to edit). That will be simpler to read.

docs/frameworks/expressive.md
```diff
@@ -0,0 +1,70 @@
...
+ $cacheConfig = [
+-    'config_cache_path' => 'data/cache/config-cache.php',
++    'config_cache_path' => 'tmp/cache/config-cache.php',
```
mnapoli: Should it be rather:

```suggestion
+    'config_cache_path' => '/tmp/cache/config-cache.php',
```

(so that the path is absolute)?

docs/frameworks/expressive.md
```diff
@@ -0,0 +1,70 @@
...
+
+```
+
+## Deploy
+
+Make sure you're not in development mode when deploying.
+
+```shell
+$ composer development-disable
+```
```
mnapoli: I'm not familiar with this command, what does it do? It doesn't seem to exist in Composer itself. Also is it really related to Zend Expressive? I'm not sure it's worth putting it here since the deployment guide covers how to properly deploy dependencies.

docs/frameworks/expressive.md
```diff
@@ -0,0 +1,70 @@
...
+  region: us-east-1
+  runtime: provided
+
+plugins:
+  - ./vendor/bref/bref
+
+functions:
+  api:
+    handler: public/index.php
+    timeout: 30 # in seconds (API Gateway has a timeout of 30 seconds)
```
mnapoli:
```suggestion
    timeout: 28 # in seconds (API Gateway has a timeout of 29 seconds)
```

To match #407 which I just merged.

leonhusmann:
@mnapoli I'm back from vacationing and will try to update the PR soon. As said I'm happy to transport the demo application to any place you would like. Just point me out what I should be doing.

docs/frameworks/expressive.md
leonhusmann: This command comes from the zfcampus/zf-development-mode package and is installed through the official Zend expressive skeleton installation.

I mentioned it here, because it's not enough to just deploy without the dev dependencies like mentioned in the guide when using Whoops in the skeleton. Without this development mode disabled, which is automatically enabled by skeleton installer, you will get the following error `NOTICE: PHP message: PHP Fatal error: Uncaught Error: Class 'Whoops\Run' not found in /var/task/vendor/zendframework/Zend-expressive/src/Container/WhoopsFactory.php: 55`

We could add an explanation that this is only needed if the zf-development-mode package or skeleton is used. What do you think?

Nyholm:
A few months have passed since this PR was opened. I like this, but I would like to suggest some changes. 

1) #521 was just merged. That means that application based on PSR 15 will be special treated (the good kind): https://github.com/brefphp/bref/blob/master/src/Runtime/LambdaRuntime.php#L98. Could you add an example not using FPM?
2) Could you move the example application to https://github.com/brefphp/examples
3) Zend Expressive has been renamed to Mezzio, could you reflect that in this PR and the demo?

mnapoli:
Closing for now.

PR #423, opened 2019-08-22
Doc: improve iamRoleStatements

j0k3r:
The documentation were wrong about how to define custom permissions.

It defines them at the `function` block which isn't possible in
Serverless.
It should be defined at the `provider` block.

I've also added a section to explain how to define them at the `function`
block (using a Serverless plugin).


mnapoli:
Oh right thank you!

Could you also remove the second example (the one with function-level permissions)? I think mentioning and linking to the plugin is enough, that lets us keep the guide as simple as possible. It would also be less confusing: I know people (including me ^^) quickly scan documentation and copy-paste code without always reading all the details. In that case that code wouldn't work out of the box.

j0k3r:
@mnapoli done!

mnapoli:
Thank you for taking the time!

PR #424, opened 2019-08-22
Add AWS profile option when using the dashboard

j0k3r:
~~This is kind of work in progress, because the `bref/dashboard` docker image will also need to be updated to use the profile parameter.~~

Some people use a lot of differents AWS credentials and the Serverless CLI allow to specify which profile to use: https://serverless.com/framework/docs/providers/aws/guide/credentials#using-the-aws-profile-option

In my case, I didn't have a `default` profile but a lot of different one. That's why I needed to be able to define the profile the dashboard command has to use to retrieve the stack.

Without the parameter I got:
> [ERROR] The command `serverless info` failed

~~But now, the dashboard fails to render because the command needs to send him the profile too~~

image

~~And it looks like the repo of the image isn't public (I got a 404) so I can't update it 🤷‍♂  https://github.com/mnapoli/bref-dashboard~~

### DIFF
--- a/bref
+++ b/bref
@@ -250,10 +250,11 @@ $app->command('deployment stack-name [--region=]', function (string $stackName, 
     return 0;
 })->descriptions('Displays the latest deployment logs from CloudFormation. Only the logs from the last 24 hours are displayed. Use these logs to debug why a deployment failed.');
 
-$app->command('dashboard [--host=] [--port=]', function (string $host = 'localhost', int $port = 8000, SymfonyStyle $io) {
+$app->command('dashboard [--host=] [--port=] [--profile=]', function (string $host = 'localhost', int $port = 8000, string $profile = 'default', SymfonyStyle $io) {
     if ($host === 'localhost') {
         $host = '127.0.0.1';
     }
+
     if (! file_exists('serverless.yml')) {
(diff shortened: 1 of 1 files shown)


mnapoli:
Thanks! That's something we are looking into with @Guillaume-Rossignol as well, that PR will definitely help.

FYI the dashboard is not an open source project, it is a project separate from Bref itself (there are more details [here](https://github.com/brefphp/bref/releases/tag/0.5.2)).

Now maybe there isn't much missing for this to work: line 327 there is the `docker run` command. Maybe you could set the `AWS_PROFILE` environment variable to what's passed to `--profile` (if defined)? The AWS SDK automatically picks up that environment variable AFAIK.

mnapoli:
(and by the way the idea that we had was to forward the `AWS_PROFILE` variable to the docker container, but your solution will help people that don't want to mess with this variable so it's great too I think)

j0k3r:
You were right, defining the `AWS_PROFILE` env var in the docker image, solved the problem. The dashboard is working great now!

j0k3r:
> to what's passed to --profile (if defined)

AWS use the `default` profile by default, so I think we can always send the `$profile` (even when the default value is used)

mnapoli:
That's awesome thanks!

PR #428, opened 2019-08-23
Updated the README.md

andresayej:
to reflect the first known existence of all in one solution that simply works

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -213,7 +213,28 @@ This matrix will be updated as Bref and AWS services evolve over time.
 
     Migrating a legacy PHP application to Bref and Lambda can be a challenge. First, as explained above, the limitations that come with MySQL/PostgreSQL often apply. On top of that legacy applications tend to be extra slow and large which can make performances suffer.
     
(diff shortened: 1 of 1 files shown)


timoschaefer:
This reads more like an ad rather than outlining a fully-managed cloud solution - albeit a proprietary one that's tightly coupled to one framework - when we're about to get [documentation for even more frameworks](https://github.com/brefphp/bref/pull/419) which will get people to notice bref naturally, instead of being "hinted" in the direction of a competing, more limiting cloud solution.

Apart from requiring knowledge of either Cloudformation or Serverless configuration setups, bref is easy enough to setup, requiring only few lines of codes changed compared to "regular" deployments using e.g. Docker. Plus the current documentation is extensive and exhaustive enough to guide you through the migration every step of the way to get you started with PHP on Lambda within hours.

andresayej:
So a better place for this would be in docs/frameworks/laravel.md?

mnapoli:
We don't need to put that advertisement in the documentation. It just doesn't make sense to me (and it seems I'm not the only one).

> So a better place for this would be in docs/frameworks/laravel.md?

No 🤷‍♂

andresayej:
@mnapoli 👍 
The pull request is closed, so that settles it i guess.

PR #429, opened 2019-08-24
Add stage command option

tiagobrito:
Serverless application might have different stages, and it can be different from aws-profile name (eg: `stage=prod` and `aws-profile=production`).
When this happens, serverless info is throwing an error:
`[ERROR] The command serverless info failed`

Screenshot 2019-08-24 at 13 39 43

### DIFF
--- a/bref
+++ b/bref
@@ -250,7 +250,7 @@ $app->command('deployment stack-name [--region=]', function (string $stackName, 
     return 0;
 })->descriptions('Displays the latest deployment logs from CloudFormation. Only the logs from the last 24 hours are displayed. Use these logs to debug why a deployment failed.');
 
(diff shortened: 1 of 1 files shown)


bref
```diff
@@ -250,7 +250,7 @@ $app->command('deployment stack-name [--region=]', function (string $stackName,
     return 0;
 })->descriptions('Displays the latest deployment logs from CloudFormation. Only the logs from the last 24 hours are displayed. Use these logs to debug why a deployment failed.');
 
-$app->command('dashboard [--host=] [--port=] [--profile=]', function (string $host = 'localhost', int $port = 8000, string $profile = 'default', SymfonyStyle $io) {
+$app->command('dashboard [--host=] [--port=] [--profile=] [--stage=]', function (string $host = 'localhost', int $port = 8000, string $profile = 'default', string $stage = 'prod', SymfonyStyle $io) {
```
tiagobrito: probably the default option should be `null`

mnapoli:
Awesome thanks!

Issue #425, opened 2019-08-23
Add a warning when installing Bref to check that the version is at least 0.5.0


mnapoli:
More and more people are installing Bref either via old blog posts or using an old version of PHP.

They end up with an old version of Bref and the serverless plugin doesn't work.

We should add a warning here: https://bref.sh/docs/installation.html#bref to encourage users to check that the have installed Bref 0.5.*.

PR #430, opened 2019-08-25
Add a warning when installing Bref

linuxjuggler:
Add a note to encourage using Bref greater than or equal to 0.5.* and close https://github.com/brefphp/bref/issues/425

> More and more people are installing Bref either via old blog posts or using an old version of PHP.
>
> They end up with an old version of Bref and the serverless plugin doesn't work.
>
> We should add a warning here: https://bref.sh/docs/installation.html#bref to encourage users to check that they have installed Bref 0.5.*.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -46,7 +46,7 @@ Install Bref in your project using [Composer](https://getcomposer.org/):
 composer require bref/bref
 ```
 
-> Make sure that you have installed a version greater than or equal to 0.5.* of Bref, to make use of the serverless plugin. Older version of Bref won't work with the serverless plugin.
+> Make sure that you have installed a version greater than or equal to 0.5.* of Bref.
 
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -46,6 +46,8 @@ Install Bref in your project using [Composer](https://getcomposer.org/):
 composer require bref/bref
 ```
 
+> Make sure that you have installed a version greater than or equal to 0.5.* of Bref, to make use of the serverless plugin. Older version of Bref won't work with the serverless plugin.
```
mnapoli: I think mentioning the serveless plugin is too early in the documentation (readers will not know what it means).

We can probably keep it simpler, "Make sure that you have installed a version greater than or equal to 0.5.* of Bref" is perfect on its own!

mnapoli:
Thanks! I have added a little note inline

mnapoli:
Sorry I missed that you pushed an update (I didn't receive an email notification).

Merging!

PR #431, opened 2019-08-26
docs: update database access from php

miketheman:
After spending a bit of time trying to figure out why the application was reporting:
`could not find driver` - I figured it might help to point the reader in the right direction,


mnapoli:
Thanks!

PR #433, opened 2019-08-26
Disable workers output for PHP-FPM

### LINKED ISSUE #432: PHP-FPM is adding a prefix warning to stderr
According to https://github.com/docker-library/php/issues/207 this is a known issue, but in PHP 7.3 seems to have a solution according to https://www.php.net/ChangeLog-7.php#7.3.0 using `decorate_workers_output`.

### Motivation:
This prefix will make very difficult to process the log entry and correctly parse the log entry.

### Example:
```bash
php_1  | [26-Aug-2019 16:26:59] WARNING: [pool default] child 6 said into stderr: "[2019-08-26 16:26:59] app.INFO: Response 200 for "GET /ping" {"method":"GET","path":"","uri":"/ping","content-type":null,"latency":708,"client-ip":"172.22.0.1","status_code":200,"user-agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/76.0.3809.100 Safari/537.36"} []"
```

tiagobrito:
Resolves #432

### DIFF
--- a/runtime/layers/fpm-dev/Dockerfile
+++ b/runtime/layers/fpm-dev/Dockerfile
@@ -2,5 +2,10 @@ ARG LAYER_IMAGE
 FROM $LAYER_IMAGE
 
 COPY ./php-fpm.conf /opt/bref/etc/php-fpm.conf
+
+RUN if [ "$LAYER_IMAGE" = "php-73-fpm" ] ||  [ "$LAYER_IMAGE" = "php-73" ]; \
+    then echo "decorate_workers_output not supported"; \
+    else echo "decorate_workers_output = no" >> /opt/bref/etc/php-fpm.conf; fi
+
 EXPOSE 9000
 CMD /opt/bin/php-fpm --nodaemonize --fpm-config /opt/bref/etc/php-fpm.conf
--- a/runtime/layers/fpm/Dockerfile
+++ b/runtime/layers/fpm/Dockerfile
@@ -5,6 +5,10 @@ COPY bootstrap /opt/bootstrap
 COPY php.ini /opt/bref/etc/php/conf.d/bref.ini
 COPY php-fpm.conf /opt/bref/etc/php-fpm.conf
 
+RUN if [ "$LAYER_IMAGE" = "php-73-fpm" ] ||  [ "$LAYER_IMAGE" = "php-73" ]; \
+    then echo "decorate_workers_output not supported"; \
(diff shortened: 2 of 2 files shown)


mnapoli:
Thanks for the PR! This is something we have disabled for now, see line 20: https://github.com/brefphp/bref/pull/433/files#diff-c5bdb8ae764cff8b941bbc75d02d8aadR20

Unfortunately, it isn't supported in PHP 7.2. Unless we separate the config (might be good to do someday if it doesn't make the whole thing too complex to maintain) we have to keep the config compatible with 7.2.

tiagobrito:
@mnapoli sorry. I didn't notice. Thanks for the heads up!

Maybe we could attach this config when the layer is PHP 7.3 on the docker build:

```bash
RUN if [ "$LAYER_IMAGE" = "php-73-fpm" ] ;  then echo "decorate_workers_output not supported"  ; else "decorate_workers_output = no" >> /opt/bref/etc/php-fpm.conf; fi
```

tiagobrito:
@mnapoli did you had change to analyze this ☝️

mnapoli:
@tiagobrito that is a good idea. I am afraid it would make the whole system more complicated, but it might be worth it. And people seem to agree with you 😄 

So yes, let's try to do this! Would you have time to send a pull request for this?

tiagobrito:
I will do the PR. Thanks for your feedback

tiagobrito:
@mnapoli done

tiagobrito:
@mnapoli done. I've tested with my custom layer and it worked as expected. Let me know if I missed something.

runtime/layers/fpm-dev/Dockerfile
```diff
@@ -2,5 +2,10 @@ ARG LAYER_IMAGE
 FROM $LAYER_IMAGE
 
 COPY ./php-fpm.conf /opt/bref/etc/php-fpm.conf
+
+RUN if [ "$LAYER_IMAGE" = "php-73-fpm" ] ||  [ "$LAYER_IMAGE" = "php-73" ]; \
+    then echo "decorate_workers_output not supported"; \
+    else echo "decorate_workers_output = no" >> /opt/bref/etc/php-fpm.conf; fi
```
mnapoli: Should it be the other way around? If 72 then "not supported", else we set the flag.

tiagobrito: true. It seems that when it's running the build image $LAYER_IMAGE is not available or empty 😩

PR #440, opened 2019-08-29
Clarify text in the maturity matrix

mnapoli:
Following questions in Slack this should make the documentation more explicit.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -187,9 +187,9 @@ This matrix will be updated as Bref and AWS services evolve over time.
 
 - **API**
 
-    APIs run well on AWS Lambda thanks to the API Gateway integration.
+    APIs run on AWS Lambda without problems.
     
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -187,9 +187,9 @@ This matrix will be updated as Bref and AWS services evolve over time.
 
 - **API**
 
-    APIs run well on AWS Lambda thanks to the API Gateway integration.
+    APIs run on AWS Lambda without problems.
```
AlessandroMinoccheri: What do you think to specify that we have to use API Gateway because Lambda  can't expose directly the app to the internet via HTTP?

mnapoli: Quoting the discussion from Slack:

> The thing is that users of Bref don't "have to" use API Gateway. It just will be used, it will be transparent for them.
> Which is why I figured we could as well remove this information altogether.

PR #444, opened 2019-09-03
Fall back to parent functions for Symfony Kernel

kasperg:
Thanks for sharing Bref!

I just ran through the initial setup of a Symfony app and noticed  that Lambda needs a customized Symfony Kernel to be able to specify were to write logs and cache.

However when the application is not executed through Lambda we can fall back to the parent definitions instead of replicating the code. There is no need to replicate the code in these functions. I suggest that we update the documentation.


mnapoli:
Good call, that's definitely better!

Thanks for the pull request!

PR #446, opened 2019-09-05
Add layer to "Keep Alive"

Nyholm:
So, the other day I got an idea. What if we removed PHP FPM that must be way quicker, right? To only have a PHP script that keeps the Lambda alive and listens to new requests. Very much inspired by https://github.com/PHPFastCGI/FastCGIDaemon.

I started to do some tests and this is what I've come up with. This PR is working but incomplete. 

------------

I've got 3 scenarios:
- One baseline running **PHP-FPM** in a normal way. 
- One were I load my application every time a new request comes in. I call this **HTTP Function**. See example [handler](https://github.com/Nyholm/small-symfony-project/blob/master/public/lambda.php).
- One were I keep my application in memory between request. I call this **Keep Alive**. This require a new runtime layer (in this PR). See example [handler](https://github.com/Nyholm/small-symfony-project/blob/master/public/lambda_keep_alive.php). This will of course suffer from potential memory leaks. 

Running my [Small Symfony project controller](https://github.com/Nyholm/small-symfony-project/blob/master/src/Controller/HelloController.php) I get the following results. I was about to be super scientific with 10x10 repetitive runs on each scenario but it was quite easy to see a pattern after just a few runs.

  | First request | Average (10 reqs.) | Memory
-- | -- | -- | --
PHP-FPM | 1719 ms | 8 ms  | 133 mb
HTTP Function | 1890 ms | 70 ms | 133 mb
Keep alive | 1861 ms | 5 ms | 133 mb

So, no real difference in memory.
Keep Alive is faster on the average run. It will be even faster compared to PHP-FPM on a larger application. 

I cannot figure out why my first request is slower on Keep Alive. I was just it would be a little quicker since we don't have to boot PHP-FPM. 

---------

Is the Keep Alive layer interesting to keep working with?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
     },
     "require-dev": {
         "phpunit/phpunit": "^6.5",
-        "zendframework/zend-diactoros": "^1.6|^2.0",
+        "riverline/multipart-parser": "^1.2",
         "doctrine/coding-standard": "^5.0",
         "phpstan/phpstan": "^0.11.5",
         "guzzlehttp/guzzle": "^6.3",
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
     },
     "require-dev": {
         "phpunit/phpunit": "^6.5",
-        "riverline/multipart-parser": "^1.2",
+        "riverline/multipart-parser": "^2.0",
         "doctrine/coding-standard": "^5.0",
         "phpstan/phpstan": "^0.11.5",
         "guzzlehttp/guzzle": "^6.3",
--- a/composer.json
+++ b/composer.json
@@ -37,6 +37,7 @@
         "zendframework/zend-diactoros": "^1.6|^2.0",
         "doctrine/coding-standard": "^5.0",
         "phpstan/phpstan": "^0.11.5",
-        "guzzlehttp/guzzle": "^6.3"
+        "guzzlehttp/guzzle": "^6.3",
+        "symfony/http-foundation": "^4.3"
     }
 }
--- a/runtime/layers/keep-alive/bootstrap
+++ b/runtime/layers/keep-alive/bootstrap
@@ -1,0 +1,43 @@
+#!/opt/bin/php
+<?php declare(strict_types=1);
+
+use Bref\Http\LambdaRequest;
+use Bref\Http\LambdaResponse;
+use Bref\Runtime\LambdaRuntime;
+
+$appRoot = getenv('LAMBDA_TASK_ROOT');
+$callableFile = $appRoot . '/' . getenv('_HANDLER');
+if (! is_file($callableFile)) {
+    $lambdaRuntime->failInitialization("Handler `$handler` doesn't exist");
+}
+
+$callable = include $callableFile;
+if (!is_callable($callable)) {
+    throw new LogicException('The handler must return a callable when using keep-alive.');
+}
+
+if (!class_exists(LambdaRuntime::class)) {
+    throw new LogicException('You must load composer\'s autoloader.');
+}
+
+$lambdaRuntime = LambdaRuntime::fromEnvironmentVariable();
+$memoryMax = getenv('BREF_MEMORY_MAX') ?: 1000000000;
+$loopMax = getenv('BREF_LOOP_MAX') ?: 50;
+$loops = 0;
+while (true) {
+    if (++$loops > $loopMax || memory_get_usage() > $memoryMax) {
+        // Quit.
+        return;
+    }
+
+    $lambdaRuntime->processNextEvent(function ($event) use ($callable): array {
+        $response = $callable(LambdaRequest::create($event));
+
+        if (!$response instanceof LambdaResponse) {
+            throw new LogicException('Wrong response type');
+        }
+
+        $multiHeader = array_key_exists('multiValueHeaders', $event);
+        return $response->toApiGatewayFormat($multiHeader);
+    });
+}
--- a/runtime/layers/keep-alive/bootstrap
+++ b/runtime/layers/keep-alive/bootstrap
@@ -30,8 +30,8 @@ while (true) {
         return;
     }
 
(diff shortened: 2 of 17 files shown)


skyrpex:
Maybe I'm wrong but doesn't the Keep Alive thing go against the whole Lambda idea, where you are just charged for what you use? Also, lambda instantiations have a life span of about 29s, which would render Keep Alive unstable since late requests may get cancelled.

Nyholm:
Excellent question.  But you are wrong. 

Like the php-fpm we have a process running that are ready to take incomming requests. We are not executing anything for more than 29s. 
Unlike php-fpm we can share data between requests. Ie your autoloaded files, your warm symfony container.... but even state. 

This should not be a replacement for fpm but rather something some developers might do to speed things up.

runtime/layers/keep-alive/bootstrap
```diff
@@ -0,0 +1,43 @@
...
+}
+
+$lambdaRuntime = LambdaRuntime::fromEnvironmentVariable();
+$memoryMax = getenv('BREF_MEMORY_MAX') ?: 1000000000;
+$loopMax = getenv('BREF_LOOP_MAX') ?: 50;
+$loops = 0;
+while (true) {
+    if (++$loops > $loopMax || memory_get_usage() > $memoryMax) {
+        // Quit.
+        return;
```
Nyholm: This will be an "early exit"

skyrpex:
Interesting, thanks for the explanation 😊will keep an eye on this!

Nyholm:
@AndrewCarterUK explains this in a conference talk. Watch from 2 min and about 5 min forward. (The audio is horrible at first but gets better) https://youtu.be/jNdzwRdEEYQ

Later in the video he explains possible issues with this. Keep in mind that most frameworks manage their memory better since this talk was recorded.

mnapoli:
Note that in Bref 0.2 everything was converted to PSR-7, and then we used the Symfony bridge to create Symfony requests. I wonder if we should do something similar or have a special case for Symfony.

What I mean is that here is your handler:

```php

return function(LambdaRequest $request) use ($kernel) {
    $sfRequest = $request->getSymfonyRequest();
    $kernel->reboot($kernel->getCacheDir());
    $response = $kernel->handle($sfRequest);
    try {
        return LambdaResponse::fromSymfonyResponse($response);
    } finally {
        $kernel->terminate($sfRequest, $response);
    }
};
```

What we could do instead is provide directly the HTTP request, and expect the HTTP response back:

```php

return function(Request $request) use ($kernel) {
    $kernel->reboot($kernel->getCacheDir());
    $response = $kernel->handle($request);
    try {
        return $response;
    } finally {
        $kernel->terminate($request, $response);
    }
};
```

But it would make sense to use PSR-7 instead:

```php
return function(ServerRequestInterface $request): ResponseInterface {
    ...
};
```

We could provide a Symfony adapter [like in Bref 0.2](https://github.com/brefphp/bref/blob/0.2.37/src/Bridge/Symfony/SymfonyAdapter.php), e.g.

```php
return symfonyAdapter(function(Request $request): Response {
    ...
});
```

Nyholm:
> So this is a great addition to Bref. I still think the PHP-FPM runtime should be presented as the default solution for HTTP applications. But the "Keep Alive" runtime could be presented as a more advanced and faster alternative.

Absolutely agree

I will look at your references and continue with this PR. 

> Symfony Bridge. 

I like that package, I've reviewed and written big parts of it. I must say it is really really good. But not perfect. In some situations you may get different requests when doing PSR7 -> Symfony -> PSR7.

There is no reason to create more objects than needed, so I prefer no to use it. Not using it will allow us to not have a PSR7 implementation in our code.

Nyholm:
I also like the fact that we are using a Lambda request. It is a class that is very easy to create an object from. And if the consumer does not care abut PSR7 or Symfony requests then there is no harm done. If we are trying to be smart, there will just be more code and more time is spent on something that may be useless.

composer.json
```diff
@@ -34,9 +34,10 @@
     },
     "require-dev": {
         "phpunit/phpunit": "^6.5",
-        "zendframework/zend-diactoros": "^1.6|^2.0",
```
Nyholm: There is no good reason to use `zendframework/zend-diactoros`.

Best option is of course nyholm/psr7. But we already have guzzlehttp/psr7 installed.

Nyholm:
I think Im done with this PR now. Tests fails because of https://github.com/Riverline/multipart-parser/pull/32.

I will open docs in a second PR. 

I added a `@experimental` annotation on `LambdaRequest::getPsr7Request()` and `LambdaRequest::getSymfonyRequest()` to mark that it is super risky to use these in production.

tests/Server.php
```diff
@@ -188,6 +191,7 @@ private static function getClient()
                 'sync'     => true,
             ]);
         }
+
```
Nyholm: These unrelated CS changes were made by `pretty fix`.

Nyholm:
Is this PR under consideration? Should I close it or keep updating it?

mnapoli:
Yes, I think it's interesting! I don't really know how the preloading would affect the future of that though, I don't have any idea about the difference in performance.

My plan would be bigger actually:

- here we provide a typed event object (the request, either PSR-7 or Symfony) to a handler -> I want users to be able to receive typed event objects for all events (see https://github.com/brefphp/bref/compare/class-handlers for an example with SQS)
- here we have a specific "loop" runtime -> I want to merge it with the current runtime, and by default the loop will have 1 execution (no loop), and users can then turn it into a "keep alive" by defining the number of loops to 10000 or even `0` (infinite)

That would make things both simpler, yet more powerful. Also easier to maintain.

Unfortunately, I lacked time lately to work more on this. I want to be able to spend more time at the beginning of 2020 to make this happen.

mnapoli:
Here is the followup for my comment above: #521

Nyholm:
This was fixed by #521.

PR #448, opened 2019-09-09
Update matrix build

### LINKED ISSUE #447: Update the maturity matrix following VPC cold starts improvements
VPC cold starts are being removed, we should update the maturity matrix to reflect that.

EDkSxB3XUAg3Czb

We should also update the text below the matrix.

tristanbes:
Fixes #447


mnapoli:
Thanks for the PR! And :+1: the idea with #447 was also to update the text below the table, else it will be outdated.

tristanbes:
Supressed by #449

PR #449, opened 2019-09-10
fix doc about cold starts

### LINKED ISSUE #447: Update the maturity matrix following VPC cold starts improvements

tristanbes:
Fixes #447


mnapoli:
Thanks!

PR #462, opened 2019-09-20
Fix #461 Build the fpm-dev images from the lambci images

### LINKED ISSUE #461: bref/php-73-fpm-dev fails to start
Docker image bref/php-73-fpm-dev (DIGEST: 92f5601d99eb) fails to start

```
/opt/bin/php-fpm: error while loading shared libraries: libpng12.so.0: cannot open shared object file: No such file or directory
exited with code 127
```

mnapoli:
It seems the AmazonLinux images do not have exactly the same libraries as the actual Lambda instances.

We build all docker images from `lambci/lambda` (which is supposed to contain everything) but we didn't build `fpm-dev` from that image as well.

This fixes the inconsistency, which should fix #461.

### DIFF
--- a/runtime/layers/fpm-dev/Dockerfile
+++ b/runtime/layers/fpm-dev/Dockerfile
@@ -9,4 +9,6 @@ FROM lambci/lambda:provided
 COPY --from=0  /opt /opt
 
 EXPOSE 9000
+# Clear the parent entrypoint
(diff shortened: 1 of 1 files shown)


mnapoli:
Tested and working for me. I have already published the new Docker image so that you can test.

PR #466, opened 2019-09-24
Allow region to be a variable

### LINKED ISSUE #372: Bref Serverless plugin does not resolve variables on time
I have configured my `region` as a variable:

```
    region: ${env:AWS_LAMBDA_REGION}
```

Bref tries to read the region before it is resolved:

```
'There is no Bref layer named "php-73-fpm" in region "${env:AWS_LAMBDA_REGION}"'
```

(the env variable is coming from a dotenv file, using the dotenv plugin: https://github.com/colynb/serverless-dotenv-plugin)

harmenjanssen:
Before, the region would be read from incoming options, sometimes
resulting in an unresolved variable.

This change uses the `getRegion()` method of the AWS provider to make
sure the region is resolved, before using it in layer resolution.

Fixes #372

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,16 +13,12 @@ class ServerlessPlugin {
         const filename = path.resolve(__dirname, 'layers.json');
         const layers = JSON.parse(fs.readFileSync(filename));
 
-        // Read the region from the `--region` option, or fallback on the `serverless.yml` config
-        const region = (typeof options.region !== 'undefined')
-            ? options.region
-            : serverless.service.provider.region;
-
         // Override the variable resolver to declare our own variables
         const delegate = serverless.variables
             .getValueFromSource.bind(serverless.variables);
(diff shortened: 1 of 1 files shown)


mnapoli:
Thanks! Does it still work if you pass a different region via the CLI option:

```
serverless deploy --region=eu-west-1
```

harmenjanssen:
It depends on the configuration, I think. Whatever is being used to deploy the function is also being returned from the `getRegion` function, whether configured as option or in `serverless.yml`. 

Interestingly, in my case I'm using `region: ${env:AWS_LAMBDA_REGION}`, but when passing `region` as an option like so:

```
serverless deploy --region foobar
```

It will complain `''There is no Bref layer named "php-73-fpm" in region "foobar"'` so it's definitely being picked up.

mnapoli:
Awesome thanks!

PR #467, opened 2019-09-27
Fix php-7*-fpm-dev images that were missing php.ini

mnapoli:
Reported by @jenschude in Slack. The `php.ini` was missing in the PHP-FPM dev image.

I have fixed this by basing the PHP-FPM dev image on the `php-73-fpm` images.


runtime/layers/fpm-dev/php-fpm.conf
```diff
@@ -1,5 +1,5 @@
 ; Logging anywhere on disk doesn't make sense on lambda since instances are ephemeral
-error_log = /dev/stderr
+error_log = /dev/null
```
mnapoli: I set the same value as in the production version. Logs are collected via `--force-stderr`, just like on Lambda (one less difference).

PR #469, opened 2019-10-01
add xdebug to fpm-dev docker container

jenschude:
For better development experience using the fpm docker container I added xdebug to the docker container.

As the container images are build using an intermediary container I added the install step here and delete the xdebug.so from the resulting container if necessary.

As xdebug uses by default the port 9000 i changed the fpm port in the dev container to 9001 one to reduce the configuration overhead for the users.

### DIFF
--- a/docs/local-development.md
+++ b/docs/local-development.md
@@ -128,6 +128,35 @@ In the example above, a `public/assets/style.css` file will be accessible at `ht
 
 > Be aware that serving assets in production will not work like this out of the box. You will need [to use a S3 bucket](/docs/runtimes/http.md#assets).
 
+### Xdebug
+
+The docker container `bref/php-<version>-fpm-dev` comes with xdebug pre-installed. In order to enable it you can create a
+folder `php/conf.dev.d` in your project and include an ini file enabling xdebug:
+
+```ini
+zend_extension=xdebug.so
+```
+
+Now start the debug session by issueing a request to your application in the [browser](https://xdebug.org/docs/remote#starting).
+
+#### OSX
+
+As Docker on Mac uses a virtual machine for running docker it can be tricky to listen to the xdebug port.
+By creating an alias for the loopback interface on the host with
+
+```bash
+sudo ifconfig lo0 alias 10.254.254.254
+```
+
+you can use this configuration start listening to xdebug connections in your preferred IDE:
+
+```ini
+[xdebug]
+xdebug.remote_enable = 1
+xdebug.remote_autostart = 0
+xdebug.remote_host = '10.254.254.254'
+```
+
 ## Console applications
 
 Console applications can be tested just like before: by running the command in your terminal.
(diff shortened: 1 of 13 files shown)


marcusirgens:
So we discussed this internally and we have two caveats with this
1. This extension should be activated manually like the other manually enabled bref extensions
2. We would like to see a format for the custom ini files that are ignored in production, eg. *.dev.ini, that are used only in the dev container (eg. xdebug.dev.ini), so we reduce the risk of committing invalid configs to our prod runtimes. 

If you agree with this we would be happy to contribute to the PR to build this functionality into.

mnapoli:
I'm really happy with adding xdebug. Enabled or not by default is a good question. If it is disabled by default, maybe it should be "enableable" not via `conf.d/php/php.ini` to avoid mistakes.

Since the dev container is not a lambda layer, it can behave differently. For example xdebug could maybe be enabled via an environment variable? Or a different container image or image version?

Also I wish we could build the extension separately and not edit the main compilation script (that would avoid having to remove the xdebug extension from the prod images). But that will require a little bit more effort and refactoring of the current images… Maybe we could have a separate Dockerfile whose single job is building the `xdebug.so` file? (even if it means reinstalling stuff in that Dockerfile)

jenschude:
As all other extensions xdebug is not enabled by default and has to be loaded using the php.ini in your project.

But I will take a look in the issue to be able to define a .dev.ini which is only loaded in the dev container

marcusirgens:
I realised my comment came across as if I’m a maintainer, which was not my intention. Our company uses bref and we’re interested in contributing, the feedback is from our devs.

jenschude:
I changed the build so that it's now possible to define extensions to be installed in the build of the intermediate image. As all the intermediate steps are cached the whole build doesn't take that much additional time.

Will take a look later into how to add a php.dev.ini

marcusirgens:
Might be simpler to use a separate directory for the dev php.ini files?

ons. 16. okt. 2019 kl. 13:13 skrev Jens Schulze <notifications@github.com>:

jenschude:
Other option could be to use the `PHP_INI_SCAN_DIR` environment variable to add an additional scan path. See https://www.php.net/manual/en/configuration.file.php#configuration.file.scan

jenschude:
I configured the dev container to set the PHP_INI_SCAN_DIR environment variable to also scan `/var/task/php/conf.dev.d` for ini files

Also as I had in my tests sometimes issue with loading extensions I added a make target with a smoke test which tests that all extensions are loadable.

mnapoli:
I am trying to catch up but this PR turned out quite complex now.

With #490 I think we will be able to use multi-stage Docker builds to simplify this a bit.

Here is an illustration of what I have in mind:

```dockerfile
# fpm-dev/Dockerfile
FROM bref/build-php-73

RUN pecl install xdebug

FROM bref/php-73-fpm

# Copy xdebug.so that we installed above
COPY --from=0  /opt/lib/php/extensions/..../xdebug.so /opt/lib/php/extensions/..../xdebug.so

# ... <rest of the Dockerfile>
```

The `php/conf.dev.d/*.ini` loading is clever, that will be useful here and for blackfire (#488).

If you could move the smoke tests to a separate PR that would be great. I'd like to keep this PR as simple as we can, that will be much easier to merge.

jenschude:
See #492 

I would suggest to first refactor the build, than I will create a new PR for Xdebug. Same for the blackfire extension.

jenschude:
I created a new clean PR with adjustments to the new container build scripts

PR #471, opened 2019-10-02
Add an example artisan command

byjujohn:
For beginners it is difficult to figure out the function name from the configuration options.


mnapoli:
👍 thanks!

PR #472, opened 2019-10-02
Update notice about cold starts

barryvdh:
See https://aws.amazon.com/blogs/compute/announcing-improved-vpc-networking-for-aws-lambda-functions/


mnapoli:
Good call, I completely missed this. Thanks!

Issue #309, opened 2019-04-17
/Prod/ path prefix ignored by frameworks (Laravel/Symfony)


headyj:
Hello,

I basically followed the tutorial about laravel integration: https://bref.sh/docs/frameworks/laravel.html

I'm stuck now because the redirection removes the "/Prod" or "/Stage" part, leading me to a 403 error. 

For example, on the first call, the application should redirect me to the login page:
https://<api_id>.execute-api.eu-west-1.amazonaws.com/Prod/login

but instead it removes the "/Prod" part:
https://<api_id>.execute-api.eu-west-1.amazonaws.com/login

I'm not quite sure how to solve this issue. I tried to compare my configuration with the demo repository (https://github.com/mnapoli/bref-laravel-demo) but I could not found how to solve it.

Thanks for your help

mnapoli:
Thank you. I think this is related to this:

https://github.com/mnapoli/bref/blob/aa941f89cd66cd1643cf9bbc75525f58dc942736/src/Runtime/PhpFpm.php#L215

and this is a problem we've discussed in the past. The root of the problem is that the `$event['path']` key does not contain the stage prefix (`/Prod/`).

Pro:

- we can deploy an application to API Gateway and all the routes will work without needing to add the `/Prod/` prefix in them

Con:

- since the app is unaware of the prefix links and redirects can be broken

If we add the prefix back, then deploying an application to API Gateway will break because the prefix will break the routes -> this is really annoying.

Since most applications are deployed with a domain name this is even more annoying to dev locally without a prefix, test for a couple minutes (after the first deployment) with the path prefix (which means all route have to be changed) and then redeploy with the domain name (all routes are broken again).

I wish we could find a all-in-one solution to this to have routes and redirects work both with and without the prefix… Any idea is welcome!

mnapoli:
Related: #67

atrope:
@mnapoli I Was trying to think how we could do that but, how can you deploy an application to APIGW without the need to adding the prefix(stage) to it? If I Deploy to a stage, and try to enter without the stage in the URL, APIGW returns me Forbidden and don't even ping the Layer (I've just tested now).

The best thing i can think of is removing the stage from the path before sending to the application.
```

        $path = $event['path'] ?? '/';
        $path = explode("/",$path);
        array_shift($path); // Remove stage
        $path = "/" . implode('/', $path);

```

That way, the application will be Stage agnostic, but links must start with the stage.

[...]

PR #474, opened 2019-10-15
Add description about the stage prefix for symfony

jenschude:
It can be quite tricky to get correctly handle the stage prefix in Symfony.

### DIFF
--- a/docs/frameworks/symfony.md
+++ b/docs/frameworks/symfony.md
@@ -116,3 +116,30 @@ Globals:
 The secrets (e.g. database passwords) must however not be committed in this file.
 
 To learn more about all this, read the [environment variables documentation](/docs/environment/variables.md).
+
+## The /dev/ prefix and Symfony routes
+
+If you didn't setup a custom domain for your lambda function will have the API gateways stage prefix.
+As symfony by default creates absolute URLs for your routes you have to configure your webserver set the script name correctly.
+
+As we don't have a server configuration you can set the environment variables also in your `public/index.php` by adding this snippet before instantiating the Kernel
+
+```php
+if (isset($_SERVER['ROUTE_BASE_URL'])) {
+    $baseUrl = $_SERVER['ROUTE_BASE_URL'];
+    $selfFile = basename(__FILE__);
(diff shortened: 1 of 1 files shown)


Nyholm:
Thank you for this PR. 

I see that the `ROUTE_BASE_URL` variable is the switch to turn this on/off. Could we do it in a bit more automatic way? Ie, not use custom environment variables

jenschude:
Yeah tried it:

https://github.com/brefphp/bref/compare/master...jenschude:set_script_name_env

Issue is still, that it's not reliable possible to get the information from where the request originates and what the real calling URI was.

Nyholm:
If #542 gets merged, would that help? I think it would, right?

jenschude:
Depends. The bref.dev service for example acts like a normal client and initially didn‘t set a Forwarded Header. Which made it impossible to read the called URI

docs/frameworks/symfony.md
```diff
@@ -116,3 +116,30 @@ Globals:
...
+}
+```
+
+This assumes that your function has a environment variable named `ROUTE_BASE_URL` which will hold the stage prefix like `/dev`.
+To automatically set this variable add this snippet to your serverless.yaml
+
+```yaml
+provider:
+  environment:
+    ROUTE_BASE_URL: '/${self:custom.stage}'
```
xavismeh, line 142: Using `/${self:provider.stage, 'dev'}` will allow to do not determine an extra custom parameter reusing an already defined one

Nemo64:
Since the http api is now generally available, wouldn't it make more sense to slowly change everything to the http api which does not need a prefix?
There are few cases where you need the rest api and if you need them you are probably very deep into it.

jenschude:
When the pull request was initially opened the default for serverless was to have stage prefixes. This can be circumvented now. But even if deployed with prefixes one has to know some internals of Symfony how to correctly setup the environment so that route prefixes are correctly identified especially as Symfony defaults to generate absolute URLs.

mnapoli:
Since #309 is fixed in #794 and we switch to API Gateway v2, this problem will disappear. Thanks for the contribution anyway!

PR #479, opened 2019-10-21
Set script_name environment variable

jenschude:
The `/dev` prefix is not identified by the Symfony HttpFoundation cause the environment variables are not correctly setup to retrieve the base URI.

By correctly setting the script name and request uri the dev prefix is correctly identified as a base URI.

See also https://github.com/symfony/symfony/blob/16d528504cd57c4f967ea095e8c603260e119125/src/Symfony/Component/HttpFoundation/Tests/RequestTest.php#L594-L611

I also added some tests for a full blown API gateway event

### DIFF
--- a/src/Runtime/PhpFpm.php
+++ b/src/Runtime/PhpFpm.php
@@ -181,9 +181,15 @@ final class PhpFpm
         $request = new FastCgiRequest($method, $this->handler, $requestBody);
 
         $queryString = $this->getQueryString($event);
-        $uri = $event['path'] ?? '/';
+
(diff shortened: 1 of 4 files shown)


jenschude:
Tried it with the new *.bref.dev service. The direct lambda URL works. With *.bref.dev not cause the event doesn't hold any useful information about the proxying.

So I will still have to figure out another way of solving this issue.

mnapoli:
> Tried it with the new *.bref.dev service. The direct lambda URL works. With *.bref.dev not cause the event doesn't hold any useful information about the proxying.

@jenschude what information would you need?

PR #488, opened 2019-10-25
Add blackfire probe extension to fpm-dev container

jenschude:
To find bottlenecks in the current implementation it's useful to have a profiler available. So by adding the probe one is able to use blackfire in the local development flow.


runtime/layers/fpm-dev/Dockerfile
```diff
@@ -4,6 +4,20 @@ FROM $LAYER_IMAGE
 # Override the config so that PHP-FPM listens on port 9000
 COPY php-fpm.conf /opt/bref/etc/php-fpm.conf
 
+USER root
+ENV current_os=linux
```
mnapoli: Is there a reason we need to change those 2 settings?

runtime/layers/fpm-dev/Dockerfile
```diff
@@ -4,6 +4,20 @@ FROM $LAYER_IMAGE
...
+RUN version=$(php -r "echo PHP_MAJOR_VERSION.PHP_MINOR_VERSION;")-zts \
+    && curl -A "Docker" -o /tmp/blackfire-probe.tar.gz -D - -L -s https://blackfire.io/api/v1/releases/probe/php/$current_os/amd64/$version \
+    && mkdir -p /tmp/blackfire \
+    && tar zxpf /tmp/blackfire-probe.tar.gz -C /tmp/blackfire \
+    && cp /tmp/blackfire/blackfire-*.so $(php -r "echo ini_get('extension_dir');")/blackfire.so \
+    && rm -rf /tmp/blackfire /tmp/blackfire-probe.tar.gz
+
+# Build the final image from the lambci image that is close to the production environment
+FROM lambci/lambda:provided
```
mnapoli: Is there a reason for this addition? I fear this will break features in the dev image.

docs/local-development.md
```diff
@@ -128,6 +128,28 @@ In the example above, a `public/assets/style.css` file will be accessible at `ht
...
+      BLACKFIRE_SERVER_ID: server-id
+      BLACKFIRE_SERVER_TOKEN: server-token
+```
+
+Now you can enable the blackfire probe extension in your projects php.ini
+
+```ini
+extension=blackfire
+blackfire.agent_socket=tcp://blackfire:8707
+```
```
mnapoli: If I understand correctly this must go in `php/conf.d/php.ini` right?

So that means this will break when deployed in production?

runtime/layers/fpm-dev/Dockerfile
mnapoli: I would rather remove this if this isn't necessary (keep things as simple as possible).

runtime/layers/fpm-dev/Dockerfile
mnapoli: OK one advantage is that it provides a final image with only 1 layer (lighter to download).

Could you change the comment on the line above to:

```
# Build the final image from scratch so that it is composed only from 1 layer (smaller image)
```

runtime/layers/fpm-dev/Dockerfile
jenschude: The root user was necessary cause the lambci image which was used for the fpm image uses a different user and you can‘t write to /opt when not root

jenschude: And the current os variable is only used for the download url of blackfire

mnapoli: OK I see now! I think I have an idea with multi-stage builds. I'll try that after #490

docs/local-development.md
jenschude: Yes it would go. But with the PHP_INI_SCAN_DIR environment variable this would not be needed in the php.ini directly

jenschude: With latest changes it's possible to use a different ini file for development extensions

mnapoli:
This is perfect!

Thanks!

PR #490, opened 2019-10-27
Refactor the build process

mnapoli:
This is an attempt at refactoring the build process (building Docker images and layers) to:

- make it simpler and more explicit (even though clearly it's not simple)
- allow custom instructions per PHP version, needed to support PHP 7.4 (see #415)
- publish new Docker images letting anyone compile their own PHP extensions (should help for #469)

---

Multiple Docker images are created:

- `bref/tmp/...` images are created locally and not published online
- `bref/...` images are published on Docker Hub

Workflow:

- 1: Create the `bref/tmp/step-1/build-environment` Docker image.
    This image contains everything needed to compile PHP. This image is created standalone because
    it is common for each PHP version. The next step involves a different Dockerfile per PHP version.

- 2: Create the `bref/build-php-XX` images.
    There is one image per PHP version. These images contain PHP compiled with all its extensions.
    It is published so that anyone can use it to compile their own extensions.
    It is not the final image because it contains too many things that need to be removed.

- 3: Create the `bref/tmp/cleaned-build-php-XX` images.
    There is one image per PHP version. These images contain PHP compiled with all its extensions,
    but with all extra files removed. These images do not contain the bootstrap files and PHP config files.

- 4: Create the `bref/php-XX`, `bref/php-XX-fpm`, `bref/php-XX-fpm-dev` images.
    There is one image per PHP version. These images contain exactly what will be in the layers in `/opt`.

- 5: Create the layers zip files in the `export/` directory.
    We zip the `/opt` directory of each Docker image.

- 6: Publish the layers on AWS.
    We upload and publish the layers in every AWS region using the zip files.


runtime/base/base.Dockerfile
```diff
@@ -1,9 +1,52 @@
...
+# Needed to build:
+# - libzip: minimum required CMAKE version 3.0.2
+RUN set -xe \
+ && mkdir -p /tmp/cmake \
+ && cd /tmp/cmake \
+ && curl -Ls  https://github.com/Kitware/CMake/releases/download/v3.13.2/cmake-3.13.2.tar.gz \
+    | tar xzC /tmp/cmake --strip-components=1 \
+ && ./bootstrap --prefix=/usr/local \
+ && make -j $(nproc) \
+ && make install
```
mnapoli: The former `compiler.Dockerfile` has been merged in there. There was no reason for a separate Dockerfile.

runtime/base/base.Dockerfile
```diff
@@ -49,11 +92,11 @@ RUN mkdir -p ${BUILD_DIR}  \
 # https://github.com/madler/zlib/releases
 # Needed for:
 #   - openssl
+#   - curl
 #   - php
 # Used By:
 #   - xml2
-ARG zlib
-ENV VERSION_ZLIB=${zlib}
+ENV VERSION_ZLIB=1.2.11
```
mnapoli: I inlined all version numbers. The `versions.ini` + script (now removed) made things harder to understand for me.

runtime/base/clean.sh
```diff
@@ -0,0 +1,37 @@
...
+# Treat unset variables and parameters as an error.
+set -u
+
+# Strip all the unneeded symbols from shared libraries to reduce size.
+find /opt/bref -type f -name "*.so*" -o -name "*.a"  -exec strip --strip-unneeded {} \;
+find /opt/bref -type f -executable -exec sh -c "file -i '{}' | grep -q 'x-executable; charset=binary'" \; -print|xargs strip --strip-all
+
+# Cleanup all the binaries we don't want.
+find /opt/bref/bin -mindepth 1 -maxdepth 1 ! -name "php" ! -name "php-fpm" -exec rm {} \+
+find /opt/bin -mindepth 1 -maxdepth 1 ! -name "php" ! -name "php-fpm" -exec rm {} \+
```
mnapoli: Slight change here: `pecl` is now removed for layers (wasn't working anyway since PEAR was completely removed).

runtime/layers/fpm-dev/Dockerfile
```diff
@@ -1,5 +1,5 @@
-ARG LAYER_IMAGE
-FROM $LAYER_IMAGE
+ARG PHP_VERSION
+FROM bref/php-$PHP_VERSION-fpm
```
mnapoli: Those changes are mainly to help understand which image is actually used.

This is a minor change.

runtime/layers/tests.php
```diff
@@ -0,0 +1,57 @@
+<?php /** @noinspection ALL */
+declare(strict_types=1);
+
+/**
+ * This file runs tests on Docker images.
+ */
```
mnapoli: I added tests for Docker images. That helps check that some things are correctly configured.

docs/environment/php.md
```diff
@@ -128,10 +128,25 @@ extension=pthreads
...
+```dockerfile
+FROM bref/build-php-73
+
+# Install system dependencies you might need
+RUN LD_LIBRARY_PATH= yum install -y ImageMagick-devel
+
+RUN pecl install imagick
+```
+
+The `.so` extension file can then be retrieved in `/opt/bref/lib/php/extensions/...`. If you installed system libraries, you may also need to retrieve them.
```
mnapoli: I have slightly improved the guide on how to compile custom extensions. This is much easier now thanks to the Docker images meant for that.

PR #492, opened 2019-10-28
add smoke test for extensions

jenschude:
Add smoke test to check if PHP extensions are loadable.


jenschude:
Close this one in favor of #499

PR #494, opened 2019-10-30
Add strpos to contentType for response

### LINKED ISSUE #482: Bref version 0.5.7 encoding the HTTP response (JSON) to Base64 string.
Hey @mnapoli , I use Bref to develop HTTP APIs for Lambda. 

**I had recently come across an issue where the response comes as a Base64 encoded string of the actual JSON response.** I reverted back the bref version to 0.5.6 and I was getting plain JSON response. I havent changed anything in my PHP or API gateway configuration. Just the bref version from 0.5.7 to 0.5.6 and It was giving me plain JSON response.

I have come to report this and ask if the responses of version 0.5.7 are expected to be Base64 encoded strings or is this a Bug?

P.S. Your work made my life better. Otherwise I would have had to write a handler and other stuff for PHP APIs myself which is not an easy task. Thanks for everything.

jasonmccallister:
When a response returns `application/json; charset=UTF-8` as the content type, Bref treats this a base64 encoded content.

Resolves #482

### DIFF
--- a/src/Runtime/LambdaRuntime.php
+++ b/src/Runtime/LambdaRuntime.php
@@ -251,7 +251,7 @@ final class LambdaRuntime
          * - else we encode all other responses to base64
          * API Gateway checks `isBase64Encoded` and decodes what we returned if necessary.
          */
-        if ($contentType && $contentType !== 'application/json' && strpos($contentType, 'text/') !== 0) {
+        if ($contentType && strpos($contentType, 'application/json') !== 0 && strpos($contentType, 'text/') !== 0) {
             $data['body'] = base64_encode($data['body']);
             $data['isBase64Encoded'] = true;
         }


jasonmccallister:
@mnapoli there is not specific test for this, I would feel more comfortable adding a test but I believe this would resolve the issue.

mnapoli:
Thanks, I do have plans to continue fixing the underlying issue but this will definitely help get things working for 99% of the use cases.

Nyholm:
Oh. I spent a good few hours with the same issue.. 

Im not sure how to you are planning to "fix the underlying issue". But I can tell you that according to json-api specifications, one should have content type like: 

```
Content-Type:  application/vnd.api+json
```

Just giving some input. Thank you for the PR and thank you for merging.

PR #495, opened 2019-10-30
Bump to postgres 10.7

Nyholm:
The Aurora serverless requires 10.7.

Im not sure what kind of effects this got.. Can you still be using AWS postgres database running 9.6.11?

### DIFF
--- a/runtime/base/base.Dockerfile
+++ b/runtime/base/base.Dockerfile
@@ -356,7 +356,7 @@ RUN set -xe; \
 #   - OpenSSL
 # Needed by:
 #   - php
-ENV VERSION_POSTGRES=9.6.11
+ENV VERSION_POSTGRES=.10.7
 ENV POSTGRES_BUILD_DIR=${BUILD_DIR}/postgres
 
 RUN set -xe; \


runtime/base/base.Dockerfile
```diff
@@ -356,7 +356,7 @@ RUN set -xe; \
 #   - OpenSSL
 # Needed by:
 #   - php
-ENV VERSION_POSTGRES=9.6.11
+ENV VERSION_POSTGRES=.10.7
```
mnapoli: Is `.10.7` correct? It seems like a weird version number.

Nyholm: Yeah. It is correct. It works with the string substitution when creating the url (they changed names of the releases)

Nyholm:
Hm. Maybe I’m wrong. Sorry for being lazy. 

I see that we download and compile Postgres/Postgres and we do that to build the php extension. 
Does this extension work with any version of Postgres server?

I assumed (did not test) that if we build the extension out of Postgres version 9.6 it would have problem connecting to a server of version 10.7.

mnapoli:
I would expect the PHP extension to work with all (most?) server versions, just like with MySQL.

Let's close this for now and reopen if there is a problem.

Nyholm:
Yeah. Let’s!

PR #498, opened 2019-11-06
Add xdebug to php-fpm-dev container

jenschude:
Replace for #469 

For better development experience using the fpm docker container I added xdebug to the docker container.

As the container images are build using an intermediary container I added the install step here and delete the xdebug.so from the resulting container if necessary.

As xdebug uses by default the port 9000 i changed the fpm port in the dev container to 9001 one to reduce the configuration overhead for the users.

### DIFF
--- a/runtime/layers/fpm-dev/Dockerfile
+++ b/runtime/layers/fpm-dev/Dockerfile
@@ -1,13 +1,27 @@
 ARG PHP_VERSION
-FROM bref/php-$PHP_VERSION-fpm
+FROM bref/build-php-$PHP_VERSION as build_extensions
 
+RUN pecl install xdebug
+RUN cp $(php -r "echo ini_get('extension_dir');")/xdebug.so /tmp
+
+FROM bref/php-${PHP_VERSION}-fpm as build_dev
+
+USER root
+COPY --from=build_extensions /tmp/*.so /tmp/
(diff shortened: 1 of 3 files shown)


runtime/layers/fpm-dev/Dockerfile
```diff
@@ -1,14 +1,28 @@
...
+
+FROM bref/php-${PHP_VERSION}-fpm as build_dev
+
+USER root
+COPY --from=build_extensions /tmp/*.so /tmp/
+RUN cp /tmp/*.so $(php -r "echo ini_get('extension_dir');")
+
+FROM bref/php-${PHP_VERSION}-fpm
+
+COPY --from=build_dev  /opt /opt
```
mnapoli: Can we replace this:

```Dockerfile
FROM bref/php-${PHP_VERSION}-fpm as build_dev

USER root

FROM bref/php-${PHP_VERSION}-fpm
```

with this:

```Dockerfile
FROM bref/php-${PHP_VERSION}-fpm

```

?

That way it's simpler, it avoids going through another intermediary container. Or isthere a reason not to do that?

jenschude: Tried it first. But the lambci container is defined to use some other user which has no rights to write to the extension folder. This is the reason for first change the user to root then copy to extension folder

mnapoli: OK noted, thanks!

PR #499, opened 2019-11-08
add test to check if extensions are loadable

jenschude:
In order to prevent broken extensions we also instantiate the containers once with all extensions enabled to check if there are some errors.

Replaces #492


mnapoli:
Thanks that's great!

PR #501, opened 2019-11-10
Expose the request context in the FPM layer via a global variable

pl4yradam:
This request context is needed if a user is using the authentication type of IAM_ROLE, when a user authenticates.

When using the IAM_ROLE authentication type with serverless, a user will pass their v4 signature token to API Gateway. API Gateway will then check and authorise the user which then invokes the lambda function, when invoking the lambda function API Gateway will pass in an updated requestContext object which contains Cognito information about the user. This prevents the requirement of having to pass user ID's publicly.

When using the layer without fpm & using the lambda function you can retrieve the event. 

Due to how FPM creates a local web server, handling the event is not accessible with the lambda function provided by bref. As the lambda function getting the LambdaApi request throws a no access error.

See below the example of the request context below contains information about the Cognito user, however in the fpmphp file there has unfortunately been no object added to this. I have written a pull request and adjusted a previous test that actually contained a requestContext but hadn't been tested to account for it.

> requestContext": {
        "resourceId": "n52n1n689a",
        "resourcePath": "/",
        "httpMethod": "GET",
        "extendedRequestId": "C9gIXGW7joEFVxA=",
        "requestTime": "10/Nov/2019:20:59:33 +0000",
        "path": "/users",
        "accountId": "<REMOVED>",
        "protocol": "HTTP/1.1",
        "stage": "dev",
        "domainPrefix": "secure-api",
        "requestTimeEpoch": 1573419573440,
        "requestId": "d679114c-4a3d-42d8-862a-892d036f7e2f",
        "identity": {
            "cognitoIdentityPoolId": "<REMOVED>",
            "accountId": "<REMOVED>",
            "cognitoIdentityId": "<REMOVED>",
            "caller": "<REMOVED>",
            "sourceIp": "<REMOVED>",
            "principalOrgId": "<REMOVED>",
            "accessKey": "<REMOVED>",
            "cognitoAuthenticationType": "authenticated",
            "cognitoAuthenticationProvider": "<REMOVED>",
            "userArn": "<REMOVED>",
            "userAgent": "<REMOVED>",
            "user": "<REMOVED>"
        },
        "domainName": "<REMOVED>",
        "apiId": "<REMOVED>"
    },`

### DIFF
--- a/src/Runtime/PhpFpm.php
+++ b/src/Runtime/PhpFpm.php
@@ -212,7 +212,7 @@ final class PhpFpm
         $request->setCustomVar('QUERY_STRING', $queryString);
         if ($event['requestContext'] ?? false) {
             $requestContext = [];
-            foreach ($event['requestContext'] as $key => $value) {
+            foreach (json_decode(json_encode($event['requestContext']),true) as $key => $value) {
                 $requestContext[strtoupper($key)] = $value;
                 $request->setCustomVar('REQUEST_CONTEXT_' . strtoupper($key), $value);
             }
--- a/src/Runtime/PhpFpm.php
+++ b/src/Runtime/PhpFpm.php
@@ -213,7 +213,7 @@ final class PhpFpm
         if ($event['requestContext'] ?? false) {
 
             $changeToAssoc = $event['requestContext'];
-            echo $changeToAssoc;
+            print_r($changeToAssoc);
             $requestContext = [];
             foreach (json_decode(json_encode($event['requestContext']),true) as $key => $value) {
                 $requestContext[strtoupper($key)] = $value;
--- a/src/Runtime/PhpFpm.php
+++ b/src/Runtime/PhpFpm.php
@@ -211,7 +211,13 @@ final class PhpFpm
         $request->setCustomVar('PATH_INFO', $event['path'] ?? '/');
         $request->setCustomVar('QUERY_STRING', $queryString);
         if ($event['requestContext'] ?? false) {
-            $request->setCustomVar('REQUEST_CONTEXT', json_decode(json_encode($event['requestContext']), true));
+
+            $requestContext = [];
+
+            foreach ($event['requestContext'] as $key => $value) {
+                $requestContext[$key] = $value[0];
+                $request->setCustomVar('REQUEST_CONTEXT_' . strtoupper($key), $requestContext);
+            }
         }
         // See https://stackoverflow.com/a/5519834/245552
         if (! empty($requestBody) && $method !== 'TRACE' && ! isset($headers['content-type'])) {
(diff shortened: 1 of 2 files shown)


mnapoli:
I see, that addition makes sense, thanks for working on that!

One thing that I am wondering about: could the global variable `$_SERVER['REQUEST_CONTEXT']` contain the array instead of JSON? Have you tried?

That would be simpler than having to decode JSON in PHP, which isn't really consistent with how global FPM variables work today.

pl4yradam:
> I see, that addition makes sense, thanks for working on that!
> 
> One thing that I am wondering about: could the global variable `$_SERVER['REQUEST_CONTEXT']` contain the array instead of JSON? Have you tried?
> 
> That would be simpler than having to decode JSON in PHP, which isn't really consistent with how global FPM variables work today.

I have experimented with setCustomVar - It only excepts a string as a 2nd param.
In which case im following suit like the below HTTP loop which simply prefixes with REQUEST_CONTEXT this keeps it consistent with the other methods in this code.

pl4yradam:
leaving it to there now, attempted multiple way to change to array however unsuccessfull

mnapoli:
I really want to make this happen. I see that now you populate a global `$_SERVER` key for each variable. I am wary about that solution:

- since we populate variables dynamically we have no control over, it is always a "warning" flag for me (regarding security and things we didn't anticipate)
- the list of variables available is not discoverable easily, and it is not consistent with the AWS documentation available online

I think we need to find a different solution. The JSON one seemed like a better option TBH. I wish we could find a way to make the array available in PHP 🤔 FastCGI does this with POST data, do we know if PHP has special handling for this global variable?

pl4yradam:
Okay i've spent more time on this solution now than I would like, The correct interpretation would be to pass the json through as is and use json_decode(). The second solution which was proposed by the author @mnapoli you wanted it to be under server variables which again I have catered for to now go back to the json idea?

I have forked and implemented my fix for my own work at the moment, people are welcome to use the code I have presented. 

If you wish for me to further develop this on the short free time I have I am welcome for sponsors to contribute.

It may also be worth documenting in the bref docs the limitations of using FPM and the lambda function solo.

adamprescott:
I've been playing about with Bref for work to see if it's viable. I came across the same issue that this attempts to solve.

As for my opinion which would be the more elegant solution. FPM is kind of being used as a crutch anyway to transform execution into a web service context. Passing the payload as JSON is probably the best we're going to get, unless there's another feature within FPM that can be used.

mnapoli:
Hi, just to clarify, this is the ideal solution I described (maybe I wasn't clear enough sorry):

```php
$id = $_SERVER['REQUEST_CONTEXT']['cognitoIdentityId'];
```

pl4yradam:
You will need to do this with all of the returned gateway response keys also. Feel free to update the PR

mnapoli:
Thanks, a rebased version of this PR has been merged in #542

PR #502, opened 2019-11-11
add test to check if development extensions have been loaded

jenschude:
This PR enhances the check for extensions if the loading of development extensions using the php/conf.dev.d/php.ini is working.


mnapoli:
Thanks!

PR #504, opened 2019-11-12
Fix fpm-dev-gateway

jenschude:
When changing the listening port for the FPM dev container I forgot about the nginx configuration. So this fixes the dev gateway to be able to talk to the fpm again.


mnapoli:
Thanks!

PR #505, opened 2019-11-13
Encode body responses in base64 only when configured

mnapoli:
This pull request fixes the issues of base64 encoded responses brought up in #482. 

It turns out we failed at properly implementing a solution with binary responses: the solution we added introduced a regression. We merged #494 which mitigated the impact, but this wasn't covering all cases. It turns out that covering all cases is impossible. Sorry about the back and forth!

This pull request implements a solution in a different direction. It requires slightly more effort from users, but at least it is completely predictable.

From now on, to support binary responses, users will have to add 2 things instead of 1 to `serverless.yml`.

Before this PR:

```yaml
provider:
    # ...
    apiGateway:
        binaryMediaTypes:
            - '*/*'
```

After this PR:

```yaml
provider:
    # ...
    apiGateway:
        binaryMediaTypes:
            - '*/*'
    environment:
        BREF_BINARY_RESPONSES: 1
```

I wish things could be made simpler, but I cannot see a reliable way out of this. I am thinking of making the Serverless plugin take care of that later (auto-defining the environment variable), but first let's fix the issue.

### DIFF
--- a/docs/runtimes/http.md
+++ b/docs/runtimes/http.md
@@ -116,7 +116,7 @@ In some cases however, you will need to serve images (or other assets) via PHP. 
 
 ## Binary responses
 
-By default API Gateway **does not support binary HTTP responses** like images, PDF, binary files… To achieve this, you need to enable the option for binary responses in `serverless.yml`:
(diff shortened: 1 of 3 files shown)


Guillaume-Rossignol:
:+1:

PR #509, opened 2019-11-23
Fix #146 Make layers lighter by preserving symlinks

### LINKED ISSUE #146: Symlinks are flattened in the layers because of a SAM bug
Here is a link to the open issue over at awslabs/aws-sam-cli#878

The bottom line is that `sam invoke local` fails to unpackage symlinks properly, which means you can not test anything locally that relies on a layer (or function for that matter) that has symlinks.

This is only an issue for local invokation. Everything unpackages properly in the cloud. However, it is a major aggrevation for anyone who is expecting to test/debug localy with SAM.

mnapoli:
This change avoids symlinks to be flattened in the AWS Lambda layers.

That means that symlinks are no longer replaced by the file they point to. Those files are not duplicated anymore.

This makes the layer about twice lighter!

Here is a comparison of the zip files (before/after):

Capture d’écran 2019-11-23 à 00 12 33
Capture d’écran 2019-11-23 à 13 19 42

And when unzipped:

Capture d’écran 2019-11-23 à 00 12 53

This is good news because:

- this leaves more room to the application code for the 250Mb of space available on disk on Lambda
- this leaves more room for the 50Mb max zip size for the whole application
- smaller layers mean faster cold start: it would be great to measure if we see an improvement

TODO:

- [ ] test on AWS Lambda
- [ ] test with `serverless invoke local`
- ❎ test with the "runtimes" test suite in Bref

### DIFF
--- a/runtime/Makefile
+++ b/runtime/Makefile
@@ -10,11 +10,12 @@ layers: export/console.zip export/php-72.zip export/php-73.zip export/php-74.zip
 
 # The PHP runtimes
 export/php%.zip: docker-images
-	PHP_VERSION=$$(echo $@ | cut -d'/' -f 2 | cut -d'.' -f 1);\
-	rm -f $@;\
-	mkdir export/tmp ; cd export/tmp ;\
-	docker run --entrypoint "tar" bref/$$PHP_VERSION:latest -ch -C /opt .  |tar -x;zip --quiet --recurse-paths ../$$PHP_VERSION.zip . ;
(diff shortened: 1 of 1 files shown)


mnapoli:
Tried again and the layers do not work anywhere, gotta figure this out later.

mnapoli:
Closing since I wasn't able to make them work for now.

Feel free to pick that up again and test it, tools (like serverless or SAM) evolve quickly and they might fix this soon.

PR #512, opened 2019-11-24
add documentation about antidot framework

kpicaza:
Thank you for your great work with this tool, bringing the Serverless universe to the PHP community.
I have added a new documentation page to configure an Antidot Framework application using Bref, with the intention of opening more options to the users of this new framework.

### DIFF
--- a/couscous.yml
+++ b/couscous.yml
@@ -100,6 +100,10 @@ menu:
                 text: Symfony
                 url: /docs/frameworks/symfony.html
                 title: Learn how to deploy serverless Symfony applications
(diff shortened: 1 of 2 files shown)


mnapoli:
Hi, thank you for contributing!

To be honest, I think it is worth keeping the documentation about the major frameworks for now. Maintaining the docs is a lot of effort.

Would you be able to publish those docs somewhere? For example in a GitHub repository available to everyone, or even better: in the documentation of the framework?

Sorry and thank you again!

PR #513, opened 2019-11-24
(Re-)add support for PSR-7-speaking lambdas

marcusirgens:
A lot of modern frameworks communicate using PSR-7 objects instead of dumping data out. PSR-7 objects can easily be converted to API Gateway data objects. This should make stuff faster™️, as we won't need to initialize the CGI runtime.

The values of the event and context are passed to the ServerRequestInterface, so they can be accessed wherever the request is used.


mnapoli:
Thanks so much for working on this, please have a look at #521 that builds upon this.

Nyholm:
This was fixed by #521. 

This PR should be closed.

mnapoli:
👍  thank you both!

PR #514, opened 2019-11-24
Add test suite to composer

marcusirgens:
Make it more obvious which tests should be performed by adding the testsuite to composer as `composer test`.


mnapoli:
Thanks!

PR #517, opened 2019-12-11
Update documentation

malitta:
Add link to 12-Factor App as a starting point for legacy applications.

Users with legacy application who are keen to move to a serverless environment might not be aware of the steps to be followed to upgrade their app. 12-Factor App methodology provides a general guideline of things to follow to make an app microservice compatible.


mnapoli:
Thank you!

PR #522, opened 2019-12-23
Example to simplify the logic

Nyholm:
FYI @mnapoli

### DIFF
--- a/src/Handler/BrefHandler.php
+++ b/src/Handler/BrefHandler.php
@@ -7,11 +7,11 @@ use Bref\Context\Context;
 /**
  * Handles any kind of Lambda events.
  */
-interface Handler
+interface BrefHandler
 {
     /**
      * @param mixed $event The raw event data.
      * @return mixed|void
      */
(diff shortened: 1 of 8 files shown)


Nyholm:
Hm... If we add a `BrefHandler::boot()` and `BrefHandler::afterEvent()` then we can get rid of `runtime/layers/fpm/bootstrap`.

mnapoli:
Thanks! That makes indeed the internals of Bref simpler, but that pushes the complexity onto Bref users.

This is what I mean:

```php
- return new MySqsHandler();
+ return new SqsWrapper(new MySqsHandler());
```

As a user, that doesn't really make sense.

Regarding `__invoke()`, I am a bit wary of this. We used to go with `__invoke()` with middlewares and it caused problems (e.g. autocompletion, impossibility to implement multiple interfaces using `__invoke()`, …). I don't think making a handler invocable is that useful, since the only thing that invokes them is Bref, we can adapt (i.e. we don't necessarily need to optimize the DX for the caller, since that's us).

PR #524, opened 2020-01-01
Fix boolean type casting

ionl:
whereas construct attempt string
we can't see die message :)

### DIFF
--- a/src/Runtime/LambdaRuntime.php
+++ b/src/Runtime/LambdaRuntime.php
@@ -38,7 +38,7 @@ final class LambdaRuntime
 
     public static function fromEnvironmentVariable(): self
     {
-        return new self((string)getenv('AWS_LAMBDA_RUNTIME_API'));
+        return new self((string) getenv('AWS_LAMBDA_RUNTIME_API'));
     }
 
     public function __construct(string $apiUrl)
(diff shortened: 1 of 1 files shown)


src/Runtime/LambdaRuntime.php
```diff
@@ -38,7 +38,7 @@ final class LambdaRuntime
 
     public static function fromEnvironmentVariable(): self
     {
-        return new self(getenv('AWS_LAMBDA_RUNTIME_API'));
+        return new self((string)getenv('AWS_LAMBDA_RUNTIME_API'));
```
mnapoli:
```suggestion
        return new self((string) getenv('AWS_LAMBDA_RUNTIME_API'));
```

to fix the coding style

Nyholm:
Thank you @ionl for this PR. 

I like it, there is only one small issue with the coding standard. Matthieu made a suggestion, could you just commit that suggestion and this PR will be merged.

src/Runtime/LambdaRuntime.php
ionl: Sorry for the late, you have right for the coding style (phpStorm don't apply it!)
I see it in https://www.php-fig.org/psr/psr-12/
Thank you

mnapoli: 👍 

You need to commit the suggestion. To do that, you need to click on "Unresolve conversation", then click on "Commit suggestion".

The other option is that you change the code yourself, commit and push on the branch to update the pull request.

ionl: I'm not a regular, thank for precisions ;)

mnapoli:
Thank you!

PR #530, opened 2020-01-16
Fix serverless-yml.md runtime usage href link

ganey:
Fix the runtime usage href link within the plugin section

The link in the docs currently takes users to the wrong / missing URL


mnapoli:
haha good catch, thank you!

PR #531, opened 2020-01-17
Adding WebP support to PHP's GD extension

aran112000:


runtime/base/base.Dockerfile
```diff
@@ -384,7 +384,8 @@ RUN set -xe; cd ${POSTGRES_BUILD_DIR}/src/include && make install
 # libicu-devel : needed for
 # libpng-devel : needed for gd
 # libjpeg-devel : needed for gd
+# libwebp-devel : needed for gd with WebP support
```
mnapoli: 👍  thank you!

tomhatzer:
@mnapoli do you possibly have any news about this? The 2 issues were closed and there didn't happen anything new after that. Is there anything where we can assist or help out with? 🙂  Thank you!

mnapoli:
Hi, yes the question here that is blocking the pull request is: how much did the layer grow (in size)?

See for example https://github.com/brefphp/extra-php-extensions/issues/10#issuecomment-575568999 and https://github.com/brefphp/extra-php-extensions/issues/10#issuecomment-576000931

tomhatzer:
Hi @mnapoli - thanks for the information!

I tried getting everything up and running (build process and compilation had no errors) but the tests are failing all the time because the file libwebp.so.4 file is not found.

Error output:
```
# Run tests
php layers/tests.php
.php: error while loading shared libraries: libwebp.so.4: cannot open shared object file: No such file or directory
PHP Fatal error:  Uncaught TypeError: trim() expects parameter 1 to be string, null given in /Downloads/bref/runtime/layers/tests.php:27
Stack trace:
#0 /Downloads/bref/runtime/layers/tests.php(27): trim(NULL)
#1 {main}
  thrown in /Downloads/bref/runtime/layers/tests.php on line 27

Fatal error: Uncaught TypeError: trim() expects parameter 1 to be string, null given in /Downloads/bref/runtime/layers/tests.php:27
Stack trace:
#0 /Downloads/bref/runtime/layers/tests.php(27): trim(NULL)
#1 {main}
  thrown in /Downloads/bref/runtime/layers/tests.php on line 27
make: *** [docker-images] Error 255
```

Is there something in the cleanup process that removes this file or does not include the file while copying?

Thanks!
Tom

mnapoli:
I think we can close this PR now: in Bref 1.0 gd will be moved to a separate layer, @shouze is working on this here: https://github.com/brefphp/extra-php-extensions/pull/76 He also added WebP support.

PR #532, opened 2020-01-18
Update docs to also reflect binary requests/uploads

Nyholm:


Nyholm:
Note: 
- upload stuff directly to S3 via pre-signed URLs. That is faster, allows for retries (with the JS library), works around limitations from API Gateway or PHP (upload size for example)
- download stuff directly from S3 via signed URLs: the download link is a PHP route which generates a signed URL and redirects to it. The S3 signed URL can be active for example for a few minutes/hours. That is faster and allows for any response size (API Gateway allows for 6Mb max)

This should maybe also be in the docs

mnapoli:
Awesome thanks!

Nyholm:
Thank you for merging.

PR #533, opened 2020-01-19
Include usage of VIEW_COMPILED_PATH in laravel.md

acairns:
It wasn't clear I needed to use this env var within the view configuration.

### DIFF
--- a/docs/frameworks/laravel.md
+++ b/docs/frameworks/laravel.md
@@ -73,6 +73,15 @@ SESSION_DRIVER=array
 LOG_CHANNEL=stderr
 ```
 
+Next we need to use the new `VIEW_COMPILED_PATH` env var when it has been set. Update the `config/view.php` file:
+
+```php
+    'compiled' => env(
+        'VIEW_COMPILED_PATH',
+        realpath(storage_path('framework/views'))
+    ),
+```
+
 Finally we need to edit `app/Providers/AppServiceProvider.php` because Laravel will not create that directory automatically:
 
 ```php


docs/frameworks/laravel.md
```diff
@@ -73,6 +73,15 @@ SESSION_DRIVER=array
...
 ```
 
+Next we need to use the new `VIEW_COMPILED_PATH` env var when it has been set. Update the `config/view.php` file:
+
+```php
+    'compiled' => env(
+        'VIEW_COMPILED_PATH',
+        realpath(storage_path('framework/views'))
+    ),
+```
```
mnapoli: Thank you. This piece of code however already exists in Laravel: https://github.com/laravel/laravel/blob/master/config/view.php#L31-L34

Are you sure you are not using an older version?

mnapoli:
Closing until further information.

acairns:
Apologies for the radio silence!

You're spot-on. Looks like the env var for `VIEW_COMPILED_PATH` was introduced for 5.7:
https://github.com/laravel/laravel/commit/5ea6fe18a89c3d0f5c0860d3777bff97510577b5#diff-968c1b3255d822f567528b245ba97df6

I'm on 5.7, but I maybe missed an upgrade step.

Could be worth noting for versions 5.6 and lower?

mnapoli:
I don't think we should support below 5.7, else the documentation will be too complex.

However if you find anything missing with the current docs (for 5.7) I'm happy to merge a fix :)

Issue #456, opened 2019-09-16
Query String encoding issues when using ALB


tobyryuk:
When running API Gateway, query strings in the event object are automagically URL decoded but when using Application Load Balancer, they're not. This leads to a problem where the `getQueryString` function ([here](https://github.com/brefphp/bref/blob/ccabf70f2e8249ed6cbb52144b8f2af62ffc5533/src/Runtime/PhpFpm.php#L297)) ends up URL encoding ALB query strings twice. ([more information can be found here](https://serverless-training.com/articles/api-gateway-vs-application-load-balancer-technical-details/))

Here's an example of what we get out when using these parameters `?single=value%20with%20spaces&multi[rise%20of%20planets]=hello%20world&multi[death%20of%20planets]=bye%20world` running Laravel.

API Gateway
![Screenshot 2019-09-16 at 15 20 48](https://user-images.githubusercontent.com/9976439/64961566-1cb79c80-d896-11e9-8d7f-c7b8898b3234.png)

Application Load Balancer
![Screenshot 2019-09-16 at 15 20 39](https://user-images.githubusercontent.com/9976439/64961571-204b2380-d896-11e9-98f4-f76fcd3c56ef.png)

From what I've found so far, there's no super clear way of figuring out if the invokation is from API Gateway or Application Load Balancer but what we could do is check the event and see if the API Gateway specific event properties are there or not and URL decode the result of the `http_build_query` once before returning it, if not from API Gateway. ([event differences from SO](https://stackoverflow.com/a/51729273))

tobyryuk:
I've temporarily solved this by overwriting the `PhpFpm.php` file before deploy where I wrap `http_build_query` in `urldecode`.

pmayet:
https://forums.aws.amazon.com/thread.jspa?threadID=297772

nuvolapl:
https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html

> If the query parameters are URL-encoded, the load balancer does not decode them. You must decode them in your Lambda function.

So, unfortunately, we need to handle it in the lambda function runtime.

deleugpn:
I've been using Lambda with Bref behind ALB quite a lot and I'm not sure why I've been oblivious to this issue.
Is there something specific my requests need in order to cause the problem? I use Laravel, is it relevant for the issue?
[...]

PR #538, opened 2020-01-23
Fixed Query String encoding issues when using ALB

nuvolapl:
Because there is an issue with query string encoding when using ALB.

https://github.com/brefphp/bref/issues/456

### DIFF
--- a/src/Event/Http/HttpRequestEvent.php
+++ b/src/Event/Http/HttpRequestEvent.php
@@ -163,6 +163,17 @@ final class HttpRequestEvent implements LambdaEvent
             return '';
         }
 
+        $queryStringParameters = [];
+
+        foreach ($this->event['queryStringParameters'] as $key => $value) {
+            \parse_str("{$key}={$value}", $params);
+
+            $queryStringParameters = \array_merge_recursive(
+                $queryStringParameters,
+                $params
+            );
+        }
+
         /*
          * Watch out: do not use $event['queryStringParameters'] directly!
          *
@@ -172,7 +183,7 @@ final class HttpRequestEvent implements LambdaEvent
          * ?array[key]=value gives ['array[key]' => 'value'] while we want ['array' => ['key' = > 'value']]
          * In that case we should recreate the original query string and use parse_str which handles correctly arrays
          */
-        return http_build_query($this->event['queryStringParameters']);
+        return http_build_query($queryStringParameters);
     }
 
     private function extractHeaders(): array
--- a/tests/Event/Http/CommonHttpTest.php
+++ b/tests/Event/Http/CommonHttpTest.php
@@ -69,8 +69,27 @@ abstract class CommonHttpTest extends TestCase implements HttpRequestProxyTest
                 'val2' => ['bar'],
             ],
         ]);
-        $this->assertQueryString('vars%5Bval1%5D=foo&vars%5Bval2%5D%5B%5D=bar');
-        $this->assertUri('/path?vars%5Bval1%5D=foo&vars%5Bval2%5D%5B%5D=bar');
+        $this->assertQueryString('vars%5Bval1%5D=foo&vars%5Bval2%5D%5B0%5D=bar');
+        $this->assertUri('/path?vars%5Bval1%5D=foo&vars%5Bval2%5D%5B0%5D=bar');
+    }
+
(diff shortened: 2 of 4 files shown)


tests/Runtime/PhpFpmTest.php
```diff
@@ -144,11 +144,57 @@ public function test request with arrays in query string()
                 ],
             ],
             '$_SERVER' => [
-                'REQUEST_URI' => '/?vars%5Bval1%5D=foo&vars%5Bval2%5D%5B%5D=bar',
+                'REQUEST_URI' => '/?vars%5Bval1%5D=foo&vars%5Bval2%5D%5B0%5D=bar',
```
mnapoli: Thank you, could you explain why this existing test has been changed?

nuvolapl: This is a default behavior of `\mb_parse_str` function.

mnapoli: Thanks, but is it a regression?

Is this change valid? What do the extra `0` mean here?

nuvolapl: I don't see any threats. `foo[]=bar` is equal to `foo[0]=bar` and `foo[]=bar&foo[]=baz` to `foo[0]=bar&foo[1]=baz`.

mnapoli: Ah OK I see, this is the array key. Thank you for the detail.

I need to look more into this and run some extra tests. Then we can merge this.

Thank you!

nuvolapl:
@mnapoli 
b16

mnapoli:
😄 

I am afraid that this behavior is needed in ALB, but will result in double decoding in other scenarios (API Gateway).

What do you think?

nuvolapl:
> I am afraid that this behavior is needed in ALB, but will result in double decoding in other scenarios (API Gateway).
> 
> What do you think?

I think that redundant code executions is better than broken (popular) scenario with ALB service. There is a third solution - condition to check if it is APIGateway or ALB however the code will be more complicated and its perfomance will be comparable (tests needed).

mnapoli:
> I think that redundant code executions is better than

I am afraid that no, it might lead to security issues for example.

I don't think performance will be a problem if we do a different behavior, but it will not be easy to implement cleanly I think.

nuvolapl:
Security issue? what security issues? 

Anyway - you stay with the statement that bref for ALB doesn't have to work?

mnapoli:
> Security issue? what security issues?

Those are always complex to anticipate. But I've learned over time that over-encoding or under-encoding must not be taken lightly, unfortunately.

> Anyway - you stay with the statement that bref for ALB doesn't have to work?

No, we should make it work. As I said, what's probably best it to differentiate whether it's ALB or API Gateway: if it's ALB, we should apply the decoding then.

The other option is to demonstrate that there is no risk of bug and security if we over-decode a string (and then we can merge this as-is). But how can we do that? I'm not really sure…

deleugpn:
Looking over the linked issue (https://github.com/brefphp/bref/issues/456) and the doc posted there (https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html) I was wondering how reliable it would be to check the headers host for `%.elb.amazonaws.com`. If we have a custom domain in front of the ALB, does it mean the headers change and makes this unreliable?

mnapoli:
If we can detect wether the event comes from ALB, that would be great.

> If we have a custom domain in front of the ALB, does it mean the headers change and makes this unreliable?

I have no idea (I don't use ALB at all).

matthewsuan:
Hi @mnapoli ,

we can detect from the `requestContext` object whether a request is coming from ALB.

```
{
    "requestContext": {
        "elb": {
            "targetGroupArn": "arn:aws:elasticloadbalancing:us-east-2:123456789012:targetgroup/lambda-279XGJDqGZ5rsrHC2Fjr/49e9d65c45c6791a"
        }
    },
```

https://docs.aws.amazon.com/lambda/latest/dg/services-alb.html

mnapoli:
Should be fixed by #693

Sorry for not being more active here, maintaining open source projects like this with many use cases, and some that I don't use or know about is hard :/  #693 felt more familiar/easy to read and in-line with the existing code base, it was easier to merge.

PR #540, opened 2020-01-26
Turn Handler interfaces into abstract classes

mnapoli:
See #521

### DIFF
--- a/src/Runtime/LambdaRuntime.php
+++ b/src/Runtime/LambdaRuntime.php
@@ -5,13 +5,7 @@ namespace Bref\Runtime;
 use Bref\Context\Context;
 use Bref\Context\ContextBuilder;
 use Bref\Event\Handler;
-use Bref\Event\Http\HttpRequestEvent;
-use Bref\Event\Http\HttpResponse;
-use Bref\Event\Http\Psr7RequestFactory;
-use Bref\Event\S3\S3Event;
-use Bref\Event\S3\S3Handler;
-use Bref\Event\Sqs\SqsEvent;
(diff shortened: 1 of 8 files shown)


Nyholm:
Yes!
I suggested that in https://github.com/brefphp/bref/pull/521#issuecomment-578481639

But why make them abstract? 

See my example of a SymfonyHandler: https://github.com/Nyholm/small-symfony-project/blob/master/src/SymfonyRequestHandler.php

src/Runtime/LambdaRuntime.php
```diff
@@ -101,17 +95,12 @@ public function processNextEvent($handler): void
         $result = null;
 
         try {
+            // PSR-15 adapter
             if ($handler instanceof RequestHandlerInterface) {
```
Nyholm: Since this is the only PSR interface for handling an incoming HTTP request, Im fine with this special logic. I dont believe we should add more though.

mnapoli: Makes sense. Adding Symfony support without adding it in here wouldn't be too bad anyway, it could be like your `SymfonyRequestHandler`, but it could go even further like this:

```php
<?php

use App\Kernel;
use Bref\Context\Context;
use Symfony\Component\HttpFoundation\Request;
use App\SymfonyRequestHandler;

include_once dirname(__DIR__).'/config/bootstrap.php';

return new SymfonyHandler($kernel);
```

And the class `SymfonyHandler` would implement `Handler` and invoke the kernel with a Symfony request.

PR #542, opened 2020-01-26
Rebase: Expose the request context in the FPM layer via a global variable

### LINKED ISSUE #500: Get the request context in PHP when using the FPM layer
Hi guys,

Using lumen with bref and it absolutely works fine - Just wondering how to get the event data passed in like the docs here: 
https://bref.sh/docs/runtimes/function.html

Im trying to pull the event out of a Lumen controller but having no luck.

Further testing:
lambda() CAN NOT be used with fpm in place with lumen.

Figured out that the fpm script wasn't catering for it: https://github.com/brefphp/bref/pull/501

### LINKED ISSUE #506: Web application does not get event or context info
When using API gateway with lamda proxy, php does not get any event or context information. I'm hoping to get authorizer info passed from cognito pool, as well as data from PUT and DELETE methods (STDIN vs $_GET and $_POST) for a REST API.

Nyholm:
Rebased #501, used JSON instead of adding many global variables. 

It is not possible to add an array, because it needs to be serialised and sent to PHP FPM. 

This will fix #506 and fix #500

### DIFF
--- a/src/Event/Http/FpmHandler.php
+++ b/src/Event/Http/FpmHandler.php
@@ -179,6 +179,9 @@ final class FpmHandler extends HttpHandler
         $request->setServerPort($event->getServerPort());
         $request->setCustomVar('PATH_INFO', $event->getPath());
         $request->setCustomVar('QUERY_STRING', $event->getQueryString());
+        if ($event['requestContext'] ?? false) {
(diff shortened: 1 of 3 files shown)


mnapoli:
Thanks!

I'll add documentation after the merge.

Nyholm:
Thank you for merging

PR #543, opened 2020-01-26
Add event and context to the Psr15Handler request

marcusirgens:
This adds the event and context objects to the request object.

When using a custom authorizer, for example, I need access to the array located at `$event["requestContext"]["authorizer"]`. This allows me to fetch the `HttpRequestEvent` from the `ServerRequestInterface` object to get the request metadata.

### DIFF
--- a/src/Event/Http/Psr15Handler.php
+++ b/src/Event/Http/Psr15Handler.php
@@ -18,6 +18,8 @@ class Psr15Handler extends HttpHandler
     public function handleRequest(HttpRequestEvent $event, Context $context): HttpResponse
     {
         $request = Psr7RequestFactory::fromEvent($event);
+        $request = $request->withAttribute('lambdaEvent', $event);
+        $request = $request->withAttribute('lambdaContext', $context);
 
         $response = $this->psr15Handler->handle($request);
 
--- a/src/Event/Http/Psr15Handler.php
+++ b/src/Event/Http/Psr15Handler.php
@@ -18,8 +18,8 @@ class Psr15Handler extends HttpHandler
     public function handleRequest(HttpRequestEvent $event, Context $context): HttpResponse
     {
         $request = Psr7RequestFactory::fromEvent($event);
-        $request = $request->withAttribute('lambdaEvent', $event);
-        $request = $request->withAttribute('lambdaContext', $context);
+        $request = $request->withAttribute('lambda-event', $event);
+        $request = $request->withAttribute('lambda-context', $context);
 
         $response = $this->psr15Handler->handle($request);
 
--- a/tests/Runtime/LambdaRuntimeTest.php
+++ b/tests/Runtime/LambdaRuntimeTest.php
@@ -4,6 +4,7 @@ namespace Bref\Test\Runtime;
 
 use Bref\Context\Context;
 use Bref\Event\Handler;
+use Bref\Event\Http\HttpRequestEvent;
 use Bref\Event\S3\S3Event;
 use Bref\Event\S3\S3Handler;
 use Bref\Event\Sqs\SqsEvent;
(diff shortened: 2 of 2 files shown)


Nyholm:
Im 👍 

I believe it make sense to "sync" the name of the parameter with https://github.com/brefphp/bref/pull/542

Or the other way around.

marcusirgens:
@Nyholm I'm not sure if `requestContext` is good, because although it makes sense semantically, I don't understand if it means the _context_ (the Context object) or the event _data_, which is the context for the request. I chose "lambdaEvent" to make the name slightly more unique.

Nyholm:
I agree. How about `LAMBDA_CONTEXT` and `LAMBDA_EVENT`?

marcusirgens:
Yep. Stick to camelCase in the attributes, you think?

Nyholm:
#542 needs to use snake case all caps because it will go in the super global and that is the convention. 

I thought that PSR7 server attributes were a place where you put server parameter. But this is just arbitrary data... the only few examples in the meta document is __BODY__. 

I guess we could use camelCase here since it is more beautiful.

mnapoli:
👍  that's a good improvement.

I thought there was some kind of convention (or standard) for PSR-7 attributes but I can't retreive it 🤔  Not sure if camelCase, snake_case or someting else should be used.

In [these middlewares](https://github.com/middlewares/psr15-middlewares) (which AFAIK are a bit popular) it seems it's `this-case` that is used. For example:

- https://github.com/middlewares/client-ip#usage
- https://github.com/middlewares/geolocation#example

src/Event/Http/Psr15Handler.php
```diff
@@ -18,6 +18,8 @@ public function __construct(RequestHandlerInterface $psr15Handler)
     public function handleRequest(HttpRequestEvent $event, Context $context): HttpResponse
     {
         $request = Psr7RequestFactory::fromEvent($event);
+        $request = $request->withAttribute('lambdaEvent', $event);
```
mnapoli: Does it make sense to expose the event? If one needs that, they can extend `HttpHandler` directly now instead of the PSR-7 handler. I don't think it brings added value.

marcusirgens: Well, there are a lot of batteries included in the existing implementation.
`\Bref\Event\Http\Psr7RequestFactory::fromEvent()` is marked as internal, so is
`\Bref\Event\Http\HttpResponse::fromPsr7Response();`, so if I want to replicate this behaviour I will need to write a lot of redundant code. I think it definitely makes sense to give the application direct access to the request context.

mnapoli: Right, I did not mean about the request context: adding this one makes sense, because else there is no way to retrieve some stuff like Cognito authentication.

But I am questioning included the `lambdaEvent` as well: what is the use case for this?

marcusirgens: I'm not sure I understand what you mean - the Context contains the `\Bref\Context\Context`, right? How will that pass on the Cognito authentication? That looks like it's specified in the event array.

mnapoli: Oh I'm sorry, maybe I'm mistaken!

OK here is what I assumed: both the `$context` variable and `$event['requestContext']` are the same thing. Are these different?

marcusirgens: Correct. `Context` is the `\Bref\Context\Context` object, which contains request metadata fields request ID, the function ARN, trace ID and the timeout deadline.

`$event` is the data passed to the function body. Here are some concrete of fields I have used today while working with PSR-7 handlers
- `$event["pathParameters"]` - I could parse the query string, but this is already parsed by API Gateway
- `$event["requestContext"]["authorizer"]["principalId"]` - the Authorizer user ID

mnapoli: OK thank you that makes sense then!

marcusirgens:
Fantastic, thanks for the merge!

PR #544, opened 2020-01-27
Add SNS event handler support

mnapoli:
This is a continuation of #521 to add support for SNS event handlers.

### DIFF
--- a/src/Event/Sns/MessageAttribute.php
+++ b/src/Event/Sns/MessageAttribute.php
@@ -1,0 +1,40 @@
+<?php declare(strict_types=1);
+
+namespace Bref\Event\Sns;
+
+/**
+ * SNS message attribute.
+ *
+ * @see https://docs.aws.amazon.com/sns/latest/api/API_MessageAttributeValue.html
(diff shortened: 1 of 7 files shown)


src/Event/Sns/MessageAttribute.php
```diff
@@ -0,0 +1,40 @@
...
+ * @see https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/SNSEvent.java
+ */
+class MessageAttribute
+{
+    /** @var array */
+    private $attribute;
+
+    public function __construct(array $attribute)
+    {
+        $this->attribute = $attribute;
```
AlessandroMinoccheri: May it be necessary to check that there is an Index Type and Value?

mnapoli: I'm not sure, at this point the shape of the event should have been validated earlier. I don't validate all keys in all value objects.

mnapoli:
Tests are failing because of a difference in date format in PHP 7.2 it seems 🤔

src/Event/Sns/MessageAttribute.php
AlessandroMinoccheri: ok, usually I validate all keys to ensure the correct use of the value object. But it's not a block for this PR I think

PR #549, opened 2020-02-01
Refactor the PSR-7 logic

mnapoli:
Logic related to PSR-7 and PSR-15 handlers was spread across multiple classes.

Because of that, some HTTP logic was applied to PHP-FPM but not to PSR-15 handlers, and vice-versa.

This PR:

- regroups global HTTP response logic into the `HttpResponse` class
- regroups PSR-7 related logic into a new `Psr7Bridge` class
- makes the `HttpResponse` class non-internal (because it can now be used by HTTP event handlers)

### DIFF
--- a/src/Event/Http/HttpResponse.php
+++ b/src/Event/Http/HttpResponse.php
@@ -4,8 +4,6 @@ namespace Bref\Event\Http;
 
 /**
  * Formats the response expected by AWS Lambda and the API Gateway integration.
- *
- * @internal
  */
 final class HttpResponse
 {
--- a/src/Event/Http/HttpResponse.php
+++ b/src/Event/Http/HttpResponse.php
@@ -2,8 +2,6 @@
 
 namespace Bref\Event\Http;
 
-use Psr\Http\Message\ResponseInterface;
-
 /**
  * Formats the response expected by AWS Lambda and the API Gateway integration.
  *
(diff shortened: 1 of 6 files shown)


src/Event/Http/HttpResponse.php
```diff
@@ -20,48 +16,40 @@ final class HttpResponse
     /** @var string */
     private $body;
 
-    public function __construct(string $body, array $headers, int $statusCode = 200)
+    public function __construct(string $body, array $headers = [], int $statusCode = 200)
```
mnapoli: This class will now be used by end-users. This helps making it more developer-friendly.

src/Event/Http/HttpResponse.php
```diff
@@ -20,48 +16,40 @@ final class HttpResponse
...
     {
         $this->body = $body;
         $this->headers = $headers;
         $this->statusCode = $statusCode;
     }
 
-    public static function fromPsr7Response(ResponseInterface $response): self
```
mnapoli: The logic related to PSR-7 was moved to the Psr7Bridge class.

The logic related to HTTP responses was kept in this class below (it will then apply to all HTTP responses).

src/Event/Http/Psr15Handler.php
```diff
@@ -17,12 +17,10 @@ public function __construct(RequestHandlerInterface $psr15Handler)
 
     public function handleRequest(HttpRequestEvent $event, Context $context): HttpResponse
     {
-        $request = Psr7RequestFactory::fromEvent($event);
-        $request = $request->withAttribute('lambda-event', $event);
-        $request = $request->withAttribute('lambda-context', $context);
```
mnapoli: These 2 lines were moved to the Psr7Bridge.

mnapoli:
I am going to release a new beta right now with these changes.

PR #555, opened 2020-02-03
Updated OSX Xdebug details with hostnames

toby-griffiths:
Recent versions of Docker for Mac provide host names to connect back to the host Mac, so I've added details of how to use these, if available, to simplify the Xdebug setup.

### DIFF
--- a/docs/local-development.md
+++ b/docs/local-development.md
@@ -141,14 +141,32 @@ Now start the debug session by issueing a request to your application in the [br
 
 #### OSX
 
-As Docker on Mac uses a virtual machine for running docker it can be tricky to listen to the xdebug port.
-By creating an alias for the loopback interface on the host with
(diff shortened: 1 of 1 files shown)


docs/local-development.md
```diff
@@ -141,14 +141,32 @@ Now start the debug session by issueing a request to your application in the [br
...
+You can then use this configuration start listening to xdebug connections in your preferred IDE (replacing `host.docker.internal` with the appropriate value for your Docker for Mac version:
+
+```ini
+[xdebug]
+xdebug.remote_enable = 1
+xdebug.remote_autostart = 0
+xdebug.remote_host = 'host.docker.internal'
+```
+
+##### Older versions of Docker for Mac
```
mnapoli: Thank you! I think we can remove this section then. It might be better to keep the documentation as "to the point"  as possible and not worry too much about outdated stuff.

mnapoli:
I'll remove the extra section after merging.

PR #556, opened 2020-02-13
Adds suport for custom vendor autoloading

jasonmccallister:
In order to keep the function code small, you can package your vendor dependencies in a custom layer. See here for more documentation: https://serverless.com/framework/docs/providers/aws/guide/layers/

Layers are version just like lambda. Chris Munns has an excellent talk about layers, including a demo, here: https://www.youtube.com/watch?v=gCQHulp3aVo

However, the current bootstrap file is hardcoded to look in the projects vendor path.

```php
require $appRoot . '/vendor/autoload.php'; // $appRoot is '/var/task'
```

In order to include composer as a layer, we need to override the path to load from vendor. This PR adds support for checking if `BREF_AUTOLOAD_PATH` is available as an environment variable and loads from that path instead of the default project root. 

In this example, I would define `BREF_AUTOLOAD_PATH` as `/opt/vendor/autoload.php`. 

One major plus size to this is the functions code is reduced drastically (in my case it went from 40 Mb to 123 Kb). This also enables the function code to be edited using the Lambda editor (not that you should but could).

### DIFF
--- a/runtime/layers/console/bootstrap
+++ b/runtime/layers/console/bootstrap
@@ -8,8 +8,13 @@ error_reporting(E_ALL);
 
 $appRoot = getenv('LAMBDA_TASK_ROOT');
 
-/** @noinspection PhpIncludeInspection */
-require $appRoot . '/vendor/autoload.php';
+if (getenv('BREF_CUSTOM_AUTOLOAD')) {
+    /** @noinspection PhpIncludeInspection */
+    require getenv('BREF_CUSTOM_AUTOLOAD');
+} else {
+    /** @noinspection PhpIncludeInspection */
+    require $appRoot . '/vendor/autoload.php';
+}
 
 $lambdaRuntime = LambdaRuntime::fromEnvironmentVariable();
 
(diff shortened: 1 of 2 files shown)


mnapoli:
That is awesome, thank you! This will also solve #226 which is great.

Regarding the variable name, prefixing with `BREF_` is what we've been doing until now so that's good, but how about `BREF_AUTOLOAD_PATH`? Any other suggestion is welcome as well. (I think mentioning that this is a path makes it a bit more explicit)

I wonder if it is possible to add a test to cover this. I can look into that later.

If you have time to add this to the docs that's great, but I can also take care of that when merging.

jasonmccallister:
@mnapoli that wording makes a lot more sense, I was struggling with the naming but I just pushed the changes.

jasonmccallister:
Also, as far as testing... If these are just PHP files, I'm not sure how to test besides making a custom vendor and "normal" vendor and making sure a dependency is loaded?

Where would you imagine the docs live? I don't see an area that really makes sense, maybe a new section about layers and customizing the environment?

dcsg:
@jasonmccallister I've tried to implement this without success. First the path to `autoload.php` was not correct because when I move the vendor to the custom layer it extracts everything to the root of `/opt` and I thought it would put everything inside a `vendor` folder. Second after getting this fixed it seems composer paths to my classes in the classmap are broken. Did you got this working properly?

Here are some samples of the code and error log: https://gist.github.com/dcsg/7cd756e0eace73ed9c5115d948fc2522

PR #561, opened 2020-02-16
Add DynamoDB typed events

jasonmccallister:
There are some considerations that are different for DynamoDB than other events. This major concern is marshaling of items from DynamoDB.

DynamoDB stores attributes as types, this has more context: https://aws.amazon.com/blogs/developer/dynamodb-json-and-array-marshaling-for-php/

There are a few things that should be discussed before continuing work, should the records returned from the event be automatically be unmarshaled? If they are not unmarshaled in the Bref DynamoDbRecord, the user/function will still need to unmarshal the record. Is it safe to assume that we should always unmarshal the record? 

Right now there is only one test (until we determine the behavior) that checks for the "normal form" record. https://github.com/brefphp/bref/blob/0f59619ba4329920c90d1e775cba25a60f50e985/tests/Event/DynamoDb/DynamoDbEventTest.php#L20

### DIFF
--- a/src/Event/DynamoDb/DynamoDbRecord.php
+++ b/src/Event/DynamoDb/DynamoDbRecord.php
@@ -43,25 +43,16 @@ final class DynamoDbRecord
         return $this->record['dynamodb']['OldImage'] ?? null;
     }
 
-    /**
-     * @return string
-     */
     public function getSequenceNumber(): string
     {
         return $this->record['dynamodb']['SequenceNumber'];
     }
 
-    /**
-     * @return int
-     */
     public function getSizeBytes(): int
     {
         return $this->record['dynamodb']['SizeBytes'];
     }
-
-    /**
-     * @return string
-     */
+    
     public function getStreamViewType(): string
     {
         return $this->record['dynamodb']['StreamViewType'];
--- a/src/Event/DynamoDb/DynamoDbRecord.php
+++ b/src/Event/DynamoDb/DynamoDbRecord.php
@@ -8,12 +8,11 @@ final class DynamoDbRecord
     private $record;
 
     /**
-     * DynamoDbRecord constructor.
-     * @param $record
+     * @param mixed $record
      */
     public function __construct($record)
     {
-        if (!is_array($record) || !isset($record['eventSource']) || $record['eventSource'] !== 'aws:dynamodb') {
+        if (! is_array($record) || ! isset($record['eventSource']) || $record['eventSource'] !== 'aws:dynamodb') {
             throw new \InvalidArgumentException('Event source must come from DynamoDB');
         }
 
--- a/src/Event/DynamoDb/DynamoDbRecord.php
+++ b/src/Event/DynamoDb/DynamoDbRecord.php
@@ -20,25 +20,35 @@ final class DynamoDbRecord
     }
 
     /**
-     * @return array
+     * Returns the key attributes of the modified item.
      */
-    public function getKeys()
+    public function getKeys(): array
     {
         return $this->record['dynamodb']['Keys'];
     }
 
     /**
-     * @return array|null
+     * Returns the new version of the DynamoDB item.
+     *
+     * Warning: this can be null depending on the `StreamViewType`.
+     *
+     * @see getStreamViewType()
+     * @see https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_StreamSpecification.html
      */
-    public function getNewImage()
+    public function getNewImage(): ?array
     {
         return $this->record['dynamodb']['NewImage'];
     }
 
     /**
-     * @return array|null
+     * Returns the old version of the DynamoDB item.
+     *
+     * Warning: this can be null depending on the `StreamViewType`.
+     *
+     * @see getStreamViewType()
+     * @see https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_StreamSpecification.html
      */
-    public function getOldImage()
+    public function getOldImage(): ?array
     {
         return $this->record['dynamodb']['OldImage'] ?? null;
     }
@@ -53,6 +63,9 @@ final class DynamoDbRecord
         return $this->record['dynamodb']['SizeBytes'];
     }
 
+    /**
+     * @see https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_StreamSpecification.html
+     */
     public function getStreamViewType(): string
     {
         return $this->record['dynamodb']['StreamViewType'];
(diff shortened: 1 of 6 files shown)


jasonmccallister:
I opted to not return the records as native PHP array and instead left them as the dynamodb types as the type can inform businesses logic/decisions when handling events.

mnapoli:
👍  that's very interesting.

The PHP AWS SDK currently supports only the basic (verbose) format when interacting with DynamoDB. It makes sense to me to offer the same kind of API/data structure here for consistency.

We can add a more high-level API/data structure later (because it would still makes sense for Bref to simplify that verbose mess). But we need to start with the basic structures and we need to support that.

So what you've implemented looks good to me!

One question is: which getters of the classes you added would be impacted if we later add that high-level representation? Is it easy to avoid future conflicts/confusion over naming?

Another question: are there different kind of events that DynamoDB can trigger in Lambda? I am wondering if the shape of the `DynamoDbRecord` can change and have different fields in some cases.

src/Event/DynamoDb/DynamoDbRecord.php
```diff
@@ -0,0 +1,69 @@
...
+     */
+    public function getSequenceNumber()
+    {
+        return $this->record['dynamodb']['SequenceNumber'];
+    }
+
+    /**
+     * @return mixed
+     */
+    public function getSizeBytes()
```
mnapoli: Should we type that into an int?

src/Event/DynamoDb/DynamoDbRecord.php
```diff
@@ -0,0 +1,69 @@
...
+     */
+    public function getOldImage()
+    {
+        return $this->record['dynamodb']['OldImage'] ?? null;
+    }
+
+    /**
+     * @return mixed
+     */
+    public function getSequenceNumber()
```
mnapoli: I wonder if we can have more precise types in this class, for example here is it a string maybe?

src/Event/DynamoDb/DynamoDbRecord.php
jasonmccallister: I will make that change!

src/Event/DynamoDb/DynamoDbRecord.php
jasonmccallister: I will update that to be a string

jasonmccallister:
@mnapoli I will make that changes and I agree with your thoughts completely. There is a really nice abstraction around DynamoDB that is here: https://github.com/guillermoandrae/php-dynamodb. Is that the kind of abstraction you are referring too?

mnapoli:
That's interesting thanks for the link, I didn't know about that. We have https://github.com/brefphp/dynamap which is still WIP as well.

The only thing I'm wondering about now is this: 

> are there different kind of events that DynamoDB can trigger in Lambda? I am wondering if the shape of the `DynamoDbRecord` can change and have different fields in some cases.

The only reason I'm worried about this is that changing the class would require a BC break. Since I don't know anything about DynamoDB events, maybe I am worrying about a problem that doesn't exist.

src/Event/DynamoDb/DynamoDbRecord.php
```diff
@@ -0,0 +1,60 @@
...
+    {
+        if (! is_array($record) || ! isset($record['eventSource']) || $record['eventSource'] !== 'aws:dynamodb') {
+            throw new \InvalidArgumentException('Event source must come from DynamoDB');
+        }
+
+        $this->record = $record;
+    }
+
+    /**
+     * @return mixed
```
mnapoli: Is it always an array? (typing array would be more helpful to users)

src/Event/DynamoDb/DynamoDbRecord.php
```diff
@@ -0,0 +1,60 @@
...
+     */
+    public function getKeys()
+    {
+        return $this->record['dynamodb']['Keys'];
+    }
+
+    /**
+     * @return mixed
```
mnapoli: Same here, are the "images" always arrays?

src/Event/DynamoDb/DynamoDbRecord.php
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    public function getSizeBytes(): int
+    {
+        return $this->record['dynamodb']['SizeBytes'];
+    }
+
+    public function getStreamViewType(): string
+    {
+        return $this->record['dynamodb']['StreamViewType'];
```
mnapoli: I had a quick look at what this field means: https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_StreamSpecification.html

It seems some DynamoDb events can only contain keys, or new images, or old images.

Does that mean that we should make all these 3 getters nullable above?

jasonmccallister:
Thank you for the budge @pmayet, this completely dropped off my radar. I'm going to review my side and make sure I covered everything from @mnapoli

src/Event/DynamoDb/DynamoDbRecord.php
jasonmccallister: It will not always be an array depending on if the record is an update or delete.

src/Event/DynamoDb/DynamoDbRecord.php
jasonmccallister: Keys should not be nullable, dynamodb always requires/returns keys

src/Event/DynamoDb/DynamoDbRecord.php
jasonmccallister: updated

pmayet:
Hi @jasonmccallister,

Thanks for your update. I think than it's important to have the EventName. In my PR https://github.com/brefphp/bref/pull/609, I have had the method for that.

I also think it's important to have more test about the StreamViewType.

Look at https://github.com/brefphp/bref/pull/609, after I will close my PR.

You take first ;-)

pmayet:
Ping @jasonmccallister

mnapoli:
Hi both! I am 👍  for merging this PR after adding the `getEventName()` and the extra tests from #609.

I just pushed f27b479 to this PR as well to add more documentation in the code, and fix the PrettyCI build.

jasonmccallister:
Sorry for the delay on this, I had been heads down on another project. I'm going to review the other branch and add tests today.

jasonmccallister:
@mnapoli added those changes, ready for review!

pmayet:
@mnapoli ok for me, I'm using it in production since two weeks

@jasonmccallister Thank's for your work

mnapoli:
Awesome! Thanks for the feedback.

PR #568, opened 2020-02-20
Update docs to reflect Laravel 6. fixes #528

### LINKED ISSUE #528: Documentation for Laravel framework doesn't cover the last framework version
It seems, that the documentation for usage with Laravel framework doesn't cover the current Laravel Framwork 6.*

They are some bootstraping differences and when calling `mkdir(config('view.compiled'), 0755, true);` in `AppServiceProvider.php` throws and exception `mkdir(): Invalid path`

viezel:
The Laravel docs are slightly incorrect. This was also pointed out in #528. 
Ive also updated the docs in regards to the serverless.yml file, so the environment vars are shown and excluded packages to keep the build size down.

### DIFF
--- a/docs/frameworks/laravel.md
+++ b/docs/frameworks/laravel.md
@@ -61,13 +61,13 @@ functions:
 
 Now we still have a few modifications to do on the application to make it compatible with AWS Lambda.
 
-Since [the filesystem is readonly](/docs/environment/storage.md) except for `/tmp` we need to customize where the cache files are stored. Add this line in `bootstrap/app.php` after `$app = new Illuminate\Foundation\Application`:
+Since [the filesystem is readonly](/docs/environment/storage.md) except for `/tmp` we need to customize where the cache files are stored. Change the line in `bootstrap/app.php` at `$app = new Illuminate\Foundation\Application(`:
 
-```php
-/*
- * Allow overriding the storage path in production using an environment variable.
- */
-$app->useStoragePath($_ENV['APP_STORAGE'] ?? $app->storagePath());
+```diff
+$app = new Illuminate\Foundation\Application(
+-    realpath(__DIR__.'/../')
++    $_ENV['APP_BASE_PATH'] ?? dirname(__DIR__)
+);
 ```
 
 We will also need to customize the location for compiled views, as well as customize a few variables in the `.env` file:
--- a/docs/frameworks/laravel.md
+++ b/docs/frameworks/laravel.md
@@ -25,12 +25,6 @@ provider:
     name: aws
     region: us-east-1
     runtime: provided
-    environment:
-        # Laravel environment variables
-        APP_STORAGE: /tmp
-        LOG_CHANNEL: stderr
-        SESSION_DRIVER: array
-        VIEW_COMPILED_PATH: /tmp/storage/framework/views
 
 plugins:
     - ./vendor/bref/bref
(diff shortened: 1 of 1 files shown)


docs/frameworks/laravel.md
```diff
@@ -51,13 +61,13 @@ functions:
...
-```php
-/*
- * Allow overriding the storage path in production using an environment variable.
- */
-$app->useStoragePath($_ENV['APP_STORAGE'] ?? $app->storagePath());
+```diff
+$app = new Illuminate\Foundation\Application(
+-    realpath(__DIR__.'/../')
++    $_ENV['APP_BASE_PATH'] ?? dirname(__DIR__)
+);
```
mnapoli: It seems it's already the content of the file:

https://github.com/laravel/laravel/blob/c78a1d8184f00f8270d8a135cc21590837b6bfbd/bootstrap/app.php#L14-L16

Are you sure this step is needed now?

docs/frameworks/laravel.md
```diff
@@ -27,11 +27,21 @@ provider:
     runtime: provided
     environment:
         # Laravel environment variables
-        APP_STORAGE: '/tmp'
+        APP_STORAGE: /tmp
+        LOG_CHANNEL: stderr
+        SESSION_DRIVER: array
+        VIEW_COMPILED_PATH: /tmp/storage/framework/views
```
mnapoli: These new variables are already defined in the `.env` file (explained below in the article). Is that necessary to add them here?

viezel: I would argue it is. Some people just go to example code and copy/paste from it. Why not have the ideal solution in the sample app?

docs/frameworks/laravel.md
viezel: I’m not sure actually. I was by the understanding that we need the APP_BASE_PATH

docs/frameworks/laravel.md
mnapoli: The problem I see is that we define the same variables twice. That's unnecessary: it could be simpler. And having 2 times the values means that you need to modify both. Which one is actually used when deployed in production?

docs/frameworks/laravel.md
mnapoli: We need it, but it's already that code in Laravel.

What I am saying is that you shouldn't have to change anything, Laravel is already like this: https://github.com/laravel/laravel/blob/c78a1d8184f00f8270d8a135cc21590837b6bfbd/bootstrap/app.php#L14-L16

docs/frameworks/laravel.md
viezel: ok. I simplified it now.

docs/frameworks/laravel.md
```diff
@@ -49,20 +53,11 @@ functions:
...
-Now we still have a few modifications to do on the application to make it compatible with AWS Lambda.
-
-Since [the filesystem is readonly](/docs/environment/storage.md) except for `/tmp` we need to customize where the cache files are stored. Add this line in `bootstrap/app.php` after `$app = new Illuminate\Foundation\Application`:
-
-```php
-/*
- * Allow overriding the storage path in production using an environment variable.
- */
-$app->useStoragePath($_ENV['APP_STORAGE'] ?? $app->storagePath());
-```
```
mnapoli: Thanks, the PR looks almost ready to merge.

Now I am wondering about this part that ends up being removed: is this intended?

In the PR we still define `APP_STORAGE=/tmp`, but the variable is not used anymore because this modification has been removed: we should probably keep it?

viezel: no you are right. Lets simplify as much as possible. Its removed now.

mnapoli: Sorry for the delay. I was actually expecting the opposite: are we sure that customizing the storage path is not necessary anymore?

viezel: in laravel 6 its $_ENV['APP_BASE_PATH'] ?? dirname(__DIR__). so that can be used.

mnapoli:
Thanks for the PR!

PR #572, opened 2020-02-21
Add --profile option to every command

### LINKED ISSUE #569: aws-profile option for the bref cli command
Got the unauthorized error when I try to pass the assumed role profile to `vendor/bin/bref cli`.  
Saw similar report to the `dashboard` command https://github.com/brefphp/bref/issues/406 
Assuming this is not supported yet.

GromNaN:
Fix #569.

I choose `--profile` as argument name because it is already used in the `bref.dev` command.
Serverless uses `--profile-aws` while aws-cli uses `--profile`.

### DIFF
--- a/bref
+++ b/bref
@@ -90,8 +90,11 @@ $app->command('init', function (SymfonyStyle $io) {
 /**
  * Run a CLI command in the remote environment.
  */
-$app->command('cli function [--region=] [arguments]*', function (string $function, ?string $region, array $arguments, SymfonyStyle $io) {
-    $lambda = new SimpleLambdaClient(($region ?: getenv('AWS_DEFAULT_REGION')) ?: 'us-east-1');
+$app->command('cli function [--region=] [--profile=] [arguments]*', function (string $function, ?string $region, ?string $profile, array $arguments, SymfonyStyle $io) {
+    $lambda = new SimpleLambdaClient(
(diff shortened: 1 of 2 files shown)


bref
```diff
@@ -359,7 +367,7 @@ $app->command('dashboard [--host=] [--port=] [--profile=] [--stage=]', function
     return $process->getExitCode();
 })->descriptions('Start the dashboard');
 
-$app->command('bref.dev [--profile=] [--stage=]', function (string $profile = 'default', string $stage = null, SymfonyStyle $io) {
+$app->command('bref.dev [--profile=] [--stage=]', function (?string $profile, string $stage = null, SymfonyStyle $io) {
```
mnapoli: I am afraid that this change will make the `$profile` variable empty, which may mess up the `serverless info` command built below (the `--aws-profile` is always passed to the command, and now it will not have a value).

Does this still work?

Nyholm:
This is a great feature. Thank you!

bref
GromNaN: Good catch. I've added the usual credential chain (`--profile` then `AWS_PROFILE` then `default`). Otherwise the env var is ignored.

mnapoli: Awesome thanks!

mnapoli:
That's perfect, thanks!

PR #575, opened 2020-02-24
Use AsyncAws instead of AWS SDK

Nyholm:
I have been hesitating of making this PR. I wrote the code a few days ago... anyhow here I am. 

Yes. Im replacing the AWS SDK with a brand new library nobody ever heard of. Except for being cool, async and awesome. There are some real benefits for Bref using Async AWS. 

### Vendor weight

AWS PHP SDK and its dependencies is about 20MB in the vendor folder. Using this PR removes 17MB of code. That is because Async AWS is using a subtree split so you can decide what services you want to download. Bref is only using CloudFormation and Lambda. 

### PSR-7 

One may think that PSR-7 is the worst PSR or you may love it. No matter what you think. Bref does not need to require both Guzzlehttp/psr7 and zendframework/zend-diactoros. That is at least one PSR-7 implementation too much. If we remove AWS SDK we can let the user decide both **if** they want a PSR-7 implementation installed and **which** implementation. 

(Nyholm/psr7 is the best btw). 

--------------

Is this a high risk using Async AWS package? No not really. Your application will run fine since that is using pure `curl` to talk with Lambda. 

The only thing Bref is currently using AWS SDK for is the `./bref deployment` command, `./bref invoke` command and some maintainer tasks (listing published layers). 

With that said; Async AWS is well tested just as the official SDK.  

See https://github.com/async-aws/aws

------------

It is only one TODO in this PR. That is to update to the next stable version of async-aws. That should just be a `composer update` or a bump with version number.

### DIFF
--- a/runtime/layers/layer-list.php
+++ b/runtime/layers/layer-list.php
@@ -6,8 +6,8 @@
  * `layers.json` contains the layer versions that Bref should use.
  */
 
-use Aws\Lambda\LambdaClient;
-use function GuzzleHttp\Promise\unwrap;
+use AsyncAws\Lambda\LambdaClient;
+use AsyncAws\Lambda\Result\LayerVersionsListItem;
 
 require_once __DIR__ . '/../../vendor/autoload.php';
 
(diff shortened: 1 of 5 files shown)


deleugpn:
Might be worth to put on the release not that if people are using the SDK without explicitly requiring it, now it's time to fix that or face a Class not found exception.

mnapoli:
@deleugpn that's a good point ^^ Might as well warn them (even though that isn't a BC break).

Nyholm:
Thank you for merging.

jderusse:
@deleugpn I recommend to periodically run tools like composer-require-checker to explicitly requires used packages, and not relying on dependency's dependencies..

runtime/layers/layer-list.php
```diff
@@ -37,28 +37,29 @@
...
-    // if any of the requests fail
-    $results = unwrap($promises);
+    }
 
     $layers = [];
     foreach ($results as $layerName => $result) {
-        $versions = $result['LayerVersions'];
-        $latestVersion = end($versions);
-        $layers[$layerName] = $latestVersion['Version'];
+        $versions = $result->getLayerVersions(true);
```
jderusse: note `getLayerVersions` is a paginated resource for AWS. The flag `true` disabled the AsyncAws pagination for backward compatibility with previous implementation.

akrabat:
composer update and suddenly the world stopped working! That was unexpected…

Maybe add a note to the 0.5.16 release notes that you need `composer require aws/aws-sdk-php` or `composer require guzzlehttp/guzzle` get back up and running?

Nyholm:
Sure. A release note could maybe be helpful. But this is a sign that you didn’t specify your dependencies properly in your app’s composer.json. 

If you rely on indirect dependencies you may have issues like this. 

Best practice is also to update one dependency at the time and review the output to make sure you know what happens with your application.

mnapoli:
Thanks for the heads up, I forgot to add it to the release notes! That's fixed: https://github.com/brefphp/bref/releases/tag/0.5.16

> Sure. A release note could maybe be helpful. But this is a sign that you didn’t specify your dependencies properly in your app’s composer.json.
> If you rely on indirect dependencies you may have issues like this.

👍

PR #576, opened 2020-02-26
Fix wrong directory name for OpenSSL build

alcaeus:
While it has no impact in the end, the wrong directory irked me while debugging the MongoDB build issue.


mnapoli:
👍  thanks!

PR #577, opened 2020-03-01
Remove decorate_workers_output from php-fpm.conf

### LINKED ISSUE #432: PHP-FPM is adding a prefix warning to stderr

tiagobrito:
Resolves #432 

Remove `decorate_workers_output` configuration from php-fpm.conf when PHP version is 7.2.

PHP 7.2 image build doesn't contain `decorate_workers_output`configuration anymore on `/opt/bref/etc/php-fpm.conf`:

bref

### DIFF
--- a/runtime/layers/fpm/Dockerfile
+++ b/runtime/layers/fpm/Dockerfile
@@ -1,9 +1,12 @@
 ARG PHP_VERSION
 FROM bref/tmp/cleaned-build-php-$PHP_VERSION
+ARG PHP_VERSION
 
 COPY bootstrap /opt/bootstrap
 COPY php.ini /opt/bref/etc/php/conf.d/bref.ini
 COPY php-fpm.conf /opt/bref/etc/php-fpm.conf
+
+RUN if [ "$PHP_VERSION" = "72" ]; then cd /opt/bref/etc/ && sed -i '/decorate_workers_output/d' php-fpm.conf; fi;
 
 # Build the final image from the lambci image that is close to the production environment
 FROM lambci/lambda:provided
--- a/runtime/layers/fpm/php-fpm.conf
+++ b/runtime/layers/fpm/php-fpm.conf
@@ -16,7 +16,7 @@ clear_env = no
 ; Forward stderr of PHP processes to stderr of PHP-FPM (so that it can be sent to cloudwatch)
 catch_workers_output = yes
 ; New PHP 7.3 option that disables a verbose log prefix
-; Disabled for now until we switch to PHP 7.3
+; This option is removed in PHP 7.2 build
 decorate_workers_output = no
 ; Limit the number of core dump logs to 1 to avoid filling up the /tmp disk
 ; See https://github.com/brefphp/bref/issues/275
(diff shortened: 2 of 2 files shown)


mnapoli:
Thanks that's great! Don't we need to set the option in `php-fpm.conf` then? (I don't see that in the diff)

tiagobrito:
True, I forgot to add that file

tiagobrito:
@mnapoli done.

runtime/layers/fpm/php-fpm.conf
```diff
@@ -17,7 +17,7 @@ clear_env = no
 catch_workers_output = yes
 ; New PHP 7.3 option that disables a verbose log prefix
 ; Disabled for now until we switch to PHP 7.3
```
mnapoli: Could you remove that comment? (or update it to signal that this line will be removed on PHP 7.2?)

tiagobrito: sure

runtime/layers/fpm/Dockerfile
```diff
@@ -1,10 +1,13 @@
 ARG PHP_VERSION
 FROM bref/tmp/cleaned-build-php-$PHP_VERSION
+ARG PHP_VERSION
```
hmazter: Duplicate `ARG PHP_VERSION`?

tiagobrito: no, you need to re-declare `PHP_VERSION`. Every `FROM` in your Dockerfile creates a new build stage, which causes values of any ARG from previous stages to be lost.
https://benkyriakou.com/posts/docker-args-empty
I also didn't know this and it took me some time to figure this out.

hmazter: Okay, did not know that 😄

hmazter: Should we include a comment about that and link to that post to not confuse others?

tiagobrito: IMO it's very Docker related, probably even very basic for someone that has more experience with Docker. Also, I think that it's why git/github is for: so people can trackdown changes and check why they were made (e.g.: git blame, history, etc..)

mnapoli: I'm always up for comments in the code but I agree too that it doesn't seem necessary to me here.

mnapoli:
Thank you so much for your patience @tiagobrito, I'm really excited by this pull request. Thanks!

PR #579, opened 2020-03-04
Cloudwatch logs formatted similar to native Runtime

deleugpn:
Looking through CloudWatch logs is a bit cumbersome, specially because every line of the stack makes a new row. This makes it particularly hard to **search** on CloudWatch, since a search result won't include the stack trace because it lost context. Example:

image

This pull request changes the log to be a single JSON row, more in line with how NodeJS, Java & Python does it: 

image

If merged, this is how the PHP logs would look like:

image

### DIFF
--- a/src/Runtime/LambdaRuntime.php
+++ b/src/Runtime/LambdaRuntime.php
@@ -206,13 +206,19 @@ final class LambdaRuntime
             $errorMessage = $error->getMessage();
         }
 
-        // Log the exception in CloudWatch
-        printf(
-            "Fatal error: %s in %s:%d\nStack trace:\n%s",
+        $errorMessage = sprintf(
+            'Fatal error: %s in %s:%d',
             $errorMessage,
             $error->getFile(),
-            $error->getLine(),
-            $error->getTraceAsString()
+            $error->getLine()
+        );
+
+        // Log the exception in CloudWatch
+        printf(
+            '{"errorType": "%s", "errorMessage": "%s", "stack": %s}',
+            get_class($error),
+            $errorMessage,
+            json_encode(explode(PHP_EOL, $error->getTraceAsString()))
         );
 
         // Send an "error" Lambda response


src/Runtime/LambdaRuntime.php
```diff
@@ -206,13 +206,19 @@ private function signalFailure(string $invocationId, \Throwable $error): void
...
+
         // Log the exception in CloudWatch
         printf(
-            "Fatal error: %s in %s:%d\nStack trace:\n%s",
+            '{"errorType": "%s", "errorMessage": "%s", "stack": %s}',
+            get_class($error),
             $errorMessage,
-            $error->getFile(),
-            $error->getLine(),
-            $error->getTraceAsString()
```
mnapoli: Thanks that definitely a good improvement.

We are loosing the file + line number in the error message. Do you know if notices/warnings/errors (i.e. everything except exceptions) include this information in the message? Else I think we should re-add those in the `else {` at the top of this function.

deleugpn: Why do you think we're losing that? I kept the message exactly the same so that nothing would be actually lost/changed per se. The only true change is the format of the message. The `$error->getFile()` and `$error->getLine()` just moved a bit up

mnapoli: I don't understand my comment either 🤦‍♂ 

Sorry ^^

mnapoli:
The tests are failing, I think some of them need to be updated.

Except that I'm 👍  to merge.

deleugpn:
I was waiting your feedback before looking into the tests. I noticed that, but I noticed I'd need more time to pay attention and learn how to fix them. I'll try to do that this this week.

mnapoli:
FYI while working on https://github.com/brefphp/logger/issues/3 I made a few discoveries that I want to port back to Bref. I ended up adding the tests missing for this PR, and incorporating some ideas from https://github.com/brefphp/logger/issues/3.

The new PR (based on this one with your commit) is in #590 -> have a look and let me know what you think!

deleugpn:
Sorry I didn't finish this PR earlier and THANK YOU for the great work you're doing for the PHP community on serverless and all this. I'll be looking into your PR!

PR #582, opened 2020-03-08
Migrate example to AsyncAws

jderusse:
Update Function Invocation PHP example to use AsyncAws for consitency with internal library.

This also Adds a `composer require ` section in the documentation.

Fixes this comment https://github.com/brefphp/bref/pull/575#issuecomment-596228502


mnapoli:
Thanks!

To be honest, I think it might be safer for now to keep the documentation using the official SDK. I prefer to avoid introducing yet another new thing to users that are getting started with serverless/lambda (there is already so much to learn). Maybe we could do a combination of:

- linking to "How to install the PHP AWS SDK" to solve https://github.com/brefphp/bref/pull/575#issuecomment-596228502
- adding a link below the example to the async-aws package as a lighter alternative

WDYT?

jderusse:
Sounds fair.
Doc updated, please have a second look.

mnapoli:
That's perfect, thank you!

PR #586, opened 2020-03-11
Added note about trusted proxies and hosts

Nyholm:
Trusted proxies are for: 
- Getting users IP
- Generating links with HTTPS

Trusted hosts are for: 
- Getting correct host name

This update is probably needed for Laravel too. I leave that some someone else to contribute.

### DIFF
--- a/docs/frameworks/symfony.md
+++ b/docs/frameworks/symfony.md
@@ -80,6 +80,22 @@ Since [the filesystem is readonly](/docs/environment/storage.md) except for `/tm
     }
 ```
 
+An optional last step is to set "Trusted proxy" and "Trusted hosts". Failing to add
+these will cause issues like: 
+ - your application don't generate links using https.
+ - impossible to capture user IP address
+ - failure to recognize host names 
+ 
+That could be done with [environment variables](#Environment variables) or by running 
+these lines in `public/index.php`.
+
+```php
+use Symfony\Component\HttpFoundation\Request;
+
+Request::setTrustedProxies(['127.0.0.1'], Request::HEADER_X_FORWARDED_ALL);
+Request::setTrustedHosts(['api\.example\.com']);
+``` 
+
 ## Deploy
 
 Your application is now ready to be deployed. Follow [the deployment guide](/docs/deploy.md).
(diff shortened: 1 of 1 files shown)


docs/frameworks/symfony.md
```diff
@@ -127,8 +143,26 @@ Since Symfony 4, the production parameters are configured through environment va
...
 The secrets (e.g. database passwords) must however not be committed in this file.
 
 To learn more about all this, read the [environment variables documentation](/docs/environment/variables.md).
+
+## Getting the user's IP
+
+If your application needs the user's IP you must fetch it from the `LAMBDA_CONTEXT` variable.
+Modify `public/index.php` accordingly.
+
+```diff
```
Nyholm: Is `diff` supported?

docs/frameworks/symfony.md
```diff
@@ -127,8 +143,28 @@ Since Symfony 4, the production parameters are configured through environment va
 provider:
     environment:
          APP_ENV: prod
+         TRUSTED_PROXIES: '127.0.0.1'
+         TRUSTED_HOSTS: '^api\.example\.com$'
```
mnapoli: Where are these environment variables used? I can't see them in the diff.

docs/frameworks/symfony.md
mnapoli: I think so yes.

docs/frameworks/symfony.md
Nyholm: They are used in the Symfony default index.php

https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/5.1/public/index.php

docs/frameworks/symfony.md
```diff
@@ -127,8 +143,28 @@ Since Symfony 4, the production parameters are configured through environment va
...
 ```
 
 The secrets (e.g. database passwords) must however not be committed in this file.
 
 To learn more about all this, read the [environment variables documentation](/docs/environment/variables.md).
+
+## Getting the user's IP
+
+If your application needs the user's IP you must fetch it from the `LAMBDA_CONTEXT` variable.
+Modify `public/index.php` accordingly.
```
mnapoli: Would it make sense to forward those values automatically in Bref itself?

I.e. maybe (not sure) set a header with the value that is in the lambda context?

docs/frameworks/symfony.md
mnapoli: Oh I think I understand from the diff above:

> That could be done with [environment variables](#Environment variables) or by running 
these lines in `public/index.php`.

Simply setting those variables is enough? I would say that it's better to document one solution, it is simpler and less confusing. The env var seem quite simple to setup (compared to editing Symfony code), what do you think?

docs/frameworks/symfony.md
Nyholm: Many of these solutions were discussed in https://github.com/brefphp/bref/pull/501

However, I do think it makes sense to cherry-pick useful variables. Ie creating `LAMBDA_CONTEXT_IDENTITY_SOURCEIP`

docs/frameworks/symfony.md
mnapoli: Thanks for the link, turns out we commented at the same time.

What do you think about recommending only the environment variables? (for simplicity)

docs/frameworks/symfony.md
mnapoli: Right, but beyond #501 (which is about the request context containing many things), I'm focusing especially on HTTP-related stuff here: if we have the clients' IP address, we could maybe "fix" the HTTP headers (in the FPM runtime only) so that they contain the correct values out of the box?

mnapoli:
Closing for now (closing all stuck issues/PR from the tracker to better track the ongoing work).

PR #590, opened 2020-03-14
Cloudwatch logs formatted similar to native Runtime, v2

mnapoli:
This PR starts from #579 and adds tests + adds a few things.

## The problem

To reuse what @deleugpn described in #579: current errors in PHP lambdas appear like this in logs (CloudWatch):

This is hard to read, **and** this is not consistent with official runtimes like NodeJS. NodeJS logs an exception like this:

## This PR

With these changes, here is how logs will look like:

Capture d’écran 2020-03-14 à 12 21 53

In short, the error is formatted in JSON. Looking good!

Other changes:

- the log message is simpler: it is the error/exception message only
  - it no longer contains `Fatal error:` (the log line contains `Invoke Error`, so that's less noise)
  - it no longer contains `Uncaught exception:`: we don't want users to think "oh I need to catch exceptions" -> exceptions should not be caught, instead Lambda will handle them
  - it no longer contains the file + line number: the first line of the stack trace already contains that
- the log line now contains the "Request ID" (or invocation ID), because:
  - it is closer to NodeJS's output
  - it allows us to easily find the logs of a single invocation
  - all this stems from https://github.com/brefphp/logger/issues/3
- I updated tests and added more of them (as well as made them more readable)

Bonus: how the execution result looks in `serverless invoke`:

Capture d’écran 2020-03-14 à 12 19 08

How the logs look in `serverless logs` (yuck…):

Capture d’écran 2020-03-14 à 12 19 56

### DIFF
--- a/src/Runtime/LambdaRuntime.php
+++ b/src/Runtime/LambdaRuntime.php
@@ -214,18 +214,17 @@ final class LambdaRuntime
         );
 
         // Log the exception in CloudWatch
-        printf(
-            '{"errorType": "%s", "errorMessage": "%s", "stack": %s}',
-            get_class($error),
-            $errorMessage,
-            json_encode(explode(PHP_EOL, $error->getTraceAsString()))
-        );
+        echo json_encode([
+            'errorType' => get_class($error),
+            'errorMessage' => $errorMessage,
+            'stack' => explode(PHP_EOL, $error->getTraceAsString()),
+        ]);
 
         // Send an "error" Lambda response
         $url = "http://{$this->apiUrl}/2018-06-01/runtime/invocation/$invocationId/error";
         $this->postJson($url, [
-            'errorMessage' => $error->getMessage(),
             'errorType' => get_class($error),
+            'errorMessage' => $error->getMessage(),
             'stackTrace' => explode(PHP_EOL, $error->getTraceAsString()),
         ]);
     }
--- a/src/Runtime/LambdaRuntime.php
+++ b/src/Runtime/LambdaRuntime.php
@@ -206,13 +206,19 @@ final class LambdaRuntime
             $errorMessage = $error->getMessage();
         }
 
-        // Log the exception in CloudWatch
-        printf(
-            "Fatal error: %s in %s:%d\nStack trace:\n%s",
+        $errorMessage = sprintf(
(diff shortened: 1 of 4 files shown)


src/Runtime/LambdaRuntime.php
```diff
@@ -78,7 +78,7 @@ private function closeReturnHandler(): void
     /**
      * Process the next event.
      *
-     * @param Handler|callable $handler If it is a callable, it takes two parameters, an $event parameter (mixed) and a $context parameter (Context) and must return anything serializable to JSON.
+     * @param Handler|RequestHandlerInterface|callable $handler If it is a callable, it takes two parameters, an $event parameter (mixed) and a $context parameter (Context) and must return anything serializable to JSON.
```
mnapoli: Unrelated fix, sorry about the noise!

src/Runtime/LambdaRuntime.php
```diff
@@ -200,27 +200,23 @@ private function sendResponse(string $invocationId, $responseData): void
...
+        ]) . PHP_EOL;
 
         // Send an "error" Lambda response
         $url = "http://{$this->apiUrl}/2018-06-01/runtime/invocation/$invocationId/error";
         $this->postJson($url, [
-            'errorMessage' => $error->getMessage(),
             'errorType' => get_class($error),
-            'stackTrace' => explode(PHP_EOL, $error->getTraceAsString()),
+            'errorMessage' => $error->getMessage(),
+            'stackTrace' => $stackTraceAsArray,
```
mnapoli: Just to be certain to have the same order as Node.

deleugpn:
One last detail of benefit that might be worth mentioning: CloudWatch search makes sense.
Before, I used to struggle (not just with bref but with ECS containers as well) due to PHP stackTrace splitting things into multiple lines.
If you search for an error message, you get a single line in CloudWatch, but dont get the context (previous and next lines), so you lose the stack trace. This is because CloudWatch understands line break as a new entry and php stack is split into lines.
By changing it to an array like this PR does, we can search on CloudWatch for a message and we get a hit that contains the entire json entry, including message + stack.

mnapoli:
Thanks for the feedback! Let's get this going then!

PR #593, opened 2020-03-16
[PHP-FPM configuration] Increase log_limit

Nyholm:
This will allow you to specify any PHP-fpm values. 

Draft: I'll will try this out first..

### DIFF
--- a/runtime/layers/fpm-dev/php-fpm.conf
+++ b/runtime/layers/fpm-dev/php-fpm.conf
@@ -6,8 +6,6 @@ pid = /tmp/php-fpm.pid
 ; TODO: report that to the PHP bug tracker
 ;log_level = 'warning'
 
-include=/var/task/php/fpm/conf.d/*.conf
-
 [default]
 pm = static
 ; We only need one child because a lambda can process only one request at a time
--- a/runtime/layers/fpm-dev/php-fpm.conf
+++ b/runtime/layers/fpm-dev/php-fpm.conf
@@ -6,6 +6,8 @@ pid = /tmp/php-fpm.pid
 ; TODO: report that to the PHP bug tracker
 ;log_level = 'warning'
 
+include=/var/task/php/fpm/conf.d/*.conf
+
 [default]
 pm = static
 ; We only need one child because a lambda can process only one request at a time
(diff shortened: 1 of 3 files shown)


mnapoli:
Should we set [`log_limit`](https://www.php.net/manual/en/install.fpm.configuration.php) to a very high value? (maybe you opened this as WIP to test only?)

Nyholm:
This is tested and working. 

I deployed the following images: 

* `arn:aws:lambda:eu-central-1:403367587399:layer:php-74-fpm:1`
* `arn:aws:lambda:eu-west-1:403367587399:layer:php-74-fpm:2`
* `arn:aws:lambda:us-east-1:403367587399:layer:php-74-fpm:2`


I added the file in my app: `php/fpm/conf.d/fpm.conf`:
```
[global]

log_limit = 4096

```

Before: 
Screenshot 2020-03-16 at 16 06 56

After: (content is scrollable to the right)
Screenshot 2020-03-16 at 16 25 23

Nyholm:
(completing the PR now)

Nyholm:
Im done with the PR. 

Big thank you to @jenschude and @markomitranic for helping tracking down this bug.

Nyholm:
Oh, wait..

>  Unable to globalize '/var/task/php/fpm/conf.d/*.conf' (ret=2) from /opt/bref/etc/php-fpm.conf at line 13.

If there is no `'/var/task/php/fpm/conf.d/` everything will fail :/

Nyholm:
Not sure how to work around that.

Nyholm:
Im done with the PR now. This increases the log_limit to 8 times default. 

I leave the "custom PHP-fpm" config for another PR.

jenschude:
> Oh, wait..
> 
> > Unable to globalize '/var/task/php/fpm/conf.d/*.conf' (ret=2) from /opt/bref/etc/php-fpm.conf at line 13.
> 
> If there is no `'/var/task/php/fpm/conf.d/` everything will fail :/

Theoretically we could add additonally to the `opt` folders also a `/var/task/fpm/conf.d` as the layers themselves are just zip files which are just extracted in order of definition.

markomitranic:
Not sure i get it, this is a common practice in containers, we either run mkdir -p for conf.d that always ensures that the directory is there, or even better and more common, always perform a copy step that copies any provided ini files from a specific folder. Not sure in the context of bref what makes more sense.

Nyholm:
I think this PR is done. But I would like to discuss the "allow edit of php-fpm config" in https://github.com/brefphp/bref/issues/594

Nyholm:
Thank you for the merge.

PR #597, opened 2020-03-20
Suggest to exclude var directory when deploying Symfony applications

pyrech:
IMO Symfony's `var` directory (containing mainly cache and log) should not be deployed to AWS and could be excluded by default. Do you agree?

It also avoids potential deploy error when archive is too big due to the cache being included :smile:

### DIFF
--- a/docs/frameworks/symfony.md
+++ b/docs/frameworks/symfony.md
@@ -36,6 +36,7 @@ package:
     exclude:
         - node_modules/**
         - tests/**
+        - var/**
 
 functions:
     website:


pyrech:
Hum, as cache pre-generation is the solution suggested in https://bref.sh/docs/environment/storage.html#pre-generating-the-cache, I'm not sure what to do here:

- don't change anything in the doc
- keep the PR as is (var directory not deployed)
- suggest to warmup the cache locally before deploy instead

mnapoli:
Yes, that's the tricky part! (and thank you for the PR by the way)

We could document the 2 approaches. I don't like giving too much choice (as this is passing the burden of choice onto the user), but I consider both approaches valid:

- don't deploy var, move the cache to the `/tmp` directory -> longer cold starts, but much simpler to get started, or for smaller applications
- precompile the cache and deploy it -> faster, harder

(by the way, in any case we never want to deploy all the other directories like `var/logs`, etc.)

Nemo64:
I have deployed symfony applications with pre warmed cache but it was a lot more tricky then just deploying the var directory because symfony will want write to it which at best results into warnings and at worst resultet into blank responses and timeouts.

I implemented copying into the cache from the var folder into the `/tmp/` folder which improved cold start from 3 seconds to about 1.2 seconds. But it is definitly more fragile and i don't know if i'd want to recommend it to someone as "the" solution.

```php
class Kernel extends BaseKernel
{
    // [...]

    public function getLogDir()
    {
        // When on the lambda only /tmp is writeable
        if (getenv('LAMBDA_TASK_ROOT') !== false) {
            return '/tmp/log/' . $this->environment;
        }

        return parent::getLogDir();
    }

    public function getCacheDir()
    {
        // When on the lambda only /tmp is writeable
        if (getenv('LAMBDA_TASK_ROOT') !== false) {
            return '/tmp/cache/' . $this->environment;
        }

        return parent::getCacheDir();
    }

    public function boot()
    {
        // When on the lambda, copy the cache dir over to /tmp where it is writable
        if (getenv('LAMBDA_TASK_ROOT') !== false && !file_exists($this->getCacheDir())) {
            $this->prepareCacheDir(parent::getCacheDir(), $this->getCacheDir());
        }

        return parent::boot();
    }

    private function prepareCacheDir(string $staticCacheDir, string $tempCacheDir): void
    {
        $filesystem = new Filesystem();
        $filesystem->mkdir($tempCacheDir);

        foreach (scandir("$staticCacheDir", SCANDIR_SORT_NONE) as $item) {
            if (in_array($item, ['.', '..'])) {
                continue;
            }

            // the pools folder needs to be writable so mirror it
            if ($item === 'pools') {
                $filesystem->mirror("$staticCacheDir/$item", "$tempCacheDir/$item");
                continue;
            }

            // symlink all folders other folders
            // this is especially important with the Container* folder since it uses require_once statements
            if (is_dir("$staticCacheDir/$item")) {
                $filesystem->symlink("$staticCacheDir/$item", "$tempCacheDir/$item");
                continue;
            }

            // and copy all other files, i had intermediate problems when linking them
            $filesystem->copy("$staticCacheDir/$item", "$tempCacheDir/$item");
        }
    }
```

mnapoli:
@Nemo64 that looks really good!

Yep, the `pools` folder shouldn't be set at all to `var` as read-only because it's supposed to be written at runtime. I assumed that nobody would use it since it meant the cache wouldn't be shared across lambda instances -> Redis or DynamoDB would be better. But I guess it's a bit ambitious to expect everyone to setup Redis immediately (especially since it needs a VPC, which is complex).

All that to say: maybe we should make "your way" the offical way for Symfony?

What we could do is create a `bref/symfony-bridge` package that provides, for example, a trait:

```php
trait BrefKernelExtension
{
    public function getLogDir()
    {
        // When on the lambda only /tmp is writeable
        if (getenv('LAMBDA_TASK_ROOT') !== false) {
            return '/tmp/log/' . $this->environment;
        }

        return parent::getLogDir();
    }

    public function getCacheDir()
    {
        // When on the lambda only /tmp is writeable
        if (getenv('LAMBDA_TASK_ROOT') !== false) {
            return '/tmp/cache/' . $this->environment;
        }

        return parent::getCacheDir();
    }

    public function boot()
    {
        // When on the lambda, copy the cache dir over to /tmp where it is writable
        if (getenv('LAMBDA_TASK_ROOT') !== false && !file_exists($this->getCacheDir())) {
            $this->prepareCacheDir(parent::getCacheDir(), $this->getCacheDir());
        }

        return parent::boot();
    }

    private function prepareCacheDir(string $staticCacheDir, string $tempCacheDir): void
    {
        ...
    }
}
```

Symfony users would `composer require bref/symfony-bridge` and add the trait to their Kernel:

```php
class Kernel extends BaseKernel
{
    use BrefKernelExtension;

    // [...]
}
```

If you think it can be implemented differently (a compiler extension or whatever) that could work as well.

The main thing is that users wouldn't have to copy that code manually. And the logic would be:

- if `var/` is deployed, we aim to use it as much as possible (for shorter cold starts)
- if `var/` is **not** deployed, we set `/tmp/cache` as the cache directory

That way it would work for everyone, and it would work out of the box. The documentation could mention the easy way (`exclude` the `var/` directory) and explain that you can also prewarm it and deploy it if you want better performances.

WDYT?

pyrech:
I really like this quick win and the idea of a bridge to avoid users to have to paste some code.

However, before going that way, maybe could we ask a Symfony expert to validate that using a pregenerated cache is safe?

mnapoli:
Right, maybe @Nyholm can answer that question?

Nemo64:
@mnapoli The trait isn't that beautiful because it references a `parent` that isn't defined in the trait. And if you need the `boot` method in your kernel, you'll accidentally override the cache copy method.

And also it is _really_ fragile. If it fails you get a white site or a timeout with no log message whatsoever because logging isn't set up yet.

I'm not against putting it into a library (actually i'd really like that) but it needs to be much more sophisticated if it's supposed to be the recommendation.
1. It should probably be a different Kernel class entirely, just to remove some gotchas with a trait
2. It should include some error handling during `boot` because you really don't get any feedback if you do something wrong. If the LAMBDA variable is defined, then it can probably just write to `php://stderr`
3. You must check if the cache is complete... I don't know how to do it but if the cache folder isn't properly warmed, then you get a white page or a timeout.
4. You should check if there is just 1 kernel class and log if there are multiple, which happens if you don't remove the cache folder before warming it.

If we are good, we can make the symfony library a serverless plugin as well which clears and warms the cache before the serverless deploy, to avoid incomplete cache problems...

Nyholm:
Hey. I like the idea/concept with the trait. 
@Nemo64 bring up some really great suggestions. Number 1 and 2 is something we should definitely do. That will make it easier to version the `bref/symfony-bridge` package and define its dependencies. 

### Addressing @Nemo64 number 3: 

I dont thing that is correct. If the cache folder isn't warm then the request will spend time warming it. (Which is fine)

However, if the cache folder is not writable, then you may get the white page as you describe. 

### Addressing @Nemo64 number 4: 

I think we avoid that, right?

- `BrefKernel` extends `Symfony\Component\HttpKernel\Kernel`
- `AppKernel` extends `BrefKernel`
- `AppAdminKernel` extends `BrefKernel` but implements it's own `getCacheDir()`

So since we are extending Symfony's `Kernel` we dont need to care about an application with multiple Kernels. 

----------------------

> Is using a pregenerated cache is safe?

Yes, before each deploy you should warm the cache. Where the cache folder is located does not matter. 
What Im not sure about is the exact logic in `prepareCacheDir()`. Ie, what is fine to copy, what is fine to symlink etc etc. But that is something we just need to figure out. 

Another question: How much time will it take to just copy the full folder compared to the logic shown in this PR? That should be nearly constant operation on the lambda filesystem, right? (or at least if it is "copy on write" like PHP =))

Nemo64:
@Nyholm 
point 4: I think I used the wrong term, there should only be 1 container in the cache, symfony does not override generated containers but only adds new containers while marking existing containers with a `.legacy` file. That will unnecessarily increase the deployment package and should be avoided if possible. So I'd love some log output that hints to this issue.

And for copying the entire cache folder: You can copy everything except the `ContainerXXXXX` folders which must be symlinked because there are require statements in them that relatively link into the `vendor` folder. Except if you warm the container up with other path parameters but that would mean building a cache which only works after you have deployed it. I haven't tried building the container `inline_class_loader` disabled though, maybe thats an option. Or cleverly symlinking the vendor folder into `tmp` might work too.

On the other hand: nothing needs to be writable except the `pools` folder when the cache is fully warmed. But It's probably safer to make everything writable for the incomplete-cache-case.

mnapoli:
All that sounds good.

> If we are good, we can make the symfony library a serverless plugin as well which clears and warms the cache before the serverless deploy, to avoid incomplete cache problems...

I'd rather avoid adding too much complexity and leave it to users to either:

- deploy without cache (simpler)
- pregenerate the cache and deploy (better performances)

Should I create a new project on GitHub?

Nyholm:
I agree. 

Yes, please.

mnapoli:
Great, I created https://github.com/brefphp/symfony-bridge and invited you both. I won't have time to contribute today but I'll have time tomorrow, feel free to send pull requests.

mnapoli:
For onboarding new users I believe this fixes inconsistencies with the current state of the documentation, I'll be merging this for now.

Thanks!

PR #600, opened 2020-03-25
Fix #599 Set up the same "max upload size" in PHP as API Gateway/Lambda

### LINKED ISSUE #599: Set up the same "max upload size" in PHP as API Gateway/Lambda
When using the Docker images for development, PHP is configured with max upload sizes of 2Mb as well as `client_max_body_size` of 1Mb.

We should change those settings to match the API Gateway limit, which is 6Mb.

mnapoli:
- I added functional tests against a real Lambda + API Gateway -> can't get more reliable tests than that
- I changed PHP's POST max size and upload size to match Lambda (6Mb)
- I changed Nginx's configuration as well so that we get the same limit in the development Docker image (previously it was 1Mb)

### DIFF
--- a/tests/Functional/FpmRuntimeTest.php
+++ b/tests/Functional/FpmRuntimeTest.php
@@ -201,6 +201,7 @@ class FpmRuntimeTest extends TestCase
 
     /**
      * The API Gateway limit is 10Mb, but Lambda is 6Mb.
+     *
      * @see https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html
      * We check with 4Mb because this works. 5Mb fails, maybe because the whole size of the event
      * is larger (because of the whole JSON formatting plus headers?).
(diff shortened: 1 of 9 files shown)


tests/Functional/FpmRuntimeTest.php
```diff
@@ -0,0 +1,244 @@
...
+{
+    /** @var Client */
+    private $http;
+
+    public function setUp(): void
+    {
+        parent::setUp();
+
+        $this->http = new Client([
+            'base_uri' => 'https://5octfcz6gc.execute-api.eu-west-1.amazonaws.com/dev/',
```
acairns: Should we be worried about exposing this url? Could it change in time or can it get hit by a bot? Wondering if this perhaps could be an env var instead?

mnapoli: It could, but I added some throttling to avoid being too exposed. I have set up billing alarms on my account, and I run these tests when creating new releases, so I am not too worried for now. TBH I think it's fine for now, I have those URLs around everywhere for demos, examples, etc.

PR #603, opened 2020-03-29
add reference to the rds-data api (resolves #596)

### LINKED ISSUE #596: document using rds-data to connect to a database
I build a doctrine driver to use the rds-data api to connect to an aurora serverless database.
https://github.com/Nemo64/dbal-rds-data

It removes the need for VPC configuration.
I dislike advertising my own stuff but I think it saves a lot of time and makes the setup much simpler.
So my question is if you'd like me to documented in your [Accessing the database from PHP](https://bref.sh/docs/environment/database.html#accessing-the-database-from-php) documentation.

Nemo64:
I added a reference to the aurora rds-data api to the documentation because it is a generally good option to avoid a vpc in your serverless setup.

I also added a reference to my [dbal-rds-data](https://github.com/Nemo64/dbal-rds-data) api which makes the usage fairly seemless because i haven't found another already ready made solution on packagist.

And at last i updated the database list a bit because it was kind of missleading by telling that it is a complete list of services offered by rds. I added a reference that aurora serverless also supports postgresql and made it clear that those are just some of the offered services.

### DIFF
--- a/docs/environment/database.md
+++ b/docs/environment/database.md
@@ -6,13 +6,12 @@ introduction: Configure Bref to use a database in your PHP application on AWS La
 
 AWS offers the [RDS](https://aws.amazon.com/rds/) service to run MySQL and PostgreSQL databases.
 
-Here are the database services offered by RDS:
+Here are some of the database services offered by RDS:
 
 - MySQL
 - PostgreSQL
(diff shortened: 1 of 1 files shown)


mnapoli:
Thanks!

Issue #602, opened 2020-03-26
Support for HTTP API


aknosis:
HTTP APIs are now generally available: https://aws.amazon.com/blogs/compute/building-better-apis-http-apis-now-generally-available/

I think we should implement HTTP APIs as an option during initialization and eventually migrate toward making them the default.

Reasons to migrate to HTTP APIs:

**Faster**
> For the majority of use cases, HTTP APIs offers up to 60% reduction in latency

**Cheaper**
> Using the pricing for us-east-1, the image shows a cost comparison for 100 million, 500 million, and 1 billion requests a month. Overall, HTTP APIs is at least 71% lower cost compared to API Gateway REST APIs.

![cheaper graph](https://d2908q01vomqb2.cloudfront.net/1b6453892473a467d07372d45eb05abc2031647a/2020/03/12/Picture3.png)

**$default Stage**

HTTP APIs allow for a $default stage, this stage will point to the root domain (e.g. https://abc123.execute-api.us-east-2.amazonaws.com), this gets rid of all the issues relating to the state prefix.

The changes required are very simple (assuming default serverless.yml setup).

REST API:
```yml
    events:
      - http: 'ANY /'
      - http: 'ANY /{proxy+}'
```

HTTP API:

```yml
    events:
      - httpApi:
          path: '*'
```

Anecdotally, I have been running a bref app on an HTTP API for about a month now with no issues.

mnapoli:
Hi, I completely agree with you. I have been using it as well in a side project and it's great.

The blog article does mention a difference in the payload though (https://aws.amazon.com/fr/blogs/compute/building-better-apis-http-apis-now-generally-available/), I think we need to add some tests and review if there is any impact to Bref before adding it in the documentation.

aknosis:
I didn't notice the payload change since serverless defaults to the 1.0 version (
https://github.com/serverless/serverless/blob/master/lib/plugins/aws/package/compile/events/httpApi/index.js#L427). 

Also not configurable to the 2.0 version yet (https://github.com/serverless/serverless/issues/7477), so once that is possible it would make sense to implement. 

It also appears the response format may have to differ when using 2.0 format: https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html#http-api-develop-integrations-lambda.proxy-format

mnapoli:
Oh that's interesting!

So if I recap to make sure we are on the same page:

[...]

PR #604, opened 2020-03-29
Support for HTTP API #602

victormacko:
I've added in support for HTTP v2.0 payloads primarily to have Bref give a "works out of the box" experience when linking Lambda with the new HTTP API via the AWS Console, rather than having to switch back to 1.0 in the integrations section in the console.

Also - i'm pretty keen to start using the new API Gateway HTTP option so hopefully this helps getting Bref one more step closer to it :) ... this should also help with issue #602 

I'd started looking at the tests to add but was wanting some guidance on it first -- I think the best approach is to add another FpmHandlerTest class for the v2 payloads as most of the test-cases are probably going to be worth testing...

The changes i've made cover;
* Adding a 'payloadVersion' in the request & response classes (the response classes needed it as the multi-header option doesn't exist in v2)
* Handling cookies in the FpmHandler & HttpRequestEvent classes (they're passed in separately and not as a HTTP Header as far as I can tell)
* Updated how the method is obtained (var names have changed)
* Updated how the path is obtained (var names have changed)
* Updated how the query-string is obtained (var names have changed) (the multiValueQueryStringParameters no longer exists either in v2)

The unit-tests all pass, but they haven't as yet been updated to test for the v2 payloads -- need some guidance (as above) on this before spending time on getting that sorted.


src/Event/Http/HttpRequestEvent.php
```diff
@@ -20,15 +20,28 @@ final class HttpRequestEvent implements LambdaEvent
...
     private $queryString;
+    /** @var float */
+    private $payloadVersion;
 
     public function __construct(array $event)
     {
-        if (! is_array($event) || ! isset($event['httpMethod'])) {
+        if (! is_array($event)) {
+            throw new InvalidLambdaEvent('API Gateway or ALB', $event);
+        }
```
aknosis: I don't think this is necessary - the type hint on the constructor should prevent this from ever occurring.

victormacko: Good call - fixed :)

src/Event/Http/HttpResponse.php
```diff
@@ -16,11 +16,15 @@ final class HttpResponse
     /** @var string */
     private $body;
 
-    public function __construct(string $body, array $headers = [], int $statusCode = 200)
+    /** @var float|null */
+    private $payloadVersion;
+
+    public function __construct(string $body, array $headers = [], int $statusCode = 200, ?float $payloadVersion = 1.0)
```
mnapoli, line 22: I think we should avoid passing `$payloadVersion` in the constructor: this class represents a HTTP response, and that's independent from the version of API Gateway.

However, when we convert the response to API Gateway -> that's where we need to know. I suggest to pass `$payloadVersion` as a parameter to `toApiGatewayFormat()` instead (just like `$multiHeaders`).

That should also simplify the places where this constructor is called. Since this constructor can also be called by Bref users now, it's a good thing as well to remove the parameter from the constructor (they shouldn't care about this parameter in their code).

src/Event/Http/HttpRequestEvent.php
```diff
@@ -94,13 +103,17 @@ public function getServerName(): string
 
     public function getPath(): string
     {
+        if (isset($this->event['rawPath'])) {
+            return $this->event['rawPath'];
+        }
```
mnapoli, line 108: `rawPath` seems to exist in v2 only.

`path` exists in v1 only.

I think we should make that clearer, even if it's not as optimum as possible, because it will make maintenance easier. Something like this:

```
if ($this->payloadVersion >= 2) {
    return $this->event['rawPath'] ?? '/';
} else {
    return $this->event['path'] ?? '/';
}
```

src/Event/Http/HttpRequestEvent.php
```diff
@@ -20,15 +20,24 @@ final class HttpRequestEvent implements LambdaEvent
     private $headers;
     /** @var string */
     private $queryString;
+    /** @var float */
+    private $payloadVersion;
 
     public function __construct(array $event)
     {
-        if (! is_array($event) || ! isset($event['httpMethod'])) {
+        $this->payloadVersion = isset($event['version']) ? (float) $event['version'] : null;
```
mnapoli, line 28: Ideally `$this->payloadVersion` should not be nullable. That would simplify code below.

Isn't the `version` always present?

One thing to consider: at this point, we are not sure if `$event` is a HTTP event`. What we could do is:

- first, check that it's a valid HTTP event (that's the lines below) -> if not, an exception is thrown
- then, we are sure it's a HTTP event, so we are sure we have `version` -> we can do `$this->payloadVersion = (float) $event['version']`

src/Event/Http/HttpRequestEvent.php
```diff
@@ -125,25 +138,40 @@ public function getRequestContext(): array
...
+    public function getPayloadVersion(): ?float
+    {
+        return $this->payloadVersion;
+    }
+
     private function rebuildQueryString(): string
     {
+        if (isset($this->event['rawQueryString'])) {
+            return $this->event['rawQueryString'];
+        }
```
mnapoli, line 173: Same as with `rawPath` -> we should differentiate entirely between v1 and v2 events. That will ease readability

```suggestion
        if ($this->payloadVersion >= 2) {
            return $this->event['rawQueryString'];
        }
```

src/Event/Http/HttpRequestEvent.php
```diff
@@ -125,25 +138,40 @@ public function getRequestContext(): array
 
     public function getCookies(): array
     {
-        if (! isset($this->headers['cookie'])) {
-            return [];
-        }
+        $cookies = [];
+        if ($this->getPayloadVersion() >= 2 && isset($this->event['cookies'])) {
```
mnapoli, line 142: Here is a suggestion:

```suggestion
        if ($this->payloadVersion >= 2) {
```

Reasons:

- avoid calling the getter internally, very small perf optimization (I don't like doing those, but in these classes in Bref this makes sense)
- if we are in v2, we should not go in the `v1` code branch

mnapoli:
Regarding the tests, here is what I noted:

- `Bref\Test\Event\Http\HttpResponseTest`: unit test, should be easy to update
  You can duplicate existing tests and name them with `v1` and `v2` in the test name to differentiate. If some tests are identicals (the behavior is identical between v1 and v2), then don't duplicate the test.
- `Bref\Test\Event\Http\CommonHttpTest`: another unit test for the requests, this one is a bit more complex because it's an abstract class.
  This class uses JSON examples for the tests. What you can do is duplicate all JSON files to have them in v1 and v2 formats.

I have pushed a new commit to your pull request to start adding tests in `CommonHttpTest`. Hopefully that should help you move forward. If you still don't understand, let me know and I'll try to find time to complete them.

src/Event/Http/HttpRequestEvent.php
victormacko: Cool, just making those changes now.

src/Event/Http/HttpRequestEvent.php
victormacko: I'll relocate the seting-part of the code to after the check RE ALB/API events so in theory it should always exist.

victormacko:
Ace, thanks so much for your help on those tests -- i'll have a proper look through it once i've made the other code changes and got them committed.

mnapoli:
Awesome thanks!

src/Event/Http/HttpHandler.php
```diff
@@ -21,6 +21,6 @@ public function handle($event, Context $context): array
 
         $response = $this->handleRequest($httpEvent, $context);
 
-        return $response->toApiGatewayFormat($httpEvent->hasMultiHeader());
+        return $response->toApiGatewayFormat($httpEvent->hasMultiHeader(), (isset($event['version']) ? (float) $event['version'] : 1.0));
```
mnapoli, line 24: This could probably be (I haven't tested):

```suggestion
        return $response->toApiGatewayFormat($httpEvent->hasMultiHeader(), $httpEvent->getPayloadVersion());
```

mnapoli: Actually I'm thinking: if v2 doesn't support multiheaders at all, then `$httpEvent->hasMultiHeader()` should return `false` in that case, and we don't even need to know the payload version here?

That could simplify things even more.

src/Event/Http/HttpRequestEvent.php
```diff
@@ -20,15 +20,23 @@ final class HttpRequestEvent implements LambdaEvent
...
+    private $payloadVersion;
 
     public function __construct(array $event)
     {
-        if (! is_array($event) || ! isset($event['httpMethod'])) {
+        // version 1.0 of the HTTP payload
+        if (isset($event['httpMethod'])) {
+            $this->method = strtoupper($event['httpMethod']);
+        } elseif (isset($event['requestContext']['http']) && isset($event['requestContext']['http']['method'])) {
+            // version 2.0 - https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html#http-api-develop-integrations-lambda.proxy-format
```
mnapoli, line 32: That URL is super useful, that's the first time I see it.

Could you add it to the comment of the class as well? That way it's very visible.

victormacko: Nps, done :)

src/Event/Http/HttpHandler.php
victormacko: Good catch - now updated

victormacko: > Actually I'm thinking: if v2 doesn't support multiheaders at all, then `$httpEvent->hasMultiHeader()` should return `false` in that case, and we don't even need to know the payload version here?
> 
> That could simplify things even more.

For sure -- have fixed that up now too :)

victormacko:
RE tests @mnapoli , the Bref\Test\Handler\FpmHandlerTest doesn't have the 'version' set on any of the data it's testing ... do you have a preferred approach ... default to v1 in the HttpRequestEvent or just add in the 'version' to each of the tests? (not sure if you wanted to keep the tests small or not).

I'll finish tackling the v2 JSON files tomorrow -- i'm almost done :)

mnapoli:
Regarding `FpmHandlerTest` yes keep v1, that's definitely sufficient given how much is covered elsewhere. I'll have a look at the rest ASAP, that's awesome.

mnapoli:
At first look that sounds really good. The build seems to be failing, but we are definitely on the right track.

victormacko:
Ace!
I've updated the FPM tests so there's only 4 overall failing (just 2 issues really);
* 2 for the multi-headers (AWS concatenates multi-headers a single string, separated by a ',' (not url-encoding ... so commas inside user-agent headers, etc may get handled incorrectly if we were to try and separate them out just using the ',')
* 2 for multi-query-string (eg. a=abc&a=def) ... the tests expect the rawQueryString to be url-encoded ... and using parse_str & http_build_query together don't handle int-indexed arrays as well as the unit-tests expect :/

Hope that helps -- not quite sure on what approach to take for those two.

mnapoli:
OK @victormacko I think we are good here!

mnapoli:
Thank you so much for all that work, that is massive. Time to try that!

PR #605, opened 2020-03-29
Updated tests

victormacko:
This is more of a helper to assist with future updates which are going to be needed when you update to a later version of PHPUnit (the latest version has changed the base-classes, so the signatures on some of the methods need to be updated)

I'd needed to update them anyway when running the PHPUnit tests locally (before realising the vendor/bin/phpunit exec was the correct command!)


mnapoli:
Thank you for helping with the future upgrades!

PR #606, opened 2020-03-29
try to make the cdn configuration simpler

Nemo64:
The serverless-pseudo-parameters plugin is not really needed and the trade of of a slightly less readable syntax is definitely worth the 1 fewer dependency and not needing to having npm.
This was annoying when I started experimenting with bref and aws a few months ago since I already had webpack and frontend dependencies in my `package.json`. It wasn't sensible to completely avoid plugins for me but the initial setup is much simpler like this.

Then I also removed the slightly inconsistent definition of the assets target. It talked and configured the source partially for an s3 website source but didn't use an s3 website endpoint in the end. It now is just a plain s3origin with a reference to the manual. There are a few advantages to an s3origin like correctly sending 404 error when something isn't there.

### DIFF
--- a/docs/websites.md
+++ b/docs/websites.md
@@ -182,8 +182,6 @@ resources:
                         -   Id: Assets
                             DomainName: !GetAtt Assets.RegionalDomainName
                             S3OriginConfig: {} # this key is required to tell CloudFront that this is an S3 origin, even though nothing is configured
-                            # you can also use a CustomOrigin but the S3Origin creates correct 404 error and has other optimizations as well
-                            # https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/DownloadDistS3AndCustomOrigins.html#concept_S3Origin
                     # The default behavior is to send everything to AWS Lambda
                     DefaultCacheBehavior:
                         AllowedMethods: [GET, HEAD, OPTIONS, PUT, POST, PATCH, DELETE]
--- a/docs/websites.md
+++ b/docs/websites.md
@@ -153,10 +153,7 @@ resources:
     Resources:
         # The S3 bucket that stores the assets
         Assets:
-            Type: AWS::S3::Bucket
-            Properties:
-                BucketName: <bucket-name>
-                AccessControl: PublicRead # needed for the cdn to read the directory structure
+            # [...] see the previous section for details 
         AssetsBucketPolicy:
             # [...] see the previous section for details 
     
--- a/docs/websites.md
+++ b/docs/websites.md
@@ -179,6 +179,13 @@ resources:
                         -   Id: Assets
                             DomainName: !GetAtt Assets.RegionalDomainName
                             S3OriginConfig: {} # this key is required to tell CloudFront that this is an S3 origin, even though nothing is configured
+                            # If you host a static website, like a SPA, use s3-website URLs instead of the config above
+                            # See https://stackoverflow.com/questions/15309113/amazon-cloudfront-doesnt-respect-my-s3-website-buckets-index-html-rules#15528757
+                            # DomainName: !Join ['', [!Ref Assets, '.s3-website-', !Ref AWS::Region, '.amazonaws.com']]
+                            # CustomOriginConfig:
+                            #     OriginProtocolPolicy: 'http-only' # S3 websites only support HTTP
+                            # You'll also need to enable website hosting on your s3 bucket by configuring the WebsiteConfiguration property
+                            # https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket.html#cfn-s3-bucket-websiteconfiguration
                     # The default behavior is to send everything to AWS Lambda
                     DefaultCacheBehavior:
                         AllowedMethods: [GET, HEAD, OPTIONS, PUT, POST, PATCH, DELETE]
--- a/docs/websites.md
+++ b/docs/websites.md
@@ -67,7 +67,8 @@ resources:
                         -   Effect: Allow
                             Principal: '*' # everyone
                             Action: 's3:GetObject' # to read
-                            Resource: 'arn:aws:s3:::<bucket-name>/*' # things in the bucket
+                            Resource: !Join ['/', [!GetAtt Assets.Arn, '*']] # things in the bucket
+                            # alternatively you can write out Resource: 'arn:aws:s3:::<bucket-name>/*'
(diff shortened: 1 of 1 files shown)


docs/websites.md
```diff
@@ -155,14 +148,15 @@ functions:
...
 
 resources:
     Resources:
         # The S3 bucket that stores the assets
         Assets:
-            # [...] see the previous section for details 
+            Type: AWS::S3::Bucket
+            Properties:
+                BucketName: <bucket-name>
+                AccessControl: PublicRead # needed for the cdn to read the directory structure
```
mnapoli, line 159: Could you explain why we need to add this?

Do you mean that this is required specifically by the `S3OriginConfig` integration with CloudFront? I am surprised we don't need it otherwise.

docs/websites.md
```diff
@@ -179,17 +173,17 @@ resources:
...
                         # The assets (S3)
                         -   Id: Assets
-                            # Use s3-website URLs instead if you host a static website (https://stackoverflow.com/questions/15309113/amazon-cloudfront-doesnt-respect-my-s3-website-buckets-index-html-rules#15528757)
-                            DomainName: '#{Assets}.s3.amazonaws.com'
-                            CustomOriginConfig:
-                                OriginProtocolPolicy: 'http-only' # S3 websites only support HTTP
+                            DomainName: !GetAtt Assets.RegionalDomainName
+                            S3OriginConfig: {} # this key is required to tell CloudFront that this is an S3 origin, even though nothing is configured
+                            # you can also use a CustomOrigin but the S3Origin creates correct 404 error and has other optimizations as well
+                            # https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/DownloadDistS3AndCustomOrigins.html#concept_S3Origin
```
mnapoli, line 186: Are there any advantages to using `CustomOrigin` instead?

From what you are writing here, it seems there are only downsides. If there are only downsides, then we can remove the last 2 lines of comment. We shouldn't suggest it at all, it's simpler for everyone.

Nemo64: I honestly don't know of any advantages of using a CustomOrigin like it was described before.
The only reason (i know of) is that you can use it with an s3 bucket that has website hosting enabled in which case you can configure redirects within the s3 bucket.

I haven't used CustomOrigin for long after reading the docs and switching because it looks like the intended way.
There is also special pricing for traffic from s3 to CloudFront... which is free, i don't know if that works when using a CustomOrigin.

docs/websites.md
Nemo64: I was confused too. But without it, the cdn only delivers access denied. But I also can't remove the `AssetsBucketPolicy` because then i also get access denied when accessing resources.

I can investigate further if you want, the official way is to create an s3 origin identity so that only the cdn can access files. I tried that once but it took an irritatingly long time before everything worked.

docs/websites.md
Nemo64: But i agree, i remove the comment.
Guess I tried to explain why i changed that within the code instead of within this pull request.
And if you need the other inclusion way, then you probably already know what you are doing.

docs/websites.md
Nemo64: Interesting,

I investigated further and noticed that it makes absolutely no difference.
I'm a bit confused why i needed it before... but yes it works.

I found this template from aws: https://github.com/awslabs/aws-cloudformation-templates/blob/5c66bbfaec08313fcfee48b49ce6ba0a38f6bb1a/community/solutions/StaticS3CloudFront.yml
It uses an s3 origin identity to restrict access to the bucket to the cdn but there is also no ACL configuration.

So i'll remove it. Good that you asked ;)

Nemo64:
I made some changes after your input.

Do you prefer multiple commits or should I amend them together?

docs/websites.md
mnapoli: 👍  good!

One  last thing I can think of: is it possible to host static websites with `S3OriginConfig`? By that I mean: I want `/` to load `index.html`.

If not, then we should probably keep this bit commented:

```yaml
DomainName: !GetAtt Assets.RegionalDomainName
S3OriginConfig: {} # this key is required to tell CloudFront that this is an S3 origin, even though nothing is configured
# If you host a static website, like a SPA, use s3-website URLs instead of the config above
# See https://stackoverflow.com/questions/15309113/amazon-cloudfront-doesnt-respect-my-s3-website-buckets-index-html-rules#15528757
# DomainName: '<bucket-name>.s3-website-us-east-1.amazonaws.com'
# CustomOriginConfig:
#     OriginProtocolPolicy: 'http-only' # S3 websites only support HTTP
```

mnapoli:
@Nemo64 I'm fine with multiple commits, no worries.

docs/websites.md
Nemo64: You can tell CloudFront to deliver a [DefaultRootObject](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-distribution-distributionconfig.html#cfn-cloudfront-distribution-distributionconfig-defaultrootobject), although it only works with the actual root of the distribution.

If you want to host a static website and use php just for an api, i'd definitely recommend you use a website origin... although I don't know if it is the best idea to put php behind CloudFront in that case since that's already a compromise to get everything under one domain. An api can live under a separate/generated domain and use the ApiGateway's caching capabilities.

I can still add the documentation about the s3 website hosting though because you can have redirects etc with them which normal s3origins can't.

Although I'd probably document it like this:
```yaml
DomainName: !GetAtt Assets.WebsiteURL
CustomOriginConfig:
    OriginProtocolPolicy: 'http-only' # S3 websites only support HTTP
```

mnapoli: :+1: that is perfect! If you can add that in (commented) that's great.

mnapoli: Ping me once that is done (I'm not sure I get notified of new commits) and we can merge that.

Nemo64: @mnapoli I did a bit of research with s3 website hosting. You'll also need to enable it. I roughly documented it but i don't guarantee it works like that.

Nemo64:
I think i'm fine. If I notice errors I just open another pull request ;)

docs/websites.md
mnapoli: Yep, that's fine by me, we are giving pointers here, that's not the main use case with Bref.

mnapoli:
Thank you!

PR #610, opened 2020-03-31
Add details about setting SSM entries on Windows

rbairwell:
I found setting the SSM details on a Windows 10 machine using Powershell to be "non-standard" and I hope this small change will save people time in the future.


mnapoli:
Thanks!

PR #611, opened 2020-04-03
Homepage Type Updates

aknosis:
Found a typo on the home page:

> They sponsors the project

In this case _sponsors_ should just be _sponsor_.

Went ahead and reviewed all the text and made a few more updates. 

Serverless Framework: In all cases I could find on their sites framework in is capitalized.

Minor change to change "on AWS Lambda" to "to AWS Lambda". This sentence is implying that bref is bringing PHP support _to_ Lambda.


mnapoli:
Thank you so much! I feel very bad with that typo in the title ^^ But at least it's fixed now.

PR #614, opened 2020-04-09
Bump Version of AsyncAWS

swiffer:
Keep up with AsyncAWS until they are releasing 1.0

### DIFF
--- a/bref
+++ b/bref
@@ -3,8 +3,8 @@
 declare(strict_types=1);
 
 use AsyncAws\CloudFormation\CloudFormationClient;
-use AsyncAws\CloudFormation\Result\Output;
-use AsyncAws\CloudFormation\Result\StackEvent;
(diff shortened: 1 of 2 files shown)


mnapoli:
Thank you. Is there anything we need to change in the codebase? In notice the build is failing now.

Nyholm:
The largest BC changes are: 
- We made many value objects final
- We moved classes used by both requests and responses to a separate namespace. (this is why PHPStan complains)

@swiffer I think the only change you need to do is to update a `use` statement. 

Sorry for all the updates. Currently (as always) I feel like there is no more BC breaks ahead of us.

mnapoli:
@Nyholm thanks for the review! I have absolutely no problem with bc breaks by the way (since this is an internal lib).

mnapoli:
Thank you!

Issue #339, opened 2019-05-28
Run a console command as a cron


steefaan:
First of all thanks for your great work on bref and thanks for sharing the code. Everything is working smoothly. I made it working to have a symfony console command deployed to AWS Lambda I can execute. 

To execute my command I need to run `vendor/bin/bref --region=eu-central-1 cli symfony-console -- app:update-inventory-quantity`. This works great but I'm struggling a lot to automate this command to run it every 15 minutes. After some googling I figured out that CloudWatch provides what I'm searching for but I stuck on the configuration part. I'm even not sure if my question belongs to CloudWatch (so out of bref's scope) or bref itself. Due to my "specific" use case (lambda, php, bref, symfony) I couldn't find anything on Google what answers my question satisfying.

The simple question is how to pass parameters to CloudWatch? Although I'm able to configure CloudWatch to execute `symfony-console` itself I need a possibility to pass the command I want to execute (`app:update-inventory-quantity`).

That's my `template.yaml`:

```yml
AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31

Globals:
    Function:
        Environment:
            Variables:
                APP_ENV: prod

Resources:
    Console:
        Type: AWS::Serverless::Function
        Properties:
            FunctionName: 'symfony-console'
            CodeUri: .
            Handler: bin/console
            Timeout: 120 # in seconds
            Runtime: provided
            Layers:
                - 'arn:aws:lambda:eu-central-1:209497400698:layer:php-73:6' # PHP
                - 'arn:aws:lambda:eu-central-1:209497400698:layer:console:6' # The "console" layer

```

deleugpn:
Here's what you can try, I'm not sure if it's going to work as I have no experience with Symfony and this solution is something you made me think right now.

```
AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31

Globals:
    Function:
        Environment:
            Variables:
                APP_ENV: prod

Resources:
    Console:
        Type: AWS::Serverless::Function
        Properties:
            FunctionName: 'symfony-console'
            CodeUri: .
            Handler: bin/console
            Timeout: 120 # in seconds
            Runtime: provided
            Layers:
                - 'arn:aws:lambda:eu-central-1:209497400698:layer:php-73:6' # PHP
                - 'arn:aws:lambda:eu-central-1:209497400698:layer:console:6' # The "console" layer
            Events:
[...]

PR #615, opened 2020-04-09
Simplify the console invocation format (use a simple string)

mnapoli:
From https://github.com/brefphp/bref/issues/339#issuecomment-611445756

The `'{"cli": "doctrine:migrate --force"}'` JSON structure used to exist for Bref 0.2 that used to mix event types for the same handler. Now, we don't really need this anymore.

We could switch the `console` runtime "event" to a simple string. That would make cron much easier to define, for example:

```yaml
        events:
            - schedule:
                rate: rate(1 minute)
                input: 'doctrine:migrate --force'
```

That could even been invokable easily via `serverless invoke`:

```bash
serverless invoke -f console -d 'doctrine:migrate --force'
```

For now I have kept backward compatibility with the current event array format.

### DIFF
--- a/bref
+++ b/bref
@@ -105,9 +105,7 @@ $app->command('cli function [--region=] [--profile=] [arguments]*', function (st
     }, $arguments);
 
     try {
-        $result = $lambda->invoke($function, json_encode([
-            'cli' => implode(' ', $arguments),
-        ]));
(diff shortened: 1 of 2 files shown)


patrickbussmann:
Very small change with big effects 😎
Thank you!

PR #616, opened 2020-04-10
Fixes typo

xavismeh:
This path does not correspond to any existing value, forcing the use of the default value

### DIFF
--- a/docs/websites.md
+++ b/docs/websites.md
@@ -172,7 +172,7 @@ resources:
                         -   Id: Website
                             DomainName: !Join ['.', [!Ref ApiGatewayRestApi, 'execute-api', !Ref AWS::Region, 'amazonaws.com']]
                             # This is the stage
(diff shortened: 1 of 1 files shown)


mnapoli:
Hi, thanks!

The thing is that the stage can be defined in 2 places:

- as a CLI option
- in `serverless.yml`

A common pattern is to use `${opt:stage, self:provider.stage}`, which might be a better option. However can you check that the variable ends up being `dev` if the stage is not defined anywhere?

xavismeh:
Oh ok. Wouldn't it be a bit too much to have `${opt:stage, ${self:provider.stage}, 'dev'}`? I haven't tried it, but that would work I guess ...

mnapoli:
Yep, this is why I opened https://github.com/serverless/serverless/pull/6049 ^^ I should probably try to pick that up again.

mnapoli:
Closing for now.

(🧹  issue tracker grooming)

PR #618, opened 2020-04-12
Fix missing dependency for PSR-15 handlers

mnapoli:
Since we don't require the AWS SDK anymore, Guzzle is not installed as a transitive dependency. The PSR-15 handler mechanism doesn't work anymore.

The root of the issue was that we did not explicitly require Guzzle's dependency originally, so we missed this. Anyway, I ended up installing @Nyholm's library because it's great (fast, lightweight).

I tried setting up https://github.com/maglnet/ComposerRequireChecker but there are false positives and I gave up after spending some time on it. Anyone is welcome to pick that up :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,8 @@
         "riverline/multipart-parser": "^2.0.6",
         "psr/http-server-handler": "^1.0",
         "async-aws/cloud-formation": "^0.4",
-        "async-aws/lambda": "^0.4"
+        "async-aws/lambda": "^0.4",
+        "nyholm/psr7": "^1.2"
     },
     "require-dev": {
         "phpunit/phpunit": "^6.5",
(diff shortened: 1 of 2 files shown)


mnapoli:
@Nyholm it seems there is a slight difference with your implementation: the body stream is not rewinded when we create the `ServerRequest`.

It's easy to adjust the tests (to rewind the body before getting its content), but I'm thinking that it's not practical for users: `$this->request->getBody()->getContents()` will return an empty string.

Is that intended? What do you think?

Nyholm:
Yes. It is intended. The psr7 does not specify the state of the string or if it should reminded or not. We do not rewinded it because it it not 100% sure you need to read the body. 

Psr7 uses the __toString function in the specification. Generally, one should always use that function over getContents(). Because it does not assume the state of the string.

mnapoli:
Thank you for the response!

I have added a fix so that users can get the body of the request with `->getContents()`.

PR #620, opened 2020-04-13
Solving php source extraction problem

gerardo-junior:
the "amazonlinux" base image does not come with xz package installed by default and is required for extracting compressed code, and it also adds hash checking after downloading.


gerardo-junior:
@mnapoli I had a little time at the end of the work and decided to look your code that we discussed on Twitter, I hope this will fix ;)

mnapoli:
Thank you!

I have a few questions:

- you mentioned that `xz` is not available on Amazon Linux, this is why you install it: however the original code downloaded the `.tar.gz` archive, and it ran fine when compiling the docker image on my machine -> is there any reason to suspect that this was an issue?
- the original issue was "Failed to connect to www.php.net port 443: Connection timed out", is there any reason that with this PR the result will be different?

Here is an example of a failed build btw: https://github.com/brefphp/bref/runs/583187471?check_suite_focus=true

```
 ---> Running in c334ffc2682e
+ mkdir -p /tmp/build/php
+ curl --location --silent --show-error --fail https://www.php.net/get/php-7.4.4.tar.gz/from/this/mirror
+ tar xzC /tmp/build/php --strip-components=1
curl: (7) Failed to connect to www.php.net port 443: Connection timed out

gzip: stdin: unexpected end of file
tar: Child returned status 1
tar: Error is not recoverable: exiting now
The command '/bin/bash -c set -xe;     mkdir -p ${PHP_BUILD_DIR};     curl --location --silent --show-error --fail https://www.php.net/get/php-${VERSION_PHP}.tar.gz/from/this/mirror   | tar xzC ${PHP_BUILD_DIR} --strip-components=1' returned a non-zero code: 2
```

gerardo-junior:
I'm sorry, in my tests I ended up leaving the .xz version, fixed! I did some tests through secure.php.net (which he had suggested on twitter) it redirects to the main mirror and in all my tests neither timeout (in my machine and github action)

https://imgur.com/a/pNXnvFl

mnapoli:
I gave a try running `curl -v https://secure.php.net/get/php-7.4.4.tar.gz/from/this/mirror` inside GitHub Actions and it still fails, here is the complete output: https://gist.github.com/mnapoli/4ded165680221f7ba81c617bcf32b0cd

I don't understand why we have this problem here. You said it works in GitHub Actions, how did you test?

PR #623, opened 2020-04-16
Revise Lambda/Request context within PHP-FPM handler

eddmann:
Currently we have access to the Request context within the PHP-FPM handler request thanks to the inclusion of the `LAMBDA_CONTEXT` env. var present per request.
This is great but sadly does not provide exposure to the true Lambda Context which is supplied per Lambda invocation, only the request context present in the event.

I feel that this naming of this env. var is a little misleading and as such this PR deprecates the current usage of this `LAMBDA_CONTEXT` env. var.
In its place it provides two new env. vars. `LAMBDA_INVOCATION_CONTEXT` and `LAMBDA_REQUEST_CONTEXT` (thanks @mnapoli for the suggestions!) which provide access to the true Lambda Context and Event Request Context.


src/Event/Http/FpmHandler.php
```diff
@@ -179,7 +179,8 @@ private function eventToFastCgiRequest(HttpRequestEvent $event): ProvidesRequest
         $request->setServerPort($event->getServerPort());
         $request->setCustomVar('PATH_INFO', $event->getPath());
         $request->setCustomVar('QUERY_STRING', $event->getQueryString());
-        $request->setCustomVar('LAMBDA_CONTEXT', json_encode($event->getRequestContext()));
+        $request->setCustomVar('REQUEST_CONTEXT', json_encode($event->getRequestContext()));
+        $request->setCustomVar('LAMBDA_CONTEXT', json_encode($context));
```
mnapoli, line 183: Here is the BC break. Your suggested naming makes perfect sense, but we need to find an alternate naming.

Either we keep `LAMBDA_CONTEXT` and find another name. Or we deprecate `LAMBDA_CONTEXT` and move to a more consistent naming, for example:

- `AWS_LAMBDA_CONTEXT`
- `REQUEST_CONTEXT`

or another cheap trick like this?

src/Context/Context.php
```diff
@@ -63,4 +63,14 @@ public function getTraceId(): string
     {
         return $this->traceId;
     }
+
+    public function jsonSerialize()
```
mnapoli, line 67:
```suggestion
    public function jsonSerialize(): array
```

marcusirgens:
You should probably change the name of the context attribute in the PSR 15 handler as well. See #543

src/Event/Http/FpmHandler.php
eddmann: I would be happy with either approach, for ease deprecating `LAMBDA_CONTEXT` and replacing with `AWS_LAMBDA_CONTEXT` makes a lot of sense.

eddmann:
@marcusirgens I think in that context the naming seems to align with the intended values. What would you propose the name change to be?

marcusirgens:
@eddmann The PSR7 ServerRequest should have the same attributes as the env var, only kebab-cased, so if you're adding `AWS_LAMBDA_CONTEXT`, be sure to add it to the PSR7 object as `aws-lambda-context` as well! :)

mnapoli:
In the PSR-7 request the names look pretty good:

Capture d’écran 2020-04-16 à 16 26 59

Additionally, we pass the whole `$event`, which is different from just the "request context".

But I agree though that it's weird that we would have a `AWS_` prefix in one place and not the other 🤔 

argh… naming is hard.

eddmann:
@mnapoli @marcusirgens Haha so true about naming 🤦, we could prefix all with `AWS_` and `aws-`?

mnapoli:
That's a possibility. Just had a thought, how about:

- `LAMBDA_CONTEXT`
- `LAMBDA_REQUEST_CONTEXT`
- `REQUEST_CONTEXT`: Deprecated
- `lambda-event`
- `lambda-context`

That seems consistent, and I would even say that's an improvement over `REQUEST_CONTEXT` which is a really generic name. `REQUEST_CONTEXT` is actually a format really specific to the Lambda HTTP request, so `LAMBDA_REQUEST_CONTEXT` makes more sense.

WDYT?

eddmann:
@mnapoli That sounds great, so we would re-purpose `LAMBDA_CONTEXT` to be the true lambda context as opposed to the current request context?

mnapoli:
🤦 

I read the diff wrong and suggested something completely dumb: the problem is with the `LAMBDA_CONTEXT` variable, not `REQUEST_CONTEXT` (what I suggested does not solve the BC break). You know, when I wrote my comment I was surprised that the answer was so obvious… Now I understand why 😄 

Back to square one!

Here is another attempt:

- Deprecated: `LAMBDA_CONTEXT`
- `LAMBDA_INVOCATION_CONTEXT`
- `LAMBDA_REQUEST_CONTEXT`
- `lambda-event`
- `lambda-context`

Yes, there is a slight inconsistency here, but it's probably something we can solve over time by re-adding `LAMBDA_CONTEXT` in a few months/years?

eddmann:
@mnapoli I think being more explicit makes a lot of sense in regards to the `LAMBDA_INVOCATION_CONTEXT` naming. So I'll revise the PR to keep `LAMBDA_CONTEXT` as is (wondering how best to document it's deprecation?), and add `LAMBDA_INVOCATION_CONTEXT` and `LAMBDA_REQUEST_CONTEXT` 👍

src/Event/Http/FpmHandler.php
```diff
@@ -179,6 +179,8 @@ private function eventToFastCgiRequest(HttpRequestEvent $event): ProvidesRequest
         $request->setServerPort($event->getServerPort());
         $request->setCustomVar('PATH_INFO', $event->getPath());
         $request->setCustomVar('QUERY_STRING', $event->getQueryString());
+        $request->setCustomVar('LAMBDA_INVOCATION_CONTEXT', json_encode($context));
+        $request->setCustomVar('LAMBDA_REQUEST_CONTEXT', json_encode($event->getRequestContext()));
         $request->setCustomVar('LAMBDA_CONTEXT', json_encode($event->getRequestContext()));
```
eddmann, line 184: @mnapoli Unsure what the best way to mark this as deprecated would be?

mnapoli: I don't know what's best either ^^

Here might be a start, and I'll add that to the release notes as well.

```suggestion
        /** @deprecated The LAMBDA_CONTEXT variable is deprecated in favor  of LAMBDA_REQUEST_CONTEXT */
        $request->setCustomVar('LAMBDA_CONTEXT', json_encode($event->getRequestContext()));
```

mnapoli:
I just realized that this feature is documented, could you update the docs? (https://bref.sh/docs/runtimes/http.html#request-context)

eddmann:
@mnapoli I've revised the context documentation based on the changes in this PR, would be great to know what you think :)

PR #626, opened 2020-04-17
Expose Lambda invocation context within console runtime

eddmann:
This allows us to access the Lambda context within console runtime invocations, in a similar fashion to how we can within PHP-FPM usage.

**Note:** This depends on the `JsonSerializable` context changes present in https://github.com/brefphp/bref/pull/623

### DIFF
--- a/runtime/layers/console/bootstrap
+++ b/runtime/layers/console/bootstrap
@@ -1,6 +1,7 @@
 #!/opt/bin/php
 <?php declare(strict_types=1);
 
+use Bref\Context\Context;
 use Bref\Runtime\LambdaRuntime;
 
 ini_set('display_errors', '1');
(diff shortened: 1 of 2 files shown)


runtime/layers/console/bootstrap
```diff
@@ -24,11 +25,16 @@ if (! is_file($handlerFile)) {
...
 
 while (true) {
-    $lambdaRuntime->processNextEvent(function ($event) use ($handlerFile): array {
+    $lambdaRuntime->processNextEvent(function ($event, Context $context) use ($handlerFile): array {
         $cliOptions = $event['cli'] ?? '';
 
         // We redirect stderr to stdout so that all the output is collected
-        $command = "/opt/bin/php $handlerFile " . $cliOptions . ' 2>&1';
+        $command = sprintf(
+            'LAMBDA_INVOCATION_CONTEXT=%s /opt/bin/php %s %s 2>&1',
```
eddmann, line 33: The only fear here is if the environment variable value is very large, but as the JSON encoded context value length is somewhat predictable we should be safe.

An example value looks like:
`{"awsRequestId":"46383f07-e8f9-4f71-b9e1-61279b075bf7","deadlineMs":1587162325806,"invokedFunctionArn":"arn:aws:lambda:us-east-1:262xxxxxx831:function:app-dev-api","traceId":"Root=1-5e9a2cb9-3cf013386c1f89109c927fd0;Parent=3a7eaeb164e48444;Sampled=0"}`
With a letter count of 251

PR #627, opened 2020-04-18
Improve the Laravel documentation (config and Laravel Passport)

mnapoli:
I updated the Laravel documentation (including instructions on how to handle Laravel Passport).

### DIFF
--- a/docs/frameworks/laravel.md
+++ b/docs/frameworks/laravel.md
@@ -116,41 +116,32 @@ For more details follow [the "Console" guide](/docs/runtimes/console.md).
 
 ## Laravel Passport
 
-Laravel Passport has a `passport:install` command. However, this command cannot be run in Lambda because it needs to write files to the `storage` directory.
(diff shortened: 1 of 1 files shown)


docs/frameworks/laravel.md
```diff
@@ -58,9 +59,9 @@ We will also need to customize the location for compiled views, as well as custo
 ```dotenv
 VIEW_COMPILED_PATH=/tmp/storage/framework/views
 
-# We cannot store sessions to disk: if you don't need sessions (e.g. API)
-# then use `array`, else store sessions in database or cookies
-SESSION_DRIVER=array
+# We cannot store sessions to disk: if you don't need sessions (e.g. API) then use `array`
+# If you write a website, use `cookie` or store sessions in database.
+SESSION_DRIVER=cookie
```
mnapoli, line 64: APIs don't use the session, but sessions for websites do not work with "array". At least now it works out of the box for both scenarios.

PR #634, opened 2020-04-27
New behaviour for config caches in Laravel 7.x

abronin:
I'm using Laravel 7.8.1.

If you don't run `artisan config:cache` you will not have `packages.php` and `services.php` inside `bootstrap/cache`.

Without these files Laravel will try to create them and this will cause the error 'The bootstrap/cache directory must be present and writable'.

However, we still need to run `artisan config:clear` to delete `bootstrap/cache/config.php` that contains wrong paths.


mnapoli:
Thank you! I haven't tried v7.8 so I'll trust you on this ^^

PR #637, opened 2020-04-27
Remove the timeout on `bref cli`

### LINKED ISSUE #629: The `bref cli` command seems to have a timeout
I upgraded Bref in a project that uses `vendor/bin/bref cli`.

One of the command that is run takes a long time, and now I get a timeout when the lambda is invoked via the HTTP API:

```
Fatal error: Uncaught Symfony\Component\HttpClient\Exception\TransportException: Idle timeout reached for "https://lambda.*********.amazonaws.com/2015-03-31/functions/app-preprod-console/invocations". in /srv/build/vendor/async-aws/core/src/Response.php on line 106

Call Stack:
    0.0008     490536   1. {main}() /srv/build/vendor/bref/bref/bref:0
    0.0404    6422864   2. Silly\Application->run() /srv/build/vendor/bref/bref/bref:439

AsyncAws\Core\Exception\Http\NetworkException: Could not contact remote server. in /srv/build/vendor/async-aws/core/src/Response.php on line 106

Call Stack:
    0.0008     490536   1. {main}() /srv/build/vendor/bref/bref/bref:0
    0.0404    6422864   2. Silly\Application->run() /srv/build/vendor/bref/bref/bref:439
  120.4705    7793104   3. AsyncAws\Core\Response->__destruct() /srv/build/vendor/async-aws/core/src/Response.php:0
  120.4705    7793104   4. AsyncAws\Core\Response->resolve() /srv/build/vendor/async-aws/core/src/Response.php:73
  120.4708    7810624   5. AsyncAws\Core\Response->getResolveStatus() /srv/build/vendor/async-aws/core/src/Response.php:109
```

From what I read in the code, [async-aws](https://github.com/async-aws/aws) doesn't have a timeout by default to fetch the response. However, here we have a `Idle timeout` which may be a different kind of timeout.

@Nyholm or @jderusse, would you know if there is such a timeout (if so, how to customize it) or if I am looking in the wrong place? I had a look at the docs but couldn't find that info. Thank you in advance :)

mnapoli:
Fix #629

### DIFF
--- a/demo/console.php
+++ b/demo/console.php
@@ -20,5 +20,8 @@ $silly->command('error', function (OutputInterface $output) {
     $output->writeln('There was an error!');
     return 1;
 });
+$silly->command('sleep', function () {
+    sleep(120);
+});
 
 $silly->run();
--- a/bref
+++ b/bref
@@ -95,7 +95,8 @@ $app->command('init', function (SymfonyStyle $io) {
 $app->command('cli function [--region=] [--profile=] [arguments]*', function (string $function, ?string $region, ?string $profile, array $arguments, SymfonyStyle $io) {
     $lambda = new SimpleLambdaClient(
         $region ?: getenv('AWS_DEFAULT_REGION') ?: 'us-east-1',
-        $profile ?: getenv('AWS_PROFILE') ?: 'default'
+        $profile ?: getenv('AWS_PROFILE') ?: 'default',
+        15 * 60 // maximum duration on Lambda
     );
 
     // Because arguments may contain spaces, and are going to be executed remotely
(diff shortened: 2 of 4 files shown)


demo/console.php
```diff
@@ -20,5 +20,8 @@
     $output->writeln('There was an error!');
     return 1;
 });
+$silly->command('sleep', function () {
```
Nyholm, line 23: Why is this needed?

mnapoli: To test what I implemented.

Those files are my "manual" test files ;)

mnapoli:
> Btw, did you test with both curl and socket?

I didn't, Bref requires curl, should I worry about that?

Nyholm:
>I didn't, Bref requires curl, should I worry about that?

No. It was more my curiosity since I didnt find the timeout setting is the source code for CurlClient the other day. (https://github.com/brefphp/bref/issues/629#issuecomment-618047148)

PR #638, opened 2020-04-28
fix for alb double encoding

matthewsuan:
This fixes same issue addressed in #538 but scoped out to ALB requests only.

### DIFF
--- a/src/Event/Http/HttpRequestEvent.php
+++ b/src/Event/Http/HttpRequestEvent.php
@@ -20,6 +20,8 @@ final class HttpRequestEvent implements LambdaEvent
     private $headers;
     /** @var string */
     private $queryString;
+    /** @var bool */
+    private $isAlb;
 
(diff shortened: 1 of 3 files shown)


deleugpn:
While working on #693 I wanted to get inspiration on the other PRs open. This one seems to partially address the issue, however, it requires multi value to be disabled on the load balancer, which leads to loss of multi value parameters such as `param[]=1&param[]=2`. 
I'm trying to come up with a solution that works for both enabled and disabled multi value.

mnapoli:
@deleugpn I'm not sure, did #693 solve this issue too?

deleugpn:
Yes, I think #693 covers the use case here, but without limiting it to just a single parameter. When you enable Multi Value on your ALB, you'll get all parameters properly decoded.

mnapoli:
Awesome thanks!

PR #640, opened 2020-05-04
Add Kinesis typed events

i906:
Added Kinesis typed events.

The data from the Kinesis stream is represented as a Base64 string. The `$record->getData()` decodes this and converted to PHP array. The raw string can also be retrieved using `$record->getRawData()`.


i906:
So far it always have the same JSON structure. I haven't tested enough yet to notice if the structure does changes.

mnapoli:
Thank you!

PR #643, opened 2020-05-05
Add Cape Town Region

richardnbanks:
AWS recently added the af-south-1 region in South Africa and we would like to use Bref in our local region.

I have created a PR for the runtimes page as well: brefphp/runtimes.bref.sh#4

### DIFF
--- a/runtime/layers/regions.json
+++ b/runtime/layers/regions.json
@@ -14,5 +14,6 @@
     "ap-northeast-1",
     "ap-northeast-2",
     "ap-southeast-1",
-    "ap-southeast-2"
+    "ap-southeast-2",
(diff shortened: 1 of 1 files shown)


mnapoli:
Thanks, that's perfect.

Could you just confirm that AWS Lambda is indeed available in that region? I'm just asking because some regions do not support all services, and I got bit by this before.

richardnbanks:
Yes, lambdas and layers can be created in the region. The only real issue I found was that the region needed to be enabled manually on our AWS accounts.

mnapoli:
Oh maybe I will need to enable it as well.

Thank you!

PR #644, opened 2020-05-05
added new region milan

AlessandroMinoccheri:
Added new region milan

### DIFF
--- a/runtime/layers/regions.json
+++ b/runtime/layers/regions.json
@@ -14,5 +14,6 @@
     "ap-northeast-1",
     "ap-northeast-2",
     "ap-southeast-1",
-    "ap-southeast-2"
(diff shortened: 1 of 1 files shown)


runtime/layers/regions.json
```diff
@@ -14,5 +14,6 @@
     "ap-northeast-1",
     "ap-northeast-2",
     "ap-southeast-1",
-    "ap-southeast-2"
+    "ap-southeast-2",
+    "eu-south-1"
```
mnapoli, line 18: Could you add that line below the other `eu-` lines above?

AlessandroMinoccheri:
Yes, AWS Lambda is available in that region.
Sure I can create a new PR.

mnapoli:
Thanks

Issue #639, opened 2020-04-28
Timeouts in lamda code leading to weird issue


cosmok:
Hi,

Thanks to the team for your amazing work!

I have used bref in a few projects and I am happy with it.

In one of the new projects that I am involved in, I received reports about timeouts persisting across requests. In other words, a timeout in Request A is sometimes leading to an 'internal server error' message response for another subsequent Request B.

Following is a setup that can reproduce this issue:
serverless.yml
```
service: demo-timeout

provider:
    name: aws
    stage: staging
    logs:
      restApi: true
    region: ap-southeast-2
    runtime: provided
    vpc: 'vpc-id'
    endpointType: 'regional' 

plugins:
    - ./vendor/bref/bref

functions:
    api:
        handler: index.php
        memorySize: 1024
        description: ''
        timeout: 5 # in seconds (API Gateway has a timeout of 29 seconds)
        layers:
            - ${bref:layer.php-72-fpm}
        events:
           -   http: 'GET /'
``` 

index.php
```
<?php

require __DIR__ . '/vendor/autoload.php';

/**
 * a simple script that concatenates
 * two passed params 'a' and 'b'
 * and optionally simulates a socket timeout
 * based off a passed param 'timeout'
 */
$a = ($_GET['a']) ?? 'Hello';
$b = ($_GET['b']) ?? 'World!';

error_log('inside function');
if (isset($_GET['timeout']) && ctype_digit($_GET['timeout'])) {
    $timeout = (int) $_GET['timeout'];
    error_log("timeout for: {$timeout} seconds");
    //non-routable ip
    fsockopen('10.255.255.1', 443, $error_no, $error_string, $timeout);
}
$message = "{$a} {$b}";
header("Content-Type: application/json");
echo json_encode([
    'message' => $message,
]);
```
composer.json
```
{
    "require": {
        "bref/bref": "dev-master#404c69f2f"
    }
}
```
Using latest bref - tried the latest master just in case as the current release has the issue.
No other packages or extensions involved.

Following siege command will reveal the issue:
```
siege -d1 -c1 -v  -t30s -p -f urls.txt 
[error] CONFIG conflict: selected time and repetition based testing
defaulting to time-based testing: 30 seconds
{"message": "Internal server error"}
{"message":"Achyuth Kesavan"}
{"message": "Internal server error"}
{"message": "Internal server error"}
{"message": "Internal server error"}
{"message": "Internal server error"}
{"message": "Internal server error"}
[...]

PR #645, opened 2020-05-05
Don't kill self

cosmok:
When a previous request times out and doesn't get a chance to clean up,
any subsequent request can many times fail.

This is because before starting a php-fpm process we are trying to kill
any existing ones. As process ids get reused between requests, if the old request has the same PID as the current one, we would be killing self.

Fixes: #639 

When this is detected my change will just do the socket and PID file
cleanups, but, leave the process alone.


mnapoli:
Wow I am impressed, that bug is nasty! Thank you for identifying it!

I intend to merge this with a bunch of other changes a little bit later, I just wanted to let you know that this looks good.

mnapoli:
Thank you!

PR #649, opened 2020-05-10
added support for running `bref/dashboard` within `bref/dev-env`

meridius:
The `HOST_AWS_CONF_DIR` ENV variable is set in brefphp/dev-environment#2

The motivation behind this is to be able to run Bref Dashboard from inside of `bref/dev-env` container so users won't have to install Serverless, PHP, NPM, etc. locally.


mnapoli:
Closing this for now, the "bref dev env" is kinda stuck and I focused on other things lately.

PR #658, opened 2020-05-23
Make sure one can access user's IP easily

Nyholm:
To access the users IP currently, one need to: 

```
if (isset($_SERVER['LAMBDA_REQUEST_CONTEXT'])) {
    $context = json_decode($_SERVER['LAMBDA_REQUEST_CONTEXT'], true);
    $userIp = $context['identity']['sourceIp'] ?? '';
}
```

To avoid decoding a json string to an array, I suggest to put the user IP in a separate variable. 

Pros: 
- It is way simpler
- The user IP is a common value that I would like to have in my logs. 

Cons: 
- We take up "space" in the global `$_SERVER`. But it is just a small string. So I guess it is fine.

### DIFF
--- a/src/Event/Http/FpmHandler.php
+++ b/src/Event/Http/FpmHandler.php
@@ -182,6 +182,11 @@ final class FpmHandler extends HttpHandler
         $request->setCustomVar('LAMBDA_INVOCATION_CONTEXT', json_encode($context));
         $request->setCustomVar('LAMBDA_REQUEST_CONTEXT', json_encode($event->getRequestContext()));
 
+        if (isset($event->toArray()['identity']['sourceIp'])) {
+            // Add user's IP as an easy to access variable
+            $request->setCustomVar('LAMBDA_IDENTITY_SOURCE_IP', $event->toArray()['identity']['sourceIp']);
+        }
+
         /** @deprecated The LAMBDA_CONTEXT has been renamed to LAMBDA_REQUEST_CONTEXT for clarity */
         $request->setCustomVar('LAMBDA_CONTEXT', json_encode($event->getRequestContext()));
 


src/Event/Http/FpmHandler.php
```diff
@@ -182,6 +182,11 @@ private function eventToFastCgiRequest(HttpRequestEvent $event, Context $context
         $request->setCustomVar('LAMBDA_INVOCATION_CONTEXT', json_encode($context));
         $request->setCustomVar('LAMBDA_REQUEST_CONTEXT', json_encode($event->getRequestContext()));
 
+        if (isset($event->toArray()['identity']['sourceIp'])) {
+            // Add user's IP as an easy to access variable
+            $request->setCustomVar('LAMBDA_IDENTITY_SOURCE_IP', $event->toArray()['identity']['sourceIp']);
+        }
```
mnapoli, line 188: Shouldn't we put that IP address directly into `$request->setRemoteAddress(...);` above?

Nyholm: No, I dont think we should. 

RemoteAddress is the last proxy. We are behind at least 2 proxies and we should not confuse things more that needed. 

Yes, it is super confusing to properly get the user IP in PHP. Here is how I will write in my index.php after this PR is merged. 


```
<?php

use App\Kernel;
use Symfony\Component\ErrorHandler\Debug;
use Symfony\Component\HttpFoundation\Request;

require dirname(__DIR__).'/config/bootstrap.php';

// Mark 127.0.0.1 as a trusted proxy
Request::setTrustedProxies(['127.0.0.1']), Request::HEADER_X_FORWARDED_ALL ^ Request::HEADER_X_FORWARDED_HOST);

// It is only trusted on this host
Request::setTrustedHosts(['api.tnyholm.se']);

// Get user IP:
$_SERVER['HTTP_X_FORWARDED_FOR'] = $_SERVER['LAMBDA_IDENTITY_SOURCE_IP'] ?? '';

$request = Request::createFromGlobals();
$response = $kernel->handle($request);
$response->send();
$kernel->terminate($request, $response);
```

This will tell Symfony to trust localhost and the `HTTP_X_FORWARDED_FOR` header. 
I can now run `$request->getClientIp()` to get the correct IP. 

Since proxies are complicated and Im no expert, I did not want to make a change to Bref to modify `HTTP_X_FORWARDED_FOR` or the `setRemoteAddress`. 

----

Note also that if we were changing `$request->setRemoteAddress(...)`, that would break BC.

mnapoli: Right, that makes sense thanks.

Sorry for delaying this, but another question then: would it make more sense to get the client's IP address from the X-Forwarded-For headers instead?

I'm just making sure that there isn't a more standard solution for this. Ideally getting the IP address of the client shouldn't be specific to Bref or AWS Lambda.

Nyholm: > would it make more sense to get the client's IP address from the X-Forwarded-For headers instead?

I think so. But Im not sure. That is why I went with the more safe option for Bref in case Im wrong.

mnapoli: Right. I am cautious about adding this because:

- removing it in the future is a BC break
- it is Bref-specific

That's why I'm wondering about those headers. After all, the same problem would exist with any PHP application behind a proxy. And since you already call `Request::setTrustedProxies()`, it might already work.

deleugpn:
I don't want to discuss the merits of the PR as I don't have a formed opinion yet. But if this were to be merged, wouldn't it be better to **not** use `LAMBDA_` in the variable prefix? I'd imagine that it would be better to use a custom namespace (maybe `BREF_`) and let AWS keep owning the `LAMBDA_` context.

mnapoli:
Sorry I'll be closing this for now as I don't think I'll be able to look into this more.

I'd rather prioritize using `X-Forwarded-For` since it's the "official" way of getting the user's IP address:

- either the IP address is in there, then we don't need to do anything (except maybe document it)
- either the IP addres is not in there, or not always, or is not reliable, then we could maybe put the real IP address in `X-Forwarded-For` manually? But I'm not sure if it makes sense: why would API Gateway provide the real IP address if CloudFront cannot provide it in `X-Forwarded-For` 🤔

chekalsky:
I propose to re-open this issue as if we are using CloudFront we would need to have not really convenient option I've described [in Slack](https://brefworkspace.slack.com/archives/CF642J24R/p1614721230035700?thread_ts=1614639997.032800&cid=CF642J24R) to get user's IP.

At the same time `['LAMBDA_REQUEST_CONTEXT']['http']['sourceIp']` seems working fine, but bref is not using it in [here](https://github.com/brefphp/bref/blob/9150472ae1607cca9d2da79378055796ea317d09/src/Event/Http/FpmHandler.php#L184).

@mnapoli @Nyholm What do you think? Wouldn't it be beneficial to have sourceIp in `REMOTE_ADDR` field?

PR #659, opened 2020-05-23
Update message Attributes

Nyholm:
I just want to make sure it is correct. 

Related to https://github.com/brefphp/symfony-messenger/pull/28

### DIFF
--- a/tests/Event/Sqs/SqsEventTest.php
+++ b/tests/Event/Sqs/SqsEventTest.php
@@ -21,7 +21,7 @@ class SqsEventTest extends TestCase
                 'stringValue' => 'my value',
                 'stringListValues' => [],
                 'binaryListValues' => [],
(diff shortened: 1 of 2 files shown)


mnapoli:
Thanks, I think the test that uses this needs to be updated because the build is now failing.

Also, are you sure about that change? Here is an example of a SQS event from the AWS docs: https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html Attributes do not look like what you changed.

Nyholm:
Yes. Im sure. 

Note that there is a difference between `messageAttributes` and `attributes`. The link you sent is not showing any `messageAttributes`. 

I'll fix the test.

mnapoli:
Yes of course! That's perfect thanks!

Nyholm:
Thank you for merging.

PR #662, opened 2020-06-03
fix: prefer not using getenv

shouze:
as it's not a thread safe function.

### DIFF
--- a/docs/frameworks/symfony.md
+++ b/docs/frameworks/symfony.md
@@ -62,7 +62,7 @@ Since [the filesystem is readonly](/docs/environment/storage.md) except for `/tm
     public function getLogDir()
     {
         // When on the lambda only /tmp is writeable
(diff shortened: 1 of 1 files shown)


mnapoli:
Thank you.

Could you explain the actual problem (in Lambda or when developing locally)? We use getenv() all over the place at the moment.

shouze:
> Thank you.
> 
> Could you explain the actual problem (in Lambda or when developing locally)? We use getenv() all over the place at the moment.

yes of course I've already fixed on async-aws because it didn't fit well with Symfony dotenv component.

https://github.com/async-aws/aws/issues/632#issuecomment-630735708

I'm not sure we should have big issues when running lambdas but would prefer to avoid using it. I can make another PR to remove usage of getenv/putenv everywhere if you want.

mnapoli:
OK, since the diff is pretty simple let's merge that.

Thanks

PR #664, opened 2020-06-05
add new command to check if we use the last runtime

Fabex:
When we update bref/bref version we should verify if our current layers set in our template are up to date.
Sometime we can forget to check on https://runtimes.bref.sh the compatible layer with our bref version.
It will be great if we can have a command usable in CI to check this.
So I add a command to check this, so we can now add this command in CI, if all layers are up to date CI will be green, if not Ci will be red.

It's little crappy but the idea is here, I hope it can be usefull and I'm open to discuss.

little screenshot on a current project
brefRuntimeCheckerCommand

### DIFF
--- a/bref
+++ b/bref
@@ -452,7 +452,7 @@ $app->command('layer-checker [--region=]', function($region, SymfonyStyle $io) {
 
     $template = Yaml::parseFile('template.yaml', Yaml::PARSE_CUSTOM_TAGS);
     $layers = [];
-    $return = 0;
+    $notUpToDate = false;
     foreach ($template['Resources'] as $resource => $resourceConfig) {
         if (false === array_key_exists('Layers', $resourceConfig['Properties'])) {
             continue;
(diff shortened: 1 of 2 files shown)


mnapoli:
First of all, sorry for the delay, I had a look at this. It is tightly coupled to `template.yaml`, the SAM format. But in Bref, we don't support that format explicitly.

Maybe a better route would be to have a command that returns the layer version to use for the current Bref version?

Something like this:

```sh
$ vendor/bin/bref layers [--region=]
The layer versions matching the current Bref version are:

- arn:aws:lambda:us-east-1:416566615250:layer:php-74:14
- arn:aws:lambda:us-east-1:416566615250:layer:php-74-fpm:14
...
```

It's a little less automated, but at least it would support SAM, Terraform, CloudFormation, serverless.yml (without the Bref plugin) and more.

mnapoli:
Replaced by #730

PR #670, opened 2020-06-18
Default HTTP request event version to 1.0

skyrpex:
The event version does not appear to be defined in some cases, so PHP throws a warning because it's trying to access an undefined property.


mnapoli:
Thank you!

I'll merge and add a test covering that.

PR #671, opened 2020-06-19
Fixed the DomainName for s3-website

Chtiadrien:
Hello,

I copy/pasted the documentation for my latest project and when I deployed it, I got an S3 error from Cloudfront. After further inspection, I found the problem in the DomainName. 

S3 website domain name are written like that: `{bucket_name}.s3-website.eu-west-3.amazonaws.com`. In the documentation, it was like that: `{bucket_name}.s3-website-eu-west-3.amazonaws.com`

Hope it helps.

Edit: I've also fixed the error from PrettyCI 😉

### DIFF
--- a/docs/websites.md
+++ b/docs/websites.md
@@ -181,7 +181,7 @@ resources:
                             S3OriginConfig: {} # this key is required to tell CloudFront that this is an S3 origin, even though nothing is configured
                             # If you host a static website, like a SPA, use s3-website URLs instead of the config above
                             # See https://stackoverflow.com/questions/15309113/amazon-cloudfront-doesnt-respect-my-s3-website-buckets-index-html-rules#15528757
-                            # DomainName: !Join ['', [!Ref Assets, '.s3-website-', !Ref AWS::Region, '.amazonaws.com']]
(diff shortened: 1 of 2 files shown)


mnapoli:
Thanks! The thing is that the URL format depends on the region, see https://docs.aws.amazon.com/AmazonS3/latest/dev/WebsiteEndpoints.html

So it may be a `.` or a `-` 😕 (which is not great for user experience)

Maybe we could update the comment to link to this documentation page instead?

Chtiadrien:
TIL

I've only worked in the eu-west-3 region, I didn't know this. 
I'll update the doc with a link to a nice table I found in the AWS documentation 😉 

Thanks!

Chtiadrien:
I've been doing for research to simplify the configuration for the end user. I've end up with this:

```yaml
DomainName: !Select [2, !Split ["/", !GetAtt Assets.WebsiteURL]]
```

The WebsiteUrl attributes return the following: `http://mystack-mybucket-kdwwxmddtr2g.s3-website-us-east-2.amazonaws.com`. I then use Split and Select functions to extract the domain name.

If this is OK with you, I'll commit my changes.

mnapoli:
Oh that's clever! Yes let's do that, thank you!

mnapoli:
Thanks!

PR #672, opened 2020-06-22
Added support for multi-value query parameters

victormacko:
This change allows multi-parameter queries to be used when using Version 1 of the API Gateway events (v2 of the API Gateway payloads had already worked);

eg. my-url?shapes[]=square&shapes[]=circle&shapes[]=oval
Resulting in;
$_GET['shapes'] = ['square', 'circle', 'oval']

I've also tested for multi-dimensional arrays;
eg. my-url?shapes[][]=square&shapes[][]=circle
Resulting in;
$_GET['shapes'] = [['square']['circle']]

AND;
eg. my-url?shapes[][a]=square&shapes[][b]=circle
Resulting in;
$_GET['shapes'] = [['a' => 'square']['b' => 'circle']]

AND;
eg. my-url?shapes[a][]=square&shapes[a][]=circle
Resulting in;
$_GET['shapes'] = ['a' => ['square', 'circle']]

The change is especially useful for sending requests of a number of elements (with the same name) via the query-string (eg. a bunch of checkboxes in a form, or when getting a bunch of data by their ID).

I've added a few scenarios in the tests to make sure this works as well as it can, but a few fresh sets of eyes over it all would be welcome. The changes have also been tested on Lambda itself... but again, some extra testing would be nice to confirm all is well :)

Thanks in advance!

### DIFF
--- a/tests/Event/Http/Fixture/ag-v1-query-string-multivalue.json
+++ b/tests/Event/Http/Fixture/ag-v1-query-string-multivalue.json
@@ -18,8 +18,11 @@
         "foo": "bar"
     },
     "multiValueQueryStringParameters": {
-        "foo": ["bar"],
-        "shapes[]": ["circle", "square"]
+        "foo[]": ["bar", "baz"],
+        "cards[]": ["birthday"],
+        "colors[][]": ["red", "blue"],
+        "shapes[a][]": ["square", "triangle"],
+        "myvar": ["abc"]
     },
     "pathParameters": null,
     "stageVariables": null,
--- a/tests/Event/Http/Fixture/ag-v1-query-string-multivalue.json
+++ b/tests/Event/Http/Fixture/ag-v1-query-string-multivalue.json
@@ -15,10 +15,11 @@
         "X-Forwarded-Proto": "https"
     },
     "queryStringParameters": {
-        "foo": "baz"
+        "foo": "bar"
     },
     "multiValueQueryStringParameters": {
-        "foo": ["bar", "baz"]
+        "foo": ["bar"],
+        "shapes[]": ["circle", "square"]
     },
     "pathParameters": null,
     "stageVariables": null,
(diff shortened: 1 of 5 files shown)


victormacko:
Hey @mnapoli, was there anything else I can help with on this to progress it further? I've been running it in prod for a month and a half already and it's been running well.

mnapoli:
Thanks, that's good to know that you are running this in prod, and it looks a lot like #693 so now I got my head around this.

I will _try_ to merge this today or tomorrow 🤞

tests/Event/Http/Fixture/ag-v1-query-string-multivalue.json
```diff
@@ -15,10 +15,14 @@
         "X-Forwarded-Proto": "https"
     },
     "queryStringParameters": {
-        "foo": "baz"
+        "foo": "bar"
```
mnapoli, line 18: Any reason for this change?

I "think" it was `baz` because (I think) API Gateway takes the last value of the array

victormacko: Pretty sure that when my changes got put in, this the data in the JSON would've failed the test if it was kept -- and from memory only the abc[] vars had multi-vars, not the 'foo' variant (eg. foo[] would have multi-vars, foo not). The 'shapes[]' inclusion below it tested multi-vars, and I went into much more detail later on for the FpmHandler

victormacko:
> Thanks, that's good to know that you are running this in prod, and it looks a lot like #693 so now I got my head around this.
> 
> I will _try_ to merge this today or tomorrow 🤞

Yep, it looks very similar, though i've url-encoded the value (so values with myval&myval2=anotherval don't get mis-processed ... it was over a month ago so wouldn't hurt checking that! ;) )

victormacko:
Was there anything else you needed on this @mnapoli ? ... let me know if I can help at all.

mnapoli:
Thank you!

PR #673, opened 2020-06-28
Use version 1.0 of AsyncAws deps

Nyholm:
Wohoo, I finally tagged 1.0 of these libraries.


mnapoli:
Awesome!

No bc-break to expect here in how we use the package? I tried to check the release history in the main repository but there are no releases there it seems (nor changelog).

Nyholm:
The change logs are at the individual packages. 

The only changes were internal (added private methods) and adding support for php8.

mnapoli:
Thank you!

PR #674, opened 2020-06-28
Add very basic analytics via a ping

mnapoli:
I've had this idea for a long time, I want to collect feedback on this and try to implement it to get a better idea on Bref's usage.

Here are the explanations, taken from the docblock in the code:

> Ping a Bref server with a statsd request.

### WHY?

This ping is used to estimate the number of Bref invocations running in production.

Such statistic can be useful in many ways:
- so that potential Bref users know how much it is used in production
- to communicate to AWS how much Bref is used, and help them consider PHP as a native runtime

### WHAT?

The data sent in the ping is anonymous. It does not contain any identifiable data about anything (the project, users, etc.).

The only data it contains is: "A Bref invocation happened".
You can verify that by checking the content of the message in the function.

### HOW?

The data is sent via the statsd protocol, over UDP.

Unlike TCP, UDP does not check that the message correctly arrived to the server. It doesn't even establishes a connection.

That means that UDP is extremely fast: the data is sent over the network and the code moves on to the next line.

When actually sending data, the overhead of that ping takes about 150 micro-seconds.

However, this function actually sends data every 100 invocation, because we don't need to measure *all* invocations. We only need an approximation.

That means that 99% of the time, no data is sent, and the function takes 30 micro-seconds.

If we average all executions, the overhead of that ping is about 31 micro-seconds. Given that it is much much less than even 1 milli-second, we consider that overhead negligible.

### CAN I DISABLE IT?

Yes, set the `BREF_PING_DISABLE` environment variable to `1`.

About the statsd server and protocol: https://github.com/statsd/statsd
About UDP: https://en.wikipedia.org/wiki/User_Datagram_Protocol

### DIFF
--- a/src/Runtime/LambdaRuntime.php
+++ b/src/Runtime/LambdaRuntime.php
@@ -92,6 +92,8 @@ final class LambdaRuntime
         /** @var Context $context */
         [$event, $context] = $this->waitNextInvocation();
 
+        $this->ping();
+
         $result = null;
 
         try {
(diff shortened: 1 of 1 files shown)


deleugpn:
Not only this is amazing for bref and I'd happily enable it, but it's also an amazing background research that I want to use myself on data collection where I work.
The performance efficiency and the ease of deployment of this is just A-grade work.

PR #677, opened 2020-07-01
Add detail to environment variables documentation when using ssm SecureString

jakelodge:
Provide further detail on using and decrypting ssm values that are encrypted, to encourage storing secrets securely

### DIFF
--- a/docs/environment/variables.md
+++ b/docs/environment/variables.md
@@ -40,7 +40,19 @@ Instead you can use the [SSM parameter store](https://docs.aws.amazon.com/system
 To create a parameter you can either do it manually in the [SSM parameter store console](https://console.aws.amazon.com/systems-manager/parameters) or use the following command:
 
 ```bash
-aws ssm put-parameter --region us-east-1 --name '/my-app/my-parameter' --type String --value 'mysecretvalue'
+aws ssm put-parameter --region us-east-1 --name '/my-app/my-parameter' --type String --value 'myvalue'
+```
+
+```bash
+aws ssm put-parameter --region us-east-1 --name '/my-app/my-secret' --type SecureString --value 'mysecretvalue'
+```
+
+Use SecureString for sensitive data, e.g. Database passwords or API keys, and decrypt as below:
+```yaml
+provider:
+    # ...
+    environment:
+        MY_PARAMETER: ${ssm:/my-app/my-secret~true}
 ```
 
 For Windows users, the first part of the parth needs to be double slashes and all subsequent forward slashes changed to backslashes:


mnapoli:
Thank you!

I'm not sure to be honest, I agree that this is more secure in theory, but the secret ends up being decrypted on deployment and inlined in the Lambda configuration. So in the end, the secret is exposed.

I think in simple cases, that's acceptable, which is why I documented that way. But for some companies, it needs to be more secure. And simply switching to a `SecureString` is not enough: you would need also to fetch the decrypted value at runtime from PHP, instead of decrypting on deployment.

Does that make sense? I'm afraid it provides a false sense of additional security.

jakelodge:
Yes, I think I can see where you're coming i.e security in the context of the PHP app, but I'd kind of expect any platforms requiring this level of security would be using something other than SSM anyway?

Still learning but it does provide additional security from other apps in the VPC, from what I understand. It's the method documented here, anyway: https://www.serverless.com/framework/docs/providers/aws/guide/variables/#reference-variables-using-the-ssm-parameter-store

mnapoli:
Right, but do we know how it is making things more secure? Here is my understanding.

Who can see the non-secured string:

- AWS employees
- anyone with access to the SSM parameter
- anyone with access to the deployed Lambda function

Who can see the secured string:

- anyone with access to the SSM parameter (it is possible to get the decrypted value)
- anyone with access to the deployed Lambda function

This is why I don't see the point, but maybe I'm missing something?

jakelodge:
When you say anyone with access to the SSM parameter can see the decrypted SecureString, that's only true if they have kms:Decrypt IAM permissions. Likewise they must have access to that deployed Lambda.

The fact they're disclosed on deployment is sub-optimal for sure, but I don't see that as a valid reason not to store them encrypted, for lots of the same reasons we never store users' passwords in plaintext.

The caveat that secrets will be decrypted on deployment when using this method `...~true` would be a good addition though.

deleugpn:
> The fact they're disclosed on deployment is sub-optimal for sure, but I don't see that as a valid reason not to store them encrypted, for lots of the same reasons we never store users' passwords in plaintext.

I'm not going to comment on the rest of the thread as I'm on the fence myself on the problem, but the comparison with password plaintext is irrelevant at best. Passwords are hashed, not encrypted. The goal is to be impossible to reverse back to the original password text. Encryption is always reversible.

mnapoli:
> When you say anyone with access to the SSM parameter can see the decrypted SecureString, that's only true if they have kms:Decrypt IAM permissions.

Thanks that's a good point.

So, I like to keep things simple. Here is how I would summarize the situation to people who don't know the details:

- storing secrets in a "clear" string + injecting it at deploy time (as documented) is "good enough" for most use cases
- if you need more security, you would need to store the secret encrypted + decrypt it at runtime using the PHP SDK (not documented yet, that would be a good addition)

A 3rd option would be what you suggest: to use `SecureString` + setup IAM permissions so that developers cannot access the KMS key as well as the Lambda itself + secure the deployment environment (because secrets pass decrypted there), but that solution seems less practical to me. That why I don't want to document it. I'd rather document the "decrypt at runtime" option which seems easier to achieve. Documenting many ways of achieving the same result brings too much confusion.

I'm going to close this one, but thank you for bringing that topic up.

deleugpn:
A 4th (more expensive) option is to use AWS Secrets. It's kind of like option 2, but AWS will do the deciphering for you on Runtime and inject the actual value as environment variable in such a way that is not visible for people accessing the console.

mnapoli:
@deleugpn do you have a link that explains what you mention? I always assumed that AWS Secrets worked like SSM parameters in that case (when injecting them in Lambda). When we reference them in serverless.yml they are resolved too right?

deleugpn:
@mnapoli I don't think they are. You can check this link: https://aws.amazon.com/blogs/security/how-to-securely-provide-database-credentials-to-lambda-functions-by-using-aws-secrets-manager/

If you see the screenshot they put on the end of the article, you see the AWS Secret IAM as an environment variable and not the actual value. The value is resolved only during runtime.

mnapoli:
Right, even if that's the case, is serverless.yml doing that?

Here is what the documentation (https://www.serverless.com/framework/docs/providers/aws/guide/variables/#reference-variables-using-aws-secrets-manager) says:

> In this example, the serverless variable will contain the decrypted value of the secret.

So the variable in the YML file will contain the value in cleartext, this is why I assume it is deployed like this (which isn't more secure). I haven't tested though, so I'm just assuming.

deleugpn:
Scratch everything I said. AWS Lambda does not have  `Secrets` support that AWS ECS (Containers) do. If you check [this link](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-taskdefinition-containerdefinitions.html#cfn-ecs-taskdefinition-containerdefinition-secrets) you'll see that in addition to `Environment`, ECS also have support for `Secrets`. When specifying secrets for a container, you may specify just the ARN of the secret and AWS will resolve it when starting the container. The actual value is not visible in plain text.

Even the article I linked above, the code inside Lambda is taking the ARN and resolving it during execution which is **option 2** you already mentioned above. I thought this was available for Lambda as well, but turns out it isn't.

About the serverless framework, if AWS Lambda ever starts supporting `Secrets`, then you wouldn't use the `resolve:ssm` on Serverless because that triggers a `resolve` by CloudFormation. What you'd probably do is use the Cloudformation support for Serverless to declare a Secret resource and `!Ref` just the ARN as a Secret. That way CloudFormation or Serverless framework would just treat it as a dumb string while AWS would parse it and resolve it during runtime.

mnapoli:
Right OK thanks for the clarification, that makes sense then!

PR #680, opened 2020-07-10
Bumped PHP versions to latest for all images

Chtiadrien:
I'm back again with an update for docker images. Bref is not concerned by the security update because it's only impacting Windows builds. 

All tests are passing:

```console
➜  runtime $: php layers/tests.php 
...................................
Tests passed
```


mnapoli:
Thanks

Issue #422, opened 2019-08-22
[regions] Publish runtimes on China regions.


ac1982:
Please publish the runtimes on chinese regions, like:
cn-northwest-1 (宁夏)
cn-north-1    (北京)
Thanks a lot.

mnapoli:
:+1: if anyone wants to have a go at this you can probably find the region list in this repository easily, but there is also https://github.com/brefphp/runtimes.bref.sh to update.

mnapoli:
Hey @AlessandroMinoccheri was fast 😄, here is the PR in the other repository: https://github.com/brefphp/runtimes.bref.sh/pull/1

I will merge https://github.com/brefphp/runtimes.bref.sh/pull/1 when runtimes are published, so I'll keep it open in the meantime.

What's missing to publish the runtimes:

- update the build scripts in this repository to add the new regions (PR welcome)
- build and publish the runtimes (I'll take care of that)

bnusunny:
@mnapoli AWS China regions is a seperate partition, which require a seperate account. I can help to build and push the runtimes in China region.

bnusunny:
Here are the layers published in China regions. 

layer | region | arn 
---|---|---
php-73|cn-north-1|arn:aws-cn:lambda:cn-north-1:215043275130:layer:php-73:2
php-73-fpm|cn-north-1|arn:aws-cn:lambda:cn-north-1:215043275130:layer:php-73-fpm:2
php-72|cn-north-1|arn:aws-cn:lambda:cn-north-1:215043275130:layer:php-72:2
php-72-fpm|cn-north-1|arn:aws-cn:lambda:cn-north-1:215043275130:layer:php-72-fpm:2
console|cn-north-1|arn:aws-cn:lambda:cn-north-1:215043275130:layer:console:2
php-73|cn-northwest-1|arn:aws-cn:lambda:cn-northwest-1:215043275130:layer:php-73:2
php-73-fpm|cn-northwest-1|arn:aws-cn:lambda:cn-northwest-1:215043275130:layer:php-73-fpm:2
php-72|cn-northwest-1|arn:aws-cn:lambda:cn-northwest-1:215043275130:layer:php-72:2
php-72-fpm|cn-northwest-1|arn:aws-cn:lambda:cn-northwest-1:215043275130:layer:php-72-fpm:2
console|cn-northwest-1|arn:aws-cn:lambda:cn-northwest-1:215043275130:layer:console:2

PR #681, opened 2020-07-10
Added all missing region

zxc20033:
Added all missing region

### DIFF
--- a/runtime/layers/regions.json
+++ b/runtime/layers/regions.json
@@ -14,8 +14,12 @@
     "ap-east-1",
     "ap-northeast-1",
     "ap-northeast-2",
+    "ap-northeast-3",
     "ap-southeast-1",
     "ap-southeast-2",
     "eu-south-1",
-    "af-south-1"
+    "af-south-1",
+    "me-south-1",
+    "cn-north-1",
+    "cn-northwest-1"
 ]
--- a/runtime/layers/regions.json
+++ b/runtime/layers/regions.json
@@ -11,6 +11,7 @@
     "us-west-1",
     "us-west-2",
     "ap-south-1",
+    "ap-east-1",
     "ap-northeast-1",
     "ap-northeast-2",
     "ap-southeast-1",


Chtiadrien:
Maybe we should add all missing regions?

With `ap-east-1` added, we are only missing:

```
ap-northeast-3 (Asia Pacific - Osaka-Local)
cn-north-1 (China - Beijing)
cn-northwest-1 (China - Ningxia)
me-south-1 (Middle East - Bahrain)
```

zxc20033:
> Maybe we should add all missing regions?
> 
> With `ap-east-1` added, we are only missing:
> 
> ```
> ap-northeast-3 (Asia Pacific - Osaka-Local)
> cn-north-1 (China - Beijing)
> cn-northwest-1 (China - Ningxia)
> me-south-1 (Middle East - Bahrain)
> ```

Yep, I have checked these regions are providing lambda. Updated the pull request to include all regions.

Chtiadrien:
I've double check and I realize that the `ap-northeast-3 (Asia Pacific - Osaka-Local)` does not provide AWS Lambda 😕  Sorry about that.

image

The region will be standardized in the early 2021 : https://aws.amazon.com/about-aws/whats-new/2020/01/aws-to-launch-standard-aws-region-in-osaka-in-early-2021/

zxc20033:
> I've double check and I realize that the `ap-northeast-3 (Asia Pacific - Osaka-Local)` does not provide AWS Lambda 😕 Sorry about that.
> 
> image
> 
> The region will be standardized in the early 2021 : https://aws.amazon.com/about-aws/whats-new/2020/01/aws-to-launch-standard-aws-region-in-osaka-in-early-2021/

weird, from this blog I saw there are lambda layers for ap-northeast-3 https://aws.amazon.com/tw/blogs/compute/upcoming-changes-to-the-python-sdk-in-aws-lambda/ 
image

richardnbanks:
The CN regions cannot be added as a separate account is required for the Chinese region. https://github.com/brefphp/bref/issues/422#issuecomment-524598714

mnapoli:
Thanks for the PR!

> The CN regions cannot be added as a separate account is required for the Chinese region. #422 (comment)

Until we have an easy way to deal with this, I will not publish layers on these regions unfortunately :/

Also, unless we are certain about `ap-northeast-3` I'd rater remove it from the list (building layers takes a lot of time, I'd rather avoid loosing time on this).

mnapoli:
Closing because it seems this PR is stuck for now.

Issue #535, opened 2020-01-22
Url broken when using cloudfront in front of lambda


jmolinap:
Hi, great package.

Currently i am testing an app built with symfony but redirections/url generation are broken.

Redirecting to a non absolute url within the app, generates a url of the type

somerandom-url.execute.aws-region.com

#67  seems to be related.

Any pointers on how to fix?

Thanks.

mnapoli:
Hi, this is a limitation of CloudFront: the host header is not forwarded by CloudFront.

That means that the HTTP request landing in API Gateway does not contain the host from CloudFront, but instead contains the API Gateway URL.

Possible solutions:

- force the HTTP header `Host` to a specific value (your domain name) in your CloudFront configuration
- force your app/framework to use a specific domain name

jmolinap:
Thanks for the proposed solutions.

Any doc that i can check to achieve properly?, i have tried to set up a lambda function to modify the header but it messes up the app, after setting up it returns a 403 forbidden, disabling it makes it work again - page loads (with broken urls)

Thanks.

mnapoli:
Right, sorry I wasn't a little more explicit.

Here is a solution to configure the CloudFront CDN in serverless.yml:

```yaml
Origins:
  # The website (AWS Lambda)
  - Id: Website
    DomainName: '#{ApiGatewayRestApi}.execute-api.#{AWS::Region}.amazonaws.com'
    OriginPath: '/dev'
    CustomOriginConfig:
      OriginProtocolPolicy: 'https-only' # API Gateway only supports HTTPS
    OriginCustomHeaders:
      - HeaderName: 'X-Forwarded-Host'
        HeaderValue: 'your-domain-here.com'
```

Then you need to configure your framework to take the `X-Forwarded-Host` header into account.

For Symfony, have a look at https://symfony.com/doc/current/deployment/proxies.html#but-what-if-the-ip-of-my-reverse-proxy-changes-constantly
Though it's interesting to see that Symfony mentions CloudFront, so maybe you can skip the configuration above and try this directly: https://symfony.com/doc/current/deployment/proxies.html#custom-headers-when-using-a-reverse-proxy (please report if it works)

For Laravel it seems that [this package](https://github.com/fideloper/TrustedProxy) is now installed by default. You can configure it: https://github.com/fideloper/TrustedProxy#configure-trusted-proxies

In case you wonder if the CloudFront proxy has an IP address: no. This is why in those docs above you often need to use `*`.

[...]

PR #682, opened 2020-07-10
Update laravel docs and include Host header forwarding

tomhatzer:
This PR includes updated docs for the `Host` header problem that came up in https://github.com/brefphp/bref/issues/535

I had the same problem yesterday and didn't find enough information on how to do this. @mnapoli was too kind responding [on twitter](https://twitter.com/matthieunapoli/status/1281479679480860673) on how to solve my problem and point me into the right direction.

### DIFF
--- a/docs/frameworks/laravel.md
+++ b/docs/frameworks/laravel.md
@@ -111,6 +111,42 @@ If your application creates links and redirections to HTTP URLs (which are inval
 +    protected $proxies = '*';
 ```
 
+### Forwarding correct Host header
+
+When not enabling host forwarding, Laravel will use the hostname/domain from Cloudfront while building internal links (routes).
+To fix this, you can use one of the following methods. Please also have a look at the HTTPS section above for allowing the AWS proxies.
+
+#### Method 1: Extending serverless.yml
+
+You can add the following `OriginCustomHeaders` part to your `Origins` setting in your `serverless.yml` file:
+
+```diff
+Origins:
+  # The website (AWS Lambda)
+  - Id: Website
+    DomainName: '#{ApiGatewayRestApi}.execute-api.#{AWS::Region}.amazonaws.com'
+    OriginPath: '/dev'
+    CustomOriginConfig:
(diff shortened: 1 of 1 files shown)


docs/frameworks/laravel.md
```diff
@@ -111,6 +111,42 @@ If your application creates links and redirections to HTTP URLs (which are inval
...
+You can whitelist the `Host` header in your Cloudfront Distribution like this (to forward the `Host` header from the request):
+
+- Open the Amazon CloudFront console, and then choose your distribution.
+- Choose the Behaviors view, and then choose the path you are using.
+- Choose Edit.
+- For Cache Based on Selected Request Headers, choose Whitelist.
+- Under Whitelist Headers, choose Host from the column on the left, and then choose Add.
+- Choose Yes, Edit.
+
+*Origin: https://aws.amazon.com/de/premiumsupport/knowledge-center/configure-cloudfront-to-forward-headers/*
```
mnapoli, line 147: Unfortunately this breaks API Gateway (because it doesn't know which API Gateway to call, see https://stackoverflow.com/a/42259852/245552). So this is not an option we can use 😞

mnapoli:
I am wondering if this is not something we should put in the global "website" documentation instead 🤔  (https://bref.sh/docs/websites.html) Because after all, it applies to all websites.

docs/frameworks/laravel.md
tomhatzer: That's odd, I've got this working with API Gateway v2 and Host header forwarding for my Laravel app.

mnapoli:
After thinking about this: we should put this config inside the big "CloudFront" configuration example in the "website" page. That way it is setup out of the box by default.

docs/frameworks/laravel.md
mnapoli: That is very weird indeed, can you double check? I tried it but it doesn't work for me. Are you simply whitelisting the "Host" header (so that it's forwarded by CloudFront to API Gateway)?

mnapoli:
Closing for now

PR #690, opened 2020-07-15
use `hello` as the function name

chapeupreto:
Use `hello` as the function name in order to avoid any confusion with the `functions` keyword.

By doing that, the template and the example in the documentation remain equals.


mnapoli:
Good point, I see how it could be very confusing.

👍

PR #691, opened 2020-07-16
change from php-73 to php-74 in template files

chapeupreto:
I don't know, but maybe it would be nice to have php74 runtime set as default. What do you think?


mnapoli:
Thanks, yes let's do this.

Could you also update the documentation so that it's all consistent?

chapeupreto:
Nice!

I've updated the documentation files as well.

mnapoli:
Perfect thanks!

PR #693, opened 2020-07-16
Handle ALB Multivalue Request

deleugpn:
Possible fix for #456 and alternative for https://github.com/brefphp/bref/pull/538.

I finally understood the issue today when trying to migrate a reporting service from Fargate to Lambda. Any request that includes values such as `parameter[key][]=value&parameter[key][]another-value` seems to be broken under the current implementation.

In order to debug this, I followed Napoli's advice on changing the `vendor` folder and deploying from local. I changed the `LambdaRuntime::processNextEvent` method as follows

```
    public function processNextEvent($handler): void
    {
        /** @var Context $context */
        [$event, $context] = $this->waitNextInvocation();

        $this->signalFailure($context->getAwsRequestId(), new Exception(json_encode($event)));

        return;
```

This allowed me to fetch the entire `$event` on CloudWatch, but double-encoded. I decoded it myself once and added the result to `alb-query-string-multivalue.json`. For privacy reasons, a few values were changed, such as AWS Account ID, Target Group Name, IP Addresses, Cookies. The Request Parameter I kept exactly as-is from my actual use case and I noticed that they don't get correctly decoded by the `HttpRequestEvent`.

By identifying the ALB request, we can `urldecode` the parameter names to get it properly working.

-----------------------------------

Digging deep through this I noticed that some reports in my application works and some don't. Further investigating I noticed that it's possible for the Javascript request to go encoded with `%5B` and `%5D`, but it's also possible to send requests without encoding that properly works. I plan on discussing this with the Frontend team tomorrow to understand what makes some requests encode and what makes some requests not encode.

Maybe it's safe to **always** try to decode, but I want to add more tests for that as well as understand why some requests are encoded and some are not.

Relevant: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html#multi-value-headers


deleugpn:
I'm testing query parameters in 3 formats: encoded, unencoded and mixed. 

- Encoded means **everything** is encoded, including the parameter names itself.
- Unencoded means **nothing** is encoded, including the values.
- Mixed includes a test where some parameters follow the **encoded** rule while 1 extra parameter has only **values** encoded, not the parameter name itself.

This is all working fine for Application Load Balancers with multi value disabled. This means that only the last parameter sent is actually received. I want to make sure that this alternative work and is supported, hence these tests.

However, multi-value disabled is not an option for the project I'm working on, so my next step now is to figure out a way to support **multi value enabled** without breaking multi value **disabled**. This includes all 3 scenarios again: encoded, unencoded and mixed with multi value enabled.

Right now my challenge is that with multi value enabled, AWS keeps the `[]` notation at the end of the key, but accumulate all values in a single array at position [0]. I'm trying to unpack that array back to the original key while decoding all values.

deleugpn:
Final checklist:

- [x] Plain parameters with Multi Value disabled
- [x] Encoded parameter keys with Multi Value disabled
- [x] Encoded parameter values with Multi Value disabled

- [x] Plain parameters with Multi Value enabled
- [x] Encoded parameter keys with Multi Value enabled
- [x] Encoded parameter values with Multi Value enabled

The primary difference between Multi Value disabled and Enabled is that when it's disabled only the last value of an array is preserved whereas with Multi Value enabled everything becomes an array.

The proposal solution relies on `parse_str` automatically doing URL Decode (credit to @nuvolapl on #538). Building out a long string with all parameters at once allows the `param[0][field]&param[0][another_field]` to properly work. Recursive array merge breaks.
The extra `&` at the end of the string is not relevant (see https://3v4l.org/3a2HL).

deleugpn:
@mnapoli please let me know if there's anything I can do to ease the burden of this on you. I'm back from my holidays and I'm ready to help out on this.

src/Event/Http/HttpRequestEvent.php
```diff
@@ -177,6 +177,28 @@ private function rebuildQueryString(): string
...
+
+                foreach ($values as $value) {
+                    $queryString .= $key . '=' . $value . '&';
+                }
+            }
+
+            parse_str($queryString, $decodedQueryParameters);
+
+            return http_build_query($decodedQueryParameters);
+        }
```
mnapoli, line 200: Hey @deleugpn, sorry for the delay. The PR looks good, thanks a lot!

One thing that could definitely help with maintenance later is being able to understand why these lines of code exist. Could you add some inline comments (in the code, not the PR) about "why"?

Here are some questions for example (that could be answered in comments):

- why do we need to handle ALB events separately here? (ideally you can add a link to the github issue or this PR)
- why the `foreach`?
- why the `parse_str`?

(try to word it so that someone else in 2 years will have to understand why some part of code exists ^^)

src/Event/Http/HttpRequestEvent.php
```diff
@@ -171,13 +171,19 @@ private function rebuildQueryString(): string
...
             // Why? To match the format we get when using PHP outside of Lambda (we get the query string URL-encoded)
             parse_str($queryString, $queryParameters);
             return http_build_query($queryParameters);
         }
 
+        // It is likely that we do not need to differentiate between API Gateway (Version 1) and ALB. However,
+        // it would lead to a breaking change since the current implementation for API Gateway does not
+        // support MultiValue query string. This way, the code is fully backward-compatible while
+        // offering complete support for multi value query parameters on ALB. Later on there can
+        // be a feature flag that allows API Gateway users to opt-in to complete support as well.
```
mnapoli, line 184: Ohh that's interesting, so you mean that maybe we can use the same logic with API Gateway to support multivalue query string parameters ? That's not our goal in this PR, but that's useful for later, thanks!

deleugpn: I will make some time in the coming days to test my project behind API Gateway just to confirm this and if I find anything that invalidate my comment, I'll send a PR to change it, but I think this is correct because A) both ALB and API Gateway uses the same attribute name and B) they both use the same strategy to always send single value or always-array depending on 'multi-value' flag.

PR #694, opened 2020-07-16
Add native support for providing handlers from PSR-11 containers

mnapoli:
This PR enables to use PSR-11 containers (e.g. PHP-DI, Symfony container, Laravel container, etc.) as source for AWS Lambda handlers.

Said another way, that allows **to set PHP classes as Lambda handlers**, instead of files.

**Read the end of the PR for use case examples**.

### Before

```yaml
# serverless.yml
functions:
    sqs-worker:
        handler: sqs-worker.php
```

```php
<?php
// src/SqsWorker.php

class SqsWorker extends SqsHandler
{
    public function handleSqs(SqsEvent $event, Context $context): void
    {
        // process $event
    }
}
```

```php
<?php
// sqs-worker.php

return new SqsWorker()

// or when using a container/framework

$app = require_once __DIR__ . '/app/bootstrap.php`;
$worker = $app->getContainer()->get(App\SqsWorker::class);
return $worker
```

### After

```yaml
# serverless.yml
functions:
    sqs-worker:
        handler: App\SqsWorker
```

## How it works

As a reminder, PSR-11 provides this interface:

```php
interface ContainerInterface
{
    public function get($id);
    public function has($id);
}
```

It is implemented by most frameworks (including Laravel and Symfony), and it allows to provide services.

With this PR, Bref starts using `ContainerInterface` **to load the Lambda function handler.**

Said another way, with `handler: App\SqsWorker`, Bref will call `$container->get(App\SqsWorker)`.

**If no container is configured**, Bref has a default implementation that loads files. The goal of that default implementation is to provide compatibility with using files as handlers (`handler: sqs-worker.php`). With PSR-11, we can then support the best of both worlds: file handlers, or class handlers!

## How to set a container

By default, Bref will support file handlers (i.e. just like it works now).

To set a PSR-11 container and support class handlers, call `Bref::setContainer()` like so:

```php
<?php

Bref\Bref::setContainer($container);
```

Where? To make sure this line of code is automatically executed on each request/invocation, 3 options are considered:

- have Bref automatically load a specific file (e.g. `bootstrap.php`) if it exists in the project -> convention over configuration.
    It is not explicit enough to my liking: nowhere would it be explicit that this file is required by Bref.

- have Bref load a file, configurable via an environment variable (e.g. `BREF_AUTOLOAD=bootstrap.php`) -> configuration over convention.
    It is explicit, however it relies on a non-standard behavior (custom environment variable), I prefer more standard solutions.

- create any file you want and have Composer autoload it via its `autoload` system.
    It is explicit, and relies on a standard in PHP: Composer. The only downside is that it will run this file even outside of Bref.

This part might be adjusted following testing.

To sum up, create a `bootstrap.php` file (or name it however you want) and add it to `composer.json`:

```json
    "autoload": {
        "psr-4": {
            ...
        },
        "files": [
            "bootstrap.php"
        ]
    },
```

#### Laravel

For Laravel, the content of `bootstrap.php` should look like this (I need to double check):

```php
<?php

$app = require __DIR__ . '/bootstrap/app.php';

$kernel = $app->make(Illuminate\Contracts\Console\Kernel::class);
$kernel->bootstrap();

Bref\Bref::setContainer($app);
```

#### Symfony

For Symfony, the content of `bootstrap.php` should look like this:

```php
<?php

require dirname(__DIR__) . '/config/bootstrap.php';

$kernel = new \App\Kernel($_SERVER['APP_ENV'], (bool) $_SERVER['APP_DEBUG']);
$kernel->boot();

Bref\Bref::setContainer($kernel->getContainer());
```

#### PHP-DI

For [PHP-DI](https://php-di.org/), the content of `bootstrap.php` should look like this:

```php
<?php

$builder = new DI\ContainerBuilder;
$builder->addDefinitions('config.php');
$container = $builder->build();

Bref\Bref::setContainer($container);
```

## Use case examples

SQS queue workers:

```yaml
functions:
    sqs-worker:
        # This class implements Bref\Event\Sqs\SqsHandler
        handler: App\SqsWorker
        events:
            -   sqs:
                    arn: 'arn:aws:sqs:region:XXXXXX:my-queue'
```

S3 file uploads:

```yaml
functions:
    s3-handler:
        # This class implements Bref\Event\S3\S3Handler
        handler: App\UploadedFileHandler
        events:
            # Triggered by a file uploaded to the S3 bucket
            -   s3:
                    bucket: my-uploaded-files
```

No-framework HTTP applications:

```yaml
functions:
    home:
        handler: App\Controller\HomeController
        events:
            -   httpApi: 'GET /'
    search:
        handler: App\Controller\SearchController
        events:
            -   httpApi:
                     method: '*'
                     path: /search
    ...
```

(☝️  in this example `HomeController` and `SearchController` are PSR-7/PSR-15 handlers.)

and more with all the [typed event handlers that Bref supports](https://github.com/brefphp/bref/tree/b3b287d0e0245ca08f836ff820486d17e74ecbf3/src/Event).


deleugpn:
This is pretty awesome 🚀🚀🚀

runtime/layers/function/bootstrap.php
```diff
@@ -17,16 +18,12 @@
...
-/** @noinspection PhpIncludeInspection */
-$handler = require $handlerFile;
+$container = Bref::getContainer();
 
-if (! $handler) {
-    $lambdaRuntime->failInitialization("Handler `$handlerFile` must return a function or object handler. See https://bref.sh/docs/runtimes/function.html");
+try {
+    $handler = $container->get(getenv('_HANDLER'));
+} catch (Throwable $e) {
+    $lambdaRuntime->failInitialization($e->getMessage());
```
GeeH, line 26: It looks like this replaces existing functionality; is it possible to still support `index.php` notation so that this ins't a BC break?

GeeH: Ignore me I now saw the FileHandlerLocator; tricksy.

src/Runtime/FileHandlerLocator.php
```diff
@@ -0,0 +1,49 @@
...
+    {
+        $handlerFile = $this->directory . '/' . $id;
+        if (! is_file($handlerFile)) {
+            throw new HandlerNotFound("Handler `$handlerFile` doesn't exist");
+        }
+
+        /** @noinspection PhpIncludeInspection */
+        $handler = require $handlerFile;
+
+        if (! (is_object($handler) || is_array($handler))) {
```
GeeH, line 38: I may be wrong but I don't see this logic covered in the test?

GeeH: I see the test but it's not covering all the logical paths here.

runtime/layers/function/bootstrap.php
mnapoli: Yep! Total backward compatibility, and I don't even like calling this "backward" because using file handlers is still definitely a good thing, not deprecated at all.

src/Runtime/FileHandlerLocator.php
mnapoli: These lines were moved from [the original bootstrap file from master](https://github.com/brefphp/bref/blob/master/runtime/layers/function/bootstrap.php#L26-L30).

Then I added [that test](https://github.com/brefphp/bref/pull/694/files#diff-02500a3d32ab6aaea871128f38c66c60R22-R29) and realized that the code on master doesn't actually work (`if (! $handler) {` doesn't work because the `require` call actually return `1` instead of `null`). That's why I changed the logic to `is_object() || is_array()`.

So, that piece of code isn't extensively tested, it just covers the basic. But good point I'll keep this thread open to remember to add more tests (allow object | array).

GeeH: The reason I think a test would be nice is I spent 10 minutes on 3v4l.org figuring out what that line meant, i think it's if it isn't an object or an array which confused me further because I thought it could also be a callable (I think I'm wrong there). Not a big deal tho.

mnapoli: Yes you're right, but AFAIR here is what a callable could be:

- object implementing `__invoke()`
- closure, which is an object
- array (object + method name)
- string (function name) -> I'm fine not supporting this to avoid any weird case, I don't want for example to check that the string contains the name of a function that exists

And of course, a handler can be an object implementing `Handler`.

Hence why I simplified this to object | array.

GeeH: It makes sense, thanks for explaining I appreciate your time!

chekalsky:
@mnapoli @GeeH This indeed is awesome. Any plans on moving this PR to production?

mnapoli:
I need to change the `Bref\Bref::setContainer($container)` solution to something that doesn't mess up the global state.

One solution is to use a closure, which could be a good solution. I need to look into this.

chekalsky:
🎉 @mnapoli thank you

deleugpn:
Hey @mnapoli, I'm guessing if I do something like this: 

```
final class MyHandler {...}

return new MyHandler
```

and just point my template to `\My\Namespace\MyHandler`, since I'm returning `MyHandler` in the end of the file, it should work without me having to set a container, right? Is this an unintended consequence or can I rely on this to override the FpmHandler class?

mnapoli:
@deleugpn I'm not sure what you mean. The default behavior of Bref is [this](https://github.com/brefphp/bref/blob/master/src/Runtime/FileHandlerLocator.php#L30-L45), which does a `require` of what's defined in the `handler` of the Lambda.

If you put a class name (`My\Namespace\MyHandler`) in the handler, then it's not a valid file and it will not work.

If however you put the file name of the code example you used, then yes it will work out of the box because the file returns a handler. But that was already the case with previous versions of Bref, that's why I'm confused and I'm not sure I'm actually answering your question.

By the way, we could update the default behavior:

- if it looks like a class name, we do a `new $handler()` (and hope it works)
- if not we try to `require` it because it's probably a file

But it's starting to get magic, and I'm not sure there's a wide use case for classes without a container/dependency injection.

deleugpn:
@mnapoli yeah, I had misunderstood the Handler property. Instead of putting a namespace there, I realized I have to put the full file path, which is totally fine. I don't need the magic `new $handler()`. I was just wondering if this way I'm not "abusing" the bootstrap process with something that works, but you might not had the intention to support it.

PR #700, opened 2020-07-27
Specify max execution time

Nyholm:
With this change we allow FPM to cleanup its resources and make sure we log things properly. 

The user will no longer see `{"message": "Internal server error"}`. They will see: 

```
❯ curl -I https://foobar.execute-api.eu-central-1.amazonaws.com/dev
HTTP/2 500
content-type: application/json
content-length: 0
date: Mon, 27 Jul 2020 08:13:18 GMT
x-amzn-requestid: 4ba9fbb8-e559-455e-ab50-75b5cb87be22
x-amz-apigw-id: QUroliAFt9A=
x-amzn-trace-id: Root=1-5f1e8c9a-5e00490e;Sampled=0
x-cache: Error from cloudfront
via: 1.1 208ed8b46a4d.cloudfront.net (CloudFront)
x-amz-cf-pop: ARN4-C2
x-amz-cf-id: LvgYYbNbSkYx_lL-eKZqpJZw==
```

Screenshot 2020-07-27 at 10 15 10

----------

This is a draft PR, I'll see if I can do better.

### DIFF
--- a/runtime/layers/fpm/php.ini
+++ b/runtime/layers/fpm/php.ini
@@ -38,7 +38,7 @@ disable_functions=fastcgi_finish_request
 ; API Gateway has a timeout of 29 seconds, the default FPM job has a timeout of
 ; 28 seconds. Setting this to 27 will give FPM some time to properly finish up
 ; its resources and flush logs to CloudWatch.
-max_execution_time = 27
+max_execution_time=27
 
 ; The total upload size limit is 6Mb, we override the defaults to match this limit
 ; API Gateway has a 10Mb limit, but Lambda's is 6Mb
--- a/runtime/layers/fpm/php.ini
+++ b/runtime/layers/fpm/php.ini
@@ -35,6 +35,11 @@ variables_order="EGPCS"
 ; See https://github.com/brefphp/bref/issues/214
 disable_functions=fastcgi_finish_request
 
+; API Gateway has a timeout of 29 seconds, the default FPM job has a timeout of
+; 28 seconds. Setting this to 27 will give FPM some time to properly finish up
+; its resources and flush logs to CloudWatch.
+max_execution_time = 27
+
 ; The total upload size limit is 6Mb, we override the defaults to match this limit
 ; API Gateway has a 10Mb limit, but Lambda's is 6Mb
 post_max_size=6M


Nyholm:
### 2 requests before (`{"message": "Internal server error"}`):


Screenshot 2020-07-27 at 10 22 37


### 2 requests with this PR: 

Screenshot 2020-07-27 at 10 23 32


As we can see, it prints both my custom message and a PHP notice about the timeout. 


Here is my script: 

```
<?php

file_put_contents('php://stderr', 'Foobar 123' . PHP_EOL, FILE_APPEND);

header('Content-type: application/json');

$x = 0;
while(true) {
    $x++;
}
```

Nyholm:
The new 500 page may look horrible, but that is actually what we want and expect. That is what would happen on a normal server.

Nyholm:
Hm.. This does not work as expected. 

I've tested this out with php max: 2 and lambda max: 4 and it works. 

But when I change to php max: 20 and lambda max: 28 then I still get the old "internal server error".

Nyholm:
Hm... I guess it depends a bit on what you are doing.. Here I do a HTTP request to a non-responsive server: 

Screenshot 2020-07-27 at 11 05 24

mnapoli:
> Hm.. This does not work as expected.

Should we merge this PR in the end? I'm not sure what to conclude 😄  but it does look good to me.

Nyholm:
This looks good and it makes sense. But it does not work 100% of the time. Even if I reduce the value to 20s it still does not work 100% of the time. 

It all depends on what the php script is doing. 

I’m honestly not sure what the best way forward is.

mnapoli:
OK thanks for the explanation. If it improves things without any downside, it's fine even if it's not 100% perfect.

Let's merge this 👍  Thanks!

deleugpn:
php execution time is only counted while doing php things. Any time php gets "blocked" to wait on external resources (db, curl, file system), the clock stops. Api Gateway will timeout obn flat 30 seconds.

Nyholm:
I know that `sleep()` does not effect maximum execution time. But is that also true when we wait for IO... 

Hm. Is there a better way to do this then?

mnapoli:
Oh good point!

Nyholm:
We should maybe use PHP-FPM's config: request_terminate_timeout

oppiansteve:
Laravel uses `pcntl_signal` / `pcntl_alarm` for its worker timeouts, if that helps

PR #702, opened 2020-07-28
Remove `enable-maintainer-zts`

Nyholm:
Im out on very deep waters at the moment. 

I remove `--enable-maintainer-zts` because we are not using pthreads in php 7.4. I dont think it is common that one wants to run parallel task **in** Lambda. You spin up multiple Lambdas and execute the jobs in a single thread. 

This will technically be a performance improvement. Im not sure how much.

### DIFF
--- a/runtime/base/php-72.Dockerfile
+++ b/runtime/base/php-72.Dockerfile
@@ -36,7 +36,6 @@ WORKDIR  ${PHP_BUILD_DIR}/
 # --enable-option-checking=fatal: make sure invalid --configure-flags are fatal errors instead of just warnings
 # --enable-ftp: because ftp_ssl_connect() needs ftp to be compiled statically (see https://github.com/docker-library/php/issues/236)
 # --enable-mbstring: because otherwise there's no way to get pecl to use it properly (see https://github.com/docker-library/php/issues/195)
-# --enable-maintainer-zts: build PHP as ZTS (Zend Thread Safe) to be able to use pthreads
 # --with-zlib and --with-zlib-dir: See https://stackoverflow.com/a/42978649/245552
 # --enable-opcache-file: allows to use the `opcache.file_cache` option
 #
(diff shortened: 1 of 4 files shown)


Nyholm:
I tested this on a small app. 

Before, 6 request average: 12,48ms
After, 6 request average: 11,30ms

### Possible side effects: 

If one have an extension that uses ZTS then it will not work anymore. That is why I updated the extensions: https://github.com/brefphp/extra-php-extensions/pull/46

One can also not run pthreads and parallel.

Nyholm:
I decided to remove ZTS support (and pthreads) from PHP 7.2 and 7.3 too. It makes maintenance easier and it avoids confusion when somebody updates from 7.3 to 7.4. 

I suggest to label this as a BC breaking change. It requires a new minor version.

mnapoli:
Hi, thanks for the PR, I agree it makes sense.

> I suggest to label this as a BC breaking change. It requires a new minor version.

Yep, might be worth jumping to 1.0, that's something I want to reach during the summer (https://github.com/brefphp/bref/milestone/3). But that means we'll have to wait a little before merging 😕 

(FYI for PHP 8 I disabled that flag)

We could also remove PHP 7.2 support for 1.0, I'll open a separate issue for that.

Nyholm:
Thank you for the review. 

When is the estimated release of 1.0? If it is too far in the future we maybe should consider 0.6.0.

mnapoli:
There is no plan/reason at the moment for a 0.6, the next will be 1.0. I'm not sure when, as I said I'm aiming for this in the summer.

Is the ZTS thing blocking something? To me it just seems like an optimization for maintenance (and maybe performances), right?

Nyholm:
If there will be a 1.0 around the end of the summer then there is no issue. I wasn’t sure if it would be closer to the end of the year or not. 

No, there is no blockers. It is just a performance boost.

mnapoli:
Hey @Nyholm, I've created a `v1.0` branch. Could you change the target branch for this PR? That way we can merge everything in that branch.

Thanks!

Nyholm:
Lovely. 

Thank you

mnapoli:
Thanks!

PR #708, opened 2020-08-07
Add environment for specifying php-fpm config file

Nyholm:
To be able to specify request_terminate_timeout

Related to #707


mnapoli:
Sorry but I'd rather avoid adding another option for now. Customizing this is doable via a layer and I have a few ideas about other solutions for this in the near future.

Issue #713, opened 2020-08-11
Error handling


Nyholm:
This is a follow up from #707 and is related to #709

After trying out #707 again I found something strange. I've got two routes, one that times out with `sleep(40)` and one normal route. My lambda function has a timeout of 16s and PHP-FPM of 15s.

T0: Access normal route,
T1: Get a response
T2: Access route with sleep
T12: (internal server error) Lambda times out with and my application logs are visible. 
T20: Access normal route
T21: (internal server error) hollodotme\\FastCGI\\Exceptions\\WriteFailedException Failed to write request to socket [broken pipe]" (hollodotme/fast-cgi-client/src/Sockets/Socket.php on line 385)

So at T12 I get a `hollodotme\\FastCGI\\Exceptions\\ReadFailedException Stream got blocked, or terminated` exception. After that I cannot write to PHP-FPM anymore..

> fwrite(): send of 3512 bytes failed with errno=32 Broken pipe in /var/task/vendor/hollodotme/fast-cgi-client/src/Sockets/Socket.php on line 385

@hollodotme do you know if this is correct? 

-------

Another thing I find weird is that my "timout" happens after 10s. I would expect it to be 15 s. 

<img width="2572" alt="Screenshot 2020-08-11 at 13 53 33" src="https://user-images.githubusercontent.com/1275206/89894604-a2dd7f80-dbda-11ea-84b7-0f7b7714a394.png">

mnapoli:
In those cases, Bref should restart PHP-FPM to restart from scratch (https://github.com/brefphp/bref/blob/master/src/Event/Http/FpmHandler.php#L58-L62). It seems it's not happening here, can you confirm?

Nyholm:
The Lambda function is never crashing, Ie, we are never running `FpmHandler::start()` again. 

I think the socket write error is because PHP-FPM's child died. PHP-FPM spins up a new one but we still try to write to the old child.

Nyholm:
Im seeing strange effects... 

When I use `request_terminate_timeout = 15s` it seams like it behaves like:  "If the PHP-FPM child is idle for 15 seconds, then we will cause Socket write errors".

Nyholm:
I use https://github.com/Nyholm/bref-overwrite-layer and I cannot come to any other conclusion than what I've stated above. 

Because request_terminate_timeout is not acting as we expected then I suggest to revert it.

hollodotme:
[...]

PR #714, opened 2020-08-11
Remove request_terminate_timeout = 27s

Nyholm:
Im not happy to say that I think we should revert #707. 

See https://github.com/brefphp/bref/issues/713


mnapoli:
👍  no problem, thanks for investigating!

Nyholm:
Thank you for merging.

PR #723, opened 2020-08-22
perf: introduce lightweight layers

### LINKED ISSUE #722: Add a lightweight/minimal layer version?
Hi,

as always deciding about what to put or not into php during compile time is a matter of arbitrary decisions.
Event if I'm comfortable with most of them, I wonder if it would be a good idea to create more layer variants, especially a minimal one.

I've cloned this repo and made a quick test on php-74-fpm.
By removing lot of things I obtain a **20% smaller layer**.

### Before

```sh
bash-4.2$ du -sh /opt/bref
68M     /opt/bref
bash-4.2$ du -sh /opt/bref/bin/php 
15M     /opt/bref/bin/php
``` 

### After

```sh
bash-4.2$ du -sh /opt/bref
55M     /opt/bref
bash-4.2$ du -sh /opt/bref/bin/php 
14M     /opt/bref/bin/php
``` 

I've also seen that both shared (*.so) and static (*.a) extensions versions are in the produced layer:
```sh
bash-4.2$ ls -lha /opt/bref/lib/php/extensions/no-debug-non-zts-20190902/
total 7.6M
drwxr-xr-x 2 root root  4.0K Aug 19 08:59 .
drwxr-xr-x 3 root root  4.0K Aug 19 08:58 ..
-rw-r--r-- 1 root root  557K Aug 19 08:59 apcu.so
-rwxr-xr-x 1 root root  1.2M Aug 19 08:59 intl.a
-rwxr-xr-x 1 root root  1.6M Aug 19 08:58 intl.so
-rwxr-xr-x 1 root root 1008K Aug 19 08:59 opcache.a
-rwxr-xr-x 1 root root  590K Aug 19 08:58 opcache.so
-rwxr-xr-x 1 root root   50K Aug 19 08:59 pdo_mysql.a
-rwxr-xr-x 1 root root   34K Aug 19 08:58 pdo_mysql.so
-rwxr-xr-x 1 root root   77K Aug 19 08:59 pdo_pgsql.a
-rwxr-xr-x 1 root root   54K Aug 19 08:58 pdo_pgsql.so
-rw-r--r-- 1 root root  2.6M Aug 19 08:59 redis.so
```

I have tested yet but I'm not sure we also need them? :thinking: 
If it's the case the cleanup script could remove them with at least 2M gain.

I haven't deployed this layer nor tested on true lambda invocations yet but as the produced layer is smaller I guess performance should be better (especially cold starts).

shouze:
Fixes #722 

## Why?

1. Because light layers reduce execution time (especially cold starts), between 5 to 15% depending of the kind of app and lambda allocated memory.
2. Because with extra extensions it seem useless to have as many extensions provided by default in the base layer.

## What

- Better clean process
- Removal of following extensions: mongodb, redis, imagick. From now use extras instead
- Removal of: soap, xsl, gd, pthread (and zts)
- Adoption of new GA [Amazon Linux 2 images that will become mandatory at the end of this year](https://aws.amazon.com/fr/blogs/compute/migrating-aws-lambda-functions-to-al2/).

## Notes:
1. Supersedes #702 
2. Find redis & mongob extra exts in https://github.com/brefphp/extra-php-extensions/pull/51

### DIFF
--- a/docs/environment/php.md
+++ b/docs/environment/php.md
@@ -75,7 +75,6 @@ Bref strives to include the most common PHP extensions. If a major PHP extension
         <li><a href="http://php.net/manual/en/book.pcre.php">pcre</a></li>
         <li><a href="http://php.net/manual/en/book.PDO.php">PDO</a></li>
         <li><a href="http://php.net/manual/en/book.pdo_sqlite.php">pdo_sqlite</a></li>
-        <li><a href="http://php.net/manual/en/ref.pdo-mysql.php">pdo_mysql</a></li>
         <li><a href="http://php.net/manual/en/book.Phar.php">Phar</a></li>
         <li><a href="http://php.net/manual/en/book.posix.php">posix</a></li>
         <li><a href="http://php.net/manual/en/book.readline.php">readline</a></li>
@@ -87,7 +86,6 @@ Bref strives to include the most common PHP extensions. If a major PHP extension
         <ul>
         <li><a href="http://php.net/manual/en/book.SimpleXML.php">SimpleXML</a></li>
         <li><a href="http://php.net/manual/en/book.sodium.php">sodium</a></li>
-        <li><a href="http://php.net/manual/en/book.soap.php">SOAP</a></li>
         <li><a href="http://php.net/manual/en/book.sockets.php">sockets</a></li>
         <li><a href="http://php.net/manual/en/book.SPL.php">SPL</a></li>
         <li><a href="http://php.net/manual/en/book.sqlite3.php">sqlite3</a></li>
@@ -96,7 +94,6 @@ Bref strives to include the most common PHP extensions. If a major PHP extension
         <li><a href="http://php.net/manual/en/book.xml.php">xml</a></li>
         <li><a href="http://php.net/manual/en/book.xmlreader.php">xmlreader</a></li>
         <li><a href="http://php.net/manual/en/book.xmlwriter.php">xmlwriter</a></li>
-        <li><a href="http://php.net/manual/en/book.xsl.php">xsl</a></li>
         <li><a href="http://php.net/manual/en/book.zlib.php">zlib</a></li>
         </ul>
       </td>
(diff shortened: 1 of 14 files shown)


shouze:
> Thanks for the PR.
> 
> Let's wait on [brefphp/benchmarks#2](https://github.com/brefphp/benchmarks/pull/2) to have the final numbers.
> 
> FYI this PR will be hard to merge as it mixes different things. But it's still interesting to see the performance improvements we can get and see what we can trim from the current layers to gain significant improvements.

in fact no big gains by making the layer thinner. So I suggest to split in smaller PRs, each focused on 1 simple goal:
1. Upgrade libraries (curl, sodium, openssl, xml2, zip, postgres, oniguruma)
2. Move redis, imagick & gd from base layer, thus maintain them as extra extensions
2. Support Amazon Linux 2. About this one I wonder if it's not important for the transition in the coming 3 months to support both Amazon Linux 1 and Amazon Linux 2, so having twice the layers from now? :thinking:. This is a lot of work so other option is to make a last release of layers supporting Amazon Linux 1 and just adopt Amazon Linux 2 but publishing them with a new prefix.

WDYT?

mnapoli:
Hey @shouze, an update on that PR: I have created a `v1.0` branch (https://github.com/brefphp/bref/tree/v1.0). We can push in that branch whatever will be in the v1.0 version (see the milestone https://github.com/brefphp/bref/milestone/3), include BC breaks.

If you want, you can split out this PR (for easier merging), these changes would be welcome:

- migrating to AL2 (v1.0 should support AL2, not AL1, that will be much simpler than maintaining the 2)
- remove imagick & gd from the main runttimes
- upgrade libraries (ideally let's do this after upgrading to AL2, also there may be some libraries that we currently build manually that could be removed and we could use the system one, let's see)

Let me know what you think.

shouze:
> Hey @shouze, an update on that PR: I have created a `v1.0` branch ([https://github.com/brefphp/bref/tree/v1.0](https://github.com/brefphp/bref/tree/v1.0?rgh-link-date=2020-09-09T07%3A51%3A14Z)). We can push in that branch whatever will be in the v1.0 version (see the milestone [brefphp/bref/milestone/3](https://github.com/brefphp/bref/milestone/3)), include BC breaks.
> 
> If you want, you can split out this PR (for easier merging), these changes would be welcome:
> 
> * migrating to AL2 (v1.0 should support AL2, not AL1, that will be much simpler than maintaining the 2)
> * remove imagick & gd from the main runttimes
> * upgrade libraries (ideally let's do this after upgrading to AL2, also there may be some libraries that we currently build manually that could be removed and we could use the system one, let's see)
> 
> Let me know what you think.

I'm fully aligned with that, I can dedicate some time by end of this week for some PRs, let's :rocket:

shouze:
closing this one as it's now done in many smaller PRs :wink:

PR #724, opened 2020-08-25
Add getter for ReceiptHandle in SQSRecord

marcusirgens:
When executing [DeleteMessage](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_DeleteMessage.html) against SQS (which AFAIK should be done when you successfully process a message), you must provide the ReceiptHandle that SQS provides as input to the Lambda function.

It's both documented [in the Lambda docs](https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html), and there is already a ReceiptHandle key in the sample data used in the existing Bref tests. :) 

Because getting this piece of data and using it is a very common operation, I think it deserves its own getter.

### DIFF
--- a/src/Event/Sqs/SqsRecord.php
+++ b/src/Event/Sqs/SqsRecord.php
@@ -51,6 +51,14 @@ final class SqsRecord
     }
 
     /**
+     * Returns the receipt handle, the unique identifier for a specific instance of receiving a message.
+     */
+    public function getReceiptHandle(): string
+    {
+        return $this->record['receiptHandle'];
+    }
+
+    /**
      * Returns the record original data as an array.
      *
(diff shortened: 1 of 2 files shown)


mnapoli:
Thanks, sounds perfectly reasonable.

> When executing DeleteMessage against SQS (which AFAIK should be done when you successfully process a message)

FYI SQS automatically deletes messages when the Lambda successfully finishes.

However, manually deleting messages can be useful if you process messages in batch, instead of 1 message per invocation.

marcusirgens:
> FYI SQS automatically deletes messages when the Lambda successfully finishes.

Yeah, I assumed so actually. I do tend to prefer the explicit deletion, though. Thanks!

PR #728, opened 2020-08-28
SSM parameter SecureString recommendation

rflavien:
I followed the documentation provided by Bref explaining SSM parameter String usage for secrets.

But that is not following AWS recommendation _(linked in Bref's documentation)_
image

I found that Bref can follow this recommendation with 
image

I suggest this documentation's update for others AWS and serverless framework rookies like me.


mnapoli:
Hi, thanks for the PR!

I think this may be duplicating #677?

PR #729, opened 2020-08-30
Fix typo

fabpot:

### DIFF
--- a/couscous.yml
+++ b/couscous.yml
@@ -90,9 +90,9 @@ menu:
                 text: Custom domains
                 url: /docs/environment/custom-domains.html
                 title: Configuring custom domain names for HTTP applications
-            performances:
-                text: Performances
-                url: /docs/environment/performances.html
+            performance:
+                text: Performance
+                url: /docs/environment/performance.html
                 title: Performance tuning and optimizations
     -   section: Frameworks
         items:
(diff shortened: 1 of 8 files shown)


couscous.yml
```diff
@@ -90,9 +90,9 @@ menu:
                 text: Custom domains
                 url: /docs/environment/custom-domains.html
                 title: Configuring custom domain names for HTTP applications
-            performances:
-                text: Performances
-                url: /docs/environment/performances.html
+            performance:
+                text: Performance
+                url: /docs/environment/performance.html
```
mnapoli, line 95: Renaming the file would break existing links (outside of bref.sh itself), I think it might be better to avoid changing the URL/file name.

mnapoli (commented):
Thanks @fabpot!

I had no idea that performance/performances might be a bit like "information". Though I did a quick search and it seems that "performances" (plural) is fine too (e.g. https://english.stackexchange.com/questions/367488/plural-of-performance). Do you have more info?

In this case I often talk about multiple performance aspects:

- cold starts
- CPU speed
- runtime overhead

which may make sense to use the plural term.

fabpot:
Using "performances" (plural) is fine when talking about a show (artists are performing, they are doing performances). But when it comes to the speed/performance of some code, then the word is uncountable.

mnapoli:
Thanks, I'll merge but will later adjust the URLs to avoid breaking existing links.

PR #730, opened 2020-09-01
Add a `bref layers` command

mnapoli:
Replaces #664

This adds a new `bref layers us-east-1` command that returns the version of the layers to use for the current Bref version.

It will make it easier for people not using `serverless.yml` to keep Bref up to date.

### DIFF
--- a/bref
+++ b/bref
@@ -455,6 +455,6 @@ $app->command('layers region', function (string $region, SymfonyStyle $io) {
     ], $array);
 
     return 0;
-})->descriptions('Create a short URL on bref.dev');
+})->descriptions('Displays the versions of the Bref layers');
 
 $app->run();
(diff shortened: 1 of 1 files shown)


bref
```diff
@@ -435,4 +435,26 @@ $app->command('bref.dev [--profile=] [--stage=]', function (?string $profile, st
...
+        ];
+    }
+    $io->table([
+        'Layer',
+        'Version',
+        'ARN',
+    ], $array);
+
+    return 0;
+})->descriptions('Create a short URL on bref.dev');
```
Fabex, line 458: bad copy past I thing

Fabex (commented):
this command only display layers but can not be use as ci test before a merge or a deploy to check if we use the good layer for the current bref in projet

bref
mnapoli: Thanks!

mnapoli:
No indeed, it cannot be used in CI. That's a first step towards maybe a more automated solution, but as I said in #664 I'd rather avoid to couple to SAM specifically for now.

Fabex:
Ok I understand

PR #734, opened 2020-09-03
Added 25ms delay to ensure concurrent invocation

peppeocchi:
When trying to warmup a function, it needs to delay a bit the response to ensure concurrent invocations.
When I invoke the function with "{warmer: true}" as input (as described in the [documentation](https://bref.sh/docs/runtimes/http.html#cold-starts)) it takes less than 1ms for the function to end. Without any delay it's very unlikely to get near the desired concurrency. Or even to get any concurrent executions.

For example, by using different warmup plugins or even a simple function that invokes the bref php function, over 5 invocations if you're lucky it will warm 1 more function. Even trying to set a concurrency of 60 you won't get more than 5 warmed up function.

By adding a small delay (25ms), you get more chances to get close to your desired concurrency. Anyway you get billed for a minimum of 100ms, so adding this delay won't impact costs.

Edit:
this is a screenshot of my invocations with {warmer: true}

image

### DIFF
--- a/src/Event/Http/HttpHandler.php
+++ b/src/Event/Http/HttpHandler.php
@@ -14,6 +14,8 @@ abstract class HttpHandler implements Handler
     {
         // See https://bref.sh/docs/runtimes/http.html#cold-starts
         if (isset($event['warmer']) && $event['warmer'] === true) {
+            // Delay the response to ensure concurrent invocation
+            usleep(25000);
             return ['Lambda is warm'];
         }
 
--- a/src/Event/Http/HttpHandler.php
+++ b/src/Event/Http/HttpHandler.php
@@ -14,9 +14,10 @@ abstract class HttpHandler implements Handler
     {
         // See https://bref.sh/docs/runtimes/http.html#cold-starts
         if (isset($event['warmer']) && $event['warmer'] === true) {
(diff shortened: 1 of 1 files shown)


mnapoli:
Thanks, that's a good idea!

The only downside I can see is that it makes the lambda instance "busy" for 25ms. If there is no traffic, that's not a problem.

But let's imagine you want 5 warmed Lambda instances, and the current traffic already uses 5 Lambdas. If you make all 5 instances busy at the same time for 25ms, then there's a risk that it boots additional instances, causing more cold starts. (or even throttle the lambda if reaching make concurrency)

Maybe a compromise could be to set a delay of 5 or 10ms? (just to lower the risk) Would that delay be sufficient to warm all the instances?

peppeocchi:
@mnapoli I know what you mean, but I think when using the warmup there is always a chance to get cold starts on real user requests. From the logs I have from the warmup function, it takes an average of 100-150ms to run a concurrency of 5.

Maybe it would be better a configurable delay? This way it could be adjusted depending on the needs.

mnapoli:
Thanks, I've merged the first commit in 1.0 and adjusted the delay to 10ms.

I try to avoid having many options, let see if it improves things first.

Thanks!

PR #736, opened 2020-09-04
Fix fpm dev gateway

shouze:
- use lightweight (alpine based) docker image :whale:  :leaves: 
- remove duplicate configuration directive :100: 
- add fastcgi buffer configuration to avoid some unexpected local development errors (logging related) :bomb:

### DIFF
--- a/runtime/layers/web/Dockerfile
+++ b/runtime/layers/web/Dockerfile
@@ -1,4 +1,4 @@
-FROM nginx:latest
+FROM nginx:alpine
 COPY default.conf /etc/nginx/conf.d/default.conf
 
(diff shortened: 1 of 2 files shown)


shouze:
ping @Guillaume-Rossignol

mnapoli:
> add fastcgi buffer configuration to avoid some unexpected local development errors (logging related)

I don't know what that means, without more explanation I'll make the choice to trust your judgement ^^

Thanks!

tyx:
> > add fastcgi buffer configuration to avoid some unexpected local development errors (logging related)
> 
> I don't know what that means, without more explanation I'll make the choice to trust your judgement ^^
> 
> Thanks!

FTR we had issue with too big entry written in stderr. 

See https://stackoverflow.com/questions/23844761/upstream-sent-too-big-header-while-reading-response-header-from-upstream for more information

PR #740, opened 2020-09-08
Enable HTTP/2 for CURL

### LINKED ISSUE #727: Enable HTTP/2 for CURL
I have a lumen based API using the `php-74-fpm` layer as well as a custom layer to add sqlsrv extensions. When using the curl extension to make a request to Apple's APNs, I am receiving the following error:

`production.INFO: APNS request response: @@uUnexpected HTTP/1.x request`

The cause seems simple enough, Bref's provided environment does not have HTTP/2 enabled for the curl extension. According to [this](https://stackoverflow.com/a/34609756/1971250) Stackoverflow answer, there is a compile time option to enable HTTP/2.

This link may also be helpful: [curl http2 docs](https://curl.haxx.se/docs/http2.html)

If anyone has been able to enable this functionality via a custom layer, I would appreciate any guidance.

shouze:
Fixes #727


runtime/base/base.Dockerfile
```diff
@@ -198,6 +198,37 @@ RUN set -xe; \
 RUN set -xe; \
     cmake  --build . --target install
 
+###############################################################################
+# LIBNGHTTP2 Build
```
mnapoli, line 202:
```suggestion
# LIBNGHTTP2 Build
# This adds support for HTTP 2 requests in curl.
# See https://github.com/brefphp/bref/issues/727 and https://github.com/brefphp/bref/pull/740
```

mnapoli (approved):
Looks good, I added an inline suggestion to improve maintenance.

runtime/base/base.Dockerfile
```diff
@@ -244,7 +277,8 @@ RUN set -xe; \
     --enable-cookies \
     --with-gnu-ld \
     --with-ssl \
-    --with-libssh2
+    --with-libssh2 \
+    --with-nghttp2=${INSTALL_DIR}
```
shouze, line 281: @mnapoli Ooops I also forgot to include that. It's all good now & ready to merge :angel:

mnapoli: Could that be replaced by `--with-nghttp2`?

Because all the other flags don't have the path, I think this is because of the `--prefix=${INSTALL_DIR}` option above.

shouze: This is a good question, and I've tried but it fail without the directory path so I had no other choice :sob: 

The only thing I haven't tried yet is to set some env vars, namely `CMAKE_INCLUDE_PATH` and `CMAKE_LIBRARY_PATH` as explained in https://github.com/curl/curl/pull/922#issuecomment-244625230.
Maybe it can work with that.

shouze: Ok sorry I've double checked and it's OK we can remove it indeed :angel:

mnapoli:
Thanks!

shouze:
:wink: @atbreuer11 will be happy soon!

atbreuer11:
@shouze Your amazing! - looking forward to testing this out.

PR #749, opened 2020-09-18
doc: remove pthread mentions

shouze:
Related to #702, ping @Nyholm 

Also, downloaded blackfire extension is a zts version... I guess that we should remove the zts suffix from the download link, no? :thinking:


Nyholm:
I’m waiting for an alpha tag. Then I’ll merge the pr in the bref php-ext repo

mnapoli:
Thanks!

PR #752, opened 2020-09-21
Fix issue 636

### LINKED ISSUE #636: Deprecate the built-in imagick extension in favor of the separate extension
Following https://github.com/brefphp/extra-php-extensions/pull/31, in order to support many formats and things in imagick, we have to include a lot of stuff in there.

That makes for a heavy layer, [around 14Mb](https://github.com/brefphp/extra-php-extensions/pull/31#issuecomment-619588633) of extra files.

Here is what is planned:

- deprecate the built-in imagick extension
- build that extension as a separate extension in https://github.com/brefphp/extra-php-extensions
- eventually remove imagick from the main PHP layers in the next major versions

shouze:
fixes #636. As it's now available as an extra extension.


mnapoli:
Wrong target branch?

shouze:
ho well... yes closed by https://github.com/brefphp/bref/pull/753

PR #762, opened 2020-09-25
add dynamo db cache bundle to readme

RikudouSage:
We talked on Reddit with @mnapoli and he said that I should add the bundle into the Readme, so I'm doing that.

Let me know if that's OK and what I can improve.

### DIFF
--- a/docs/frameworks/symfony.md
+++ b/docs/frameworks/symfony.md
@@ -80,6 +80,67 @@ Since [the filesystem is readonly](/docs/environment/storage.md) except for `/tm
     }
 ```
 
+## Using cache
+
+As mentioned above, the filesystem is readonly so if you need persistent cache you need to store it somewhere else.
+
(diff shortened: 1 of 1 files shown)


mnapoli:
Thank you!

That looks good, but I'd rather not have the details of the instructions in that page: if they change in the bundle that means that this documentation will not be up to date. So maybe it's better to add a paragraph and link to direct users to the bundle, and leave them to read the README of the bundle?

RikudouSage:
That's why I included only the basic example, these options are the core so they're staying forever, I thought it would be better to have at least some example of usage. What do you think?

mnapoli:
Unfortunately I've learned from experience that things eventually change. Let's keep it a link. Could you add it at the end of the page, after this section:

CleanShot 2020-09-28 at 10 28 03

mnapoli:
Thanks!

PR #764, opened 2020-10-02
bump PHP versions

chapeupreto:


mnapoli:
Thanks, TBH I don't really want to rebuild layers every week (it's a very very long and manual process). Yes this could be automated, but it's a long work to automate that.

Do we need the absolute latest versions every time? I'm perfectly fine to update when there are significant fixes, but most of the time this brings no benefit to anyone.

Nyholm:
I've automated things a bit in https://github.com/brefphp/extra-php-extensions/blob/master/.github/workflows/blackfire.yml

Could be nice to look at if one is starting with the automation

mnapoli:
Yep that should be automated some day 😅

andremohrmann:
This would be much appreciated since we are required to run the latest PHP version asap due to internal security guidelines.

mnapoli:
@andremohrmann I can invite you to get in touch privately at matthieu@null.tc, I can provide support for such requirements via [Bref Enterprise](https://bref.sh/#enterprise).

mnapoli:
Closing, I'm working on Bref 1.0 right now and I'm upgrading things there.

Thanks!

PR #769, opened 2020-10-06
Updating query string rebuilding in HttpRequestEvent

DaMitchell:
Updated the rebuilding of the query string so that parameter names such as `foo.bar` will not be converted to `foo_bar` by `parse_str`.


DaMitchell:
@mnapoli any thoughts/feedback on this?

src/Event/Http/HttpRequestEvent.php
```diff
@@ -286,4 +283,25 @@ private function extractHeaders(): array
...
+        // to find it when replacing names with the regexp below.
+        $queryString = str_replace('%5B', '[', $queryString);
+
+        $queryString = preg_replace_callback(
+            '/(^|(?<=&))[^=[&]+/',
+            function ($key) {
+                return bin2hex(urldecode($key[0]));
+            },
+            $queryString
+        );
```
mnapoli, line 300: Sorry for the delay, I've been uncertain about this PR but I should have voiced that earlier.

The thing is that I find this fragile and hacky. I'm really afraid merging this will expose us to more problems or edge cases down the line. Is there any way that PHP solves that natively somehow? Or any PHP library out there? I had a quick look at https://github.com/symfony/symfony/pull/37272/files but I'm wondering how other libraries work around this.

DaMitchell: No problem. I did see the PHP League has a package that can deal with URI extraction, https://github.com/thephpleague/uri-query-parser#extracting-php-variables. I didn't know if you would be in favour of using another library for it or have an internal solution.

mnapoli: Oh that library looks good, and very targeted to that specific problem. Let's give it a try yes (hopefully they cover every use case 😅)

Thanks for digging into this!

DaMitchell: No worries, Ill give it an update on Monday.

DaMitchell: So it looks like that library is no longer being maintained, but is now part of https://github.com/thephpleague/uri-components. 

To use it you can do `(Query::createFromRFC1738($queryString))->params()` and for the issue I am experiencing the package works, but when it parses `colors[][]` instead of producing:
```
[
  "colors": [
    ["red"],
    ["blue"]
  ]
]
```
It parses it as:
```
[
  "colors": [
    "red",
    "blue"
  ]
]
```

The Symfony solution you link to above works for all the tests currently, but that hasn't been released and looks to be released in 5.2 and I am not sure what impact that would have on people running a Symfony app but on versions lower than 5.2.

This is a tricky one 😅

mnapoli: Indeed 😕 

We indeed can't require Symfony components with version 5.2, this would be a mess for users. I think the "best" approach would be to copy/paste the content of their method into Bref, and also copy/paste their tests. At least until we can remove that eventually.

Not ideal :/ But it would work and we wouldn't have to reinvent the wheel.

Note that when copying code from other open-source projects (if you want to go that way), make sure to include the original license and credit the original authors, for example: https://github.com/brefphp/bref/blob/43ba5fb589c719bb414905ad8ce0235619aeae10/tests/Server.php#L3-L23

mnapoli:
Hey, since this pull request has been stuck for a while I will close it for now.

Please reopen one if you manage to make to update it.

PR #772, opened 2020-10-09
Remove Socket Timeout from FPM Layer

deleugpn:
After working on #770 and #771, @mnapoli had an amazing idea of simply getting rid of the timeout altogether. It makes total sense:

- API Gateway has a timeout of 29 seconds and the socket used to have it at 30 seconds. The Socket timeout was never being reached anyway.
- AWS Lambda has a timeout of 15 minutes at this time, we can use that same number here as it would be impossible to have any lambda execution longer than that.

No need for custom environment variable control or weird hacks on the bootstrap process. I tested this on API Gateway as well as ALB and both work fine with FPM.


mnapoli:
Thank you!

PR #773, opened 2020-10-09
Add typed ApiGateway websocket events

danniehansen:
Work in progress. This adds support for typed ApiGateway websocket events.

I'm unsure of how this should be namespaced. Rest of the typed events is in matching namespaces for their service name, so I thought to do the same for this.

Example:

handler/websocket.php
```
<?php

use Bref\Context\Context;
use Bref\Event\ApiGateway\WebsocketEvent;
use Bref\Event\Http\HttpResponse;
use Bref\Event\ApiGateway\WebsocketHandler;

class MyHandler extends WebsocketHandler
{
    public function handleWebsocket(WebsocketEvent $event, Context $context): HttpResponse
    {
        echo json_encode($event->toArray()) . "\n";
        echo json_encode($context->jsonSerialize()) . "\n";

        echo 'Route key: ' . $event->getRouteKey() . "\n";
        echo 'Event type: ' . $event->getEventType() . "\n";
        echo 'Body: ' . $event->getBody() . "\n";
        echo 'connectionId: ' . $event->getConnectionId() . "\n";
        echo 'domainName: ' . $event->getDomainName() . "\n";
        echo 'stage: ' . $event->getStage() . "\n";
        echo 'apiId: ' . $event->getApiId() . "\n";
        echo  "\n\n";

        return new HttpResponse($event->getEventType());
    }
}

return new MyHandler();
```

serverless.yml
```
service: websocket-test
provider:
    name: aws
    region: eu-west-1
    stage: dev
    runtime: provided
plugins:
    - ./vendor/bref/bref
functions:
    websocket:
        handler: handler/websocket.php
        layers:
            - ${bref:layer.php-74}
        events:
            - websocket: $connect
            - websocket: $disconnect
            - websocket: $default
```

What's missing?

Feedback.


danniehansen:
This solves one side of the puzzle. Incoming requests. Outgoing requests is possible to make from the parsed meta data of the event. The request it-self for posting to an existing connection is rather straight forward - however, I'm not sure how this should be abled internally in bref?

https://stackoverflow.com/questions/58572046/how-to-send-a-message-to-specific-connectionid-to-aws-api-gateway-websockets

Do we already have utility for making network requests? & should this even be part of bref?

Once all is settled and (if all goes well) this get's merged I'd be looking into https://laravel.com/docs/8.x/broadcasting and see if it can be somehow implemented into https://github.com/brefphp/laravel-bridge as a broadcast connection driver.

deleugpn:
Inside `src/Lambda` there is one `SimpleLambdaClient`. It's not something bref needs to work, except the UI invocation but it is kind of convenient. Maybe it could have a convenient `SimpleWebsocketClient` class tbat offers a simple communication back (accepting the necessary parameters).
I would recommend it on a separate PR though, as to focus this one on Incoming connection and keep it simple.

danniehansen:
@deleugpn Sounds good. I'll look into this in a separate PR if this one get's accepted.

src/Event/ApiGateway/ApiGatewayHandler.php
```diff
@@ -0,0 +1,20 @@
...
+ * Handles ApiGateway events.
+ */
+abstract class ApiGatewayHandler implements Handler
+{
+    abstract public function handleWebsocket(WebsocketEvent $event, Context $context): WebsocketResponse;
+
+    /** {@inheritDoc} */
+    public function handle($event, Context $context)
+    {
+        return $this->handleWebsocket(new WebsocketEvent($event), $context)->toApiGatewayFormat();
```
mnapoli, line 18: Is a websocket response only a status code?

Because I'm wondering if we need the `WebsocketResponse` class at all? (we could return the status code a `int`).

One argument for keeping the class would be if there are other fields (not implemented yet), or if there will be (is it the case?).

src/Event/ApiGateway/ApiGatewayHandler.php
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Bref\Event\ApiGateway;
+
+use Bref\Context\Context;
+use Bref\Event\Handler;
+
+/**
+ * Handles ApiGateway events.
+ */
+abstract class ApiGatewayHandler implements Handler
```
mnapoli, line 11: Could be `WebsocketHandler` instead? (to be more precise, and we already are in the `ApiGateway` namespace so no risk of collision in the future.

src/Event/ApiGateway/WebsocketEvent.php
```diff
@@ -0,0 +1,141 @@
...
+    private $event;
+
+    /** @var string */
+    private $routeKey;
+
+    /** @var int */
+    private $eventType;
+
+    /** @var mixed */
+    private $body;
```
mnapoli, line 23:
```suggestion
    /** @var mixed|null */
    private $body = null;
```

That was the property is always initialized (thinking of PHP 7.4+ migration), and this will simplify the getter below (see my next comment)

src/Event/ApiGateway/WebsocketEvent.php
```diff
@@ -0,0 +1,141 @@
...
+    /**
+     * @return mixed|null
+     */
+    public function getBody()
+    {
+        if (isset($this->body)) {
+            return $this->body;
+        }
+
+        return null;
```
mnapoli, line 119: Could be simplified to `return $this->body` if the property is initialized with `null` (my previous comment).

src/Event/ApiGateway/WebsocketEvent.php
```diff
@@ -0,0 +1,141 @@
...
+                $this->eventType = self::EVENT_TYPE_CONNECT;
+                break;
+
+            case 'DISCONNECT':
+                $this->eventType = self::EVENT_TYPE_DISCONNECT;
+                break;
+
+            case 'MESSAGE':
+                $this->eventType = self::EVENT_TYPE_MESSAGE;
+                break;
```
mnapoli, line 87: Why not have `self::EVENT_TYPE_CONNECT` constants (and the other 2) have the string values (`'CONNECT'`, etc.)? That way no need to map events manually to an integer value.

tests/Event/ApiGateway/WebsocketEventTest.php
```diff
@@ -0,0 +1,92 @@
...
+
+use Bref\Event\ApiGateway\WebsocketEvent;
+use Bref\Event\ApiGateway\WebsocketResponse;
+use PHPUnit\Framework\TestCase;
+
+class WebsocketEventTest extends TestCase
+{
+    public function test_connect()
+    {
+        // Arrange
```
mnapoli, line 13: Could you remove these comments? (these would not be inline with the rest of the codebase)

Keep the practice however, that's in general how the rest of the tests are organized.

mnapoli (commented):
I added minor comments, the structure of the PR is perfect, there's not much left to merge, thank you!

mnapoli:
Regarding the `SimpleWebsocketClient`, a PR indeed would be great to get an idea. That's something that makes sense, if it is too big, we'll move it to a separate repository but let's start with the PR!

src/Event/ApiGateway/ApiGatewayHandler.php
danniehansen: @mnapoli You've got a point. Initially I thought that response looked similar to HTTP responses and thus took inspiration in how it handled it's response. What I found how-ever was that forwarding a 'body' in the response broke it and connection received an error. Looking at the documentation of Websockets further I've found that it does support headers: https://docs.aws.amazon.com/apigateway/latest/developerguide/websocket-connect-route-subprotocol.html

Maybe to simplify things we should have the response accept both a return type of int and WebsocketResponse. This way we could add in support for headers in WebsocketResponse and just have the handler wrap int returns in the response class for ApiGateway formatting?This would means that for 9/10 scenarios where you just want to return a status code you just return the int, but still leaves room for future additions to the ApiGateway Websocket API.

src/Event/ApiGateway/WebsocketEvent.php
danniehansen: I agree. Much better approach.

src/Event/ApiGateway/WebsocketEvent.php
danniehansen: @mnapoli The reasoning behind this is to force developers to use the constants rather than passing around strings. I work with a large code base on a daily basis and what we saw when having string values for constants was that developers choose to just pass the string value rather than passing the constant. This meant that once we had to re-iterate on our code base it was near impossible to locate peoples that referenced the constants & resulted it methods being near impossible to refactor. Simply because we had no way of knowing how things was passed around.

Having a integer as a value for the constants sort of forces developers to use constants. Obviously they could still pass around the numeric values - but they're less obvious for what they are and people then tend to use the constant instead.

This was at least my observation on this.

src/Event/ApiGateway/WebsocketEvent.php
danniehansen: Agreed. That's much simpler. Changing here in a minute.

tests/Event/ApiGateway/WebsocketEventTest.php
danniehansen: Sure. Actually copied over the test from DynamoDbEventTest.php - so I thought it was standard practice in the code-base and left them. I'll remove them.

src/Event/ApiGateway/ApiGatewayHandler.php
danniehansen: This obviously also depends on the rest of the code base. If it's more in tune doing string values - then I'll just change it. Better to have same style for each implementation.

src/Event/ApiGateway/ApiGatewayHandler.php
danniehansen: Sure. I see the pattern. Just looked over the other event handlers and they're doing the same thing. I'll align it.

src/Event/ApiGateway/ApiGatewayHandler.php
danniehansen: I've done some further investigation and it appears only `Sec-WebSocket-Protocol` can be specified in the headers object. I'm unable to find much else on it besides the aws docs already forwarded and https://stackoverflow.com/questions/4361173/http-headers-in-websockets-client-api

Updating PR to support int|WebsocketResponse both as return type. Where as WebsocketResponse second constructor argument would be the supported protocols.

danniehansen:
> I added minor comments, the structure of the PR is perfect, there's not much left to merge, thank you!

I've gone over them and applied most of them but also commented on two of them. Appreciate you taking your time with this.

Regarding recent Travis CI failure. It appears PHPStan doesn't like `WebsocketResponse` L26. Due to union types not being supported I felt like this runtime check was needed to ensure proper data types passed. Should these type of runtime data type checks be left out?

src/Event/ApiGateway/ApiGatewayHandler.php
mnapoli: Hi, really sorry, the fact that the response is actually an object with possible additional properties in JS is very helpful: keeping `WebsocketResponse` only makes a lot of sense then (and it's future-proof as well).

Could you drop the `int` then 😅 ? (the less choice there is the better)

tests/Event/ApiGateway/WebsocketEventTest.php
mnapoli: Right sorry I probably missed that in the DynamoDB PR, thanks!

src/Event/ApiGateway/WebsocketEvent.php
mnapoli: Right, but then we would need a `default` case in that switch: but what would we put in there?

If we put an exception, then we are not future-proof (someone will need to patch that method the day a new event is added by AWS). If we silently ignore it, like right now, then we have a non-initialized property and the getter will throw an error (which is not future-proof as well).

I'd rather take the same approach as these event classes: they all expose a `toArray()` method, so that if users want to access data in the event that is not exposed by the class, then they can.

And while I agree with you about not using strings directly, I am also weary of adding a layer of abstraction on top of what AWS provides. Because that's one more thing to maintain, explain, etc. I would agree here it's not a huge deal, but small things like that add up.

src/Event/ApiGateway/ApiGatewayHandler.php
mnapoli: Wait so I had a look around, I'm not sure what to think TBH. The examples I've found online look a lot like a HTTP response ([e.g. this example by AWS](https://github.com/aws-samples/simple-websockets-chat-app/blob/e4566f8a30986e53e90024c92a41ea6f5f863e18/sendmessage/app.js#L39-L45)).

But I'm not sure if that response format is required by API Gateway (and not used by websockets itself), or if a websocket response is actually a HTTP response (which I don't think it is but I don't really know much about websockets).

So I don't know if we should have a custom class for that response, or use a HTTP response class, or do something else 🤔

src/Event/ApiGateway/WebsocketEvent.php
danniehansen: @mnapoli You're right. Now that i look at it again there is no need for adding this extra layer of abstraction on top of the event data. I suppose simple getters will do just fine here. I'll remove the constants and change it to just supply the event type forwarded.

src/Event/ApiGateway/ApiGatewayHandler.php
danniehansen: @mnapoli Hm - that's interesting. I was unable to find anything officially in the documentation documenting the 'body' index. & I'm also not familiar with websockets having the ability to return data in the connect, message or & disconnect events. Maybe this is just their way re-using ApiGateway's existing & powerful API? I'll do some further investigation on this.

Initially I tried supplying a body, but what I experienced was that ApiGateway websocket connect got rejected to due a syntax error. If websocket support statusCode, headers & body just like normal HTTP responses for ApiGateway - then we could as you wrote reuse HttpResponse for websockets. But that being said, I really don't see where 'body' comes into play.

I will do some more investigation on this and come back with my findings.

src/Event/ApiGateway/WebsocketEvent.php
danniehansen: @mnapoli Fixed now. Now it returns the string value - so it's just a simple getter for accessing what ApiGateway forwards in the event 👍

src/Event/ApiGateway/ApiGatewayHandler.php
danniehansen: @mnapoli  Apparently I was mistaken regarding `body` index in response. It's perfectly acceptable and ApiGateway doesn't react to it at all. I've gone ahead and made the change so it now expect a HttpResponse instead of the new WebsocketResponse.

Looked at the [RFC](https://tools.ietf.org/html/rfc6455#page-6) and I see no mention of any possibility of directly giving a response other than a status code for either events. Even sending a 500 status code error on a message event does nothing in the frontend (no exception or 'onerror' trigger). The only event that allows for anything other than a status code to be put as response appears to be the connect event. Within connect event you can specify a few headers (also mentioned in https://tools.ietf.org/html/rfc6455#page-6).

It's very unclear to me why Amazon has even put a `body` index in their example. When looking from a RFC standpoint there is absolutely no way of sending anything back in either event matching a response body. In addition to this it doesn't appear to do anything whether you're responding with a body or not.

I'm conflicted. Utilizing the HttpResponse class does make sense as it appears to use the exact same response structure as for Http requests. But - HttpResponse expects first constructor argument to be a string (not allowing null). This means you'll have to provide a empty string as a body or anything really just to signal a 200 success. At the same time - supporting HttpResponse makes the event handling more in tune with AWS's own examples 😕 

-

Update:

Asked AWS regarding this and got the following:
```
Them: So generally speaking, it is more used for Data Transformations in the Integration Response. If for example you're using the same Lambda Function to provide responses on both a REST API and a Websocket API, you can use the Data Mapping templates to take the body being returned from the function and map it into a part of the response supported by the protocol.

Me: Ah. So it's a way of enabling the same lambda function to respond to regular Http/Rest requests from ApiGateway?

Them: That is one example use case, I personally don't have a wide range of experience with the Websocket API Gateways, but I can take you through the Data Transformations to help you better understand the implementation.

Me: That would be very helpful for understanding it's possible use-cases. Do you have an example of this? Just to get the idea?

Them: So this page of the docs has a couple of examples for Integration Request mappings, but the same approach can be used to remap data in an Integration Response.

https://docs.aws.amazon.com/apigateway/latest/developerguide/websocket-api-data-mapping.html

When dealing with REST APIs you can configure these within the AWS Console, however with a Websocket API you have to use SDK/CLI to implement

And follow this, plus the Data mapping and WebSocket mapping template reference pages underneath, you'll be able to see that it is quite powerful, and will allow you to re-use existing Lambdas/HTTP Servers and offload the work of re-working data for Websockets onto API gateway instead of having to build another method.
```

So all in all i suppose they're one and the same (Rest, Http & Websocket). Which means it would make sense to utilize the same HttpResponse, right?

danniehansen: @mnapoli Let me know if there is anything i can do on this. I know there is a lot of dialog / discussion on this. I got the time to adjust any changes you'd like on this if need be.

mnapoli: Very sorry about the delay, I was unsure about merging something I'm unsure about.

Look, we have one life, and worst case scenario we got the return type wrong: we can always remove it later and support 2 kind of returned value. v1.0 is also around the corner, so there's that. The rest of the codebase is not impacted.

Let's do it!

mnapoli:
Thanks a lot @danniehansen for all that work!

mnapoli:
Regarding the `SimpleWebsocketClient` do you plan on adding that in the near future? Or would you prefer if I release a new version of Bref right now?

src/Event/ApiGateway/ApiGatewayHandler.php
danniehansen: @mnapoli No worries at all. Appreciate it!

danniehansen:
@mnapoli I'm planning to implement this tonight as this have gotten merged. Just wanted to wait until the event handler was finished. I'll open a new merge request later tonight or tomorrow for implementing a SimpleWebsocketClient so that we can see if it fits into Bref :)

Thank you for taking your time with this!

danniehansen:
@mnapoli Might take a bit longer than I thought. I think we would need a async-aws repo for a websocket client api integration. I'm currently creating a MR meant for discussion and showcasing a implementation. But it would require an official repo from async-aws to support this. Any other implementation just seems fragile as all else uses async-aws.

mnapoli:
Understood, no worries. Let's see how it goes, and thank you for following through with this.

PR #774, opened 2020-10-14
Update gd extension reference

skyrpex:
Remove the gd extension from the "installed but disabled by default" section and added it to the "installed and enabled" section


mnapoli:
Thanks

PR #775, opened 2020-10-26
Migrate to Amazon Linux 2 images

shouze:
[As announced on 12 AUG 2020](https://aws.amazon.com/fr/blogs/compute/migrating-aws-lambda-functions-to-al2/), Amazon Linux 1 EOL (End Of Life) is coming in December 2020.

### DIFF
--- a/runtime/base/php-74.Dockerfile
+++ b/runtime/base/php-74.Dockerfile
@@ -109,9 +109,10 @@ RUN set -xe; \
  cp php.ini-production ${INSTALL_DIR}/etc/php/php.ini
 
 # Symlink all our binaries into /opt/bin so that Lambda sees them in the path.
-RUN mkdir -p /opt/bin
-RUN ln -s /opt/bref/bin/* /opt/bin
-RUN ln -s /opt/bref/sbin/* /opt/bin
+RUN mkdir -p /opt/bin \
+    && cd /opt/bin \
+    && ln -s ../bref/bin/* . \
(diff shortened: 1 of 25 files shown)


mnapoli:
Thank you @shouze!

runtime/base/php-74.Dockerfile
```diff
@@ -128,7 +129,7 @@ RUN /tmp/clean.sh && rm /tmp/clean.sh
 # Now we start back from a clean image.
 # We get rid of everything that is unnecessary (build tools, source code, and anything else
 # that might have created intermediate layers for docker) by copying online the /opt directory.
-FROM amazonlinux:2018.03
+FROM amazonlinux:2
```
mnapoli, line 132: Here we switch from `amazonlinux:2018.03` to `amazonlinux:2` and for PHP 7.3 and 8.0 we switch to `lambci/lambda:build-provided.al2`.

@shouze do you remember the reason? I'm not sure which one to keep in the end (to make this consistent).

shouze: Good catch! We should keep `amazonlinux:2` everywhere for this build stage. lambci one is a heavy docker image:
image
amazonlinux one is lighter, and more conform to the true production amazon environment:
image

shouze: Just in time before new year's eve! :clock1130:  https://github.com/brefphp/bref/pull/830

PR #779, opened 2020-11-03
docs: AWS new interface for API gateway and custom domains

BurningDog:
The AWS API Gateway interface for working with custom domains has changed. This PR updates the instructions and screenshot.

A note: I couldn't get a Symfony 4.4 install working on Lambda _without_ setting up custom domain names. So this guide should probably be linked to from https://bref.sh/docs/frameworks/symfony.html ?


docs/environment/custom-domains.md
```diff
@@ -31,9 +31,10 @@ After validating the domain and the certificate we can now link the custom domai
 - click "Create"
 - enter your domain name, select the certificate you created above and save
 - edit the domain that was created
-- add a "Path mapping" to map the `/` URL (or any URL you want) to your HTTP application in the `Prod` stage, for example:
+- click "Configure API mappings" to add an "API mapping" which maps the `/` URL (or any URL you want) to your HTTP application in the `Dev` stage, for example:
 
   ![](custom-domains-path-mapping.png)
+If you are using Cloudfront:
```
BurningDog, line 37: I'm not using Cloudfront, and so I couldn't verify if the following 2 instructions show in the new API gateway interface.

docs/environment/custom-domains.md
```diff
@@ -31,9 +31,10 @@ After validating the domain and the certificate we can now link the custom domai
...
 - after clicking "Save", note the URL in the `Target Domain Name` section (`<random>.cloudfront.net`)
```
mnapoli, line 38:
```suggestion
- after saving the "API mappings", find the `API Gateway domain name` in the "Configurations" tab (`<random>.cloudfront.net`)
```

This is what I mean:

CleanShot 2020-11-03 at 11 50 59

docs/environment/custom-domains.md
mnapoli: It's not really related to CloudFront. It's simply that after setting up the "Custom domain", we need to create a DNS entry to point our custom domain to API Gateway.

Check out my other comment, I've suggested to reword this a little bit. The "If you are using Cloudfront:" should be removed then.

```suggestion

```

docs/environment/custom-domains.md
BurningDog: Mine looks a bit different, though:

Screenshot 2020-11-03 at 13 54 34

docs/environment/custom-domains.md
BurningDog, line 37:
```suggestion
If your endpoint type is `Edge`:
```

docs/environment/custom-domains.md
mnapoli: Yes, ok here is an alternative suggestion:

```suggestion
- after saving the "API mappings", find the `API Gateway domain name` in the "Configurations" tab
```

docs/environment/custom-domains.md
mnapoli: This line needs to be removed even with regional domains. The DNS entry must be created.

BurningDog: Ok, I'll remove it - but I had to create my CNAME entry and an AWS certificate before I was even able to get to this screen.

mnapoli:
Perfect thank you!

PR #780, opened 2020-11-03
Update websites.md

BurningDog:
Explained why the S3 bucket name must be universally unique within Amazon, as I got an error which confounded me because I didn't understand understand this. Thanks to @mnapoli for his help on slack!


mnapoli:
Thanks!

PR #783, opened 2020-11-11
SimpleWebsocketClient implementation

danniehansen:
Work in progress implementation of a SimpleWebsocketClient utility class for interacting with ApiGateway websocket connections. 

### Problem to solve
After latest websocket merge we now have a way of handling the events coming from ApiGateway websocket. However - this is only one side of the card. While we're able to handle incoming requests, it's not yet natively possible in bref to handle outgoing. This PR solves that by supplying a simple websocket client class to interact with connections.

### The process
The original idea was to utilize async-aws - but I quickly realized that async-aws does not have a client for ApiGateway websockets. This lead me to try various paths to manually implement the request signing that async-aws does & quickly made up my mind of utilizing async-aws abstract nature and implement a simple client to process the requests.

Utilizing async-aws with a tiny class for enabling API connectivity comes with few benefits.

1. No manual signing
2. Exceptions thrown by the request/response wrappers in async-aws will be exactly the same (that's the theory) if a native async-aws implementation becomes a reality.

To avoid backward compatibility issues in case of a native implementation by async-aws I've made it so that `SimpleWebsocketClient` class abstracts the response away from the async-aws client by returning bool for message/disconnect method & a `WebsocketClientStatus` class for the status method. This means we can safely change out the underlying async-aws implementation with a native one in due time.

### Result

Based on a apiId, region, stage & connectionId it's possible to either disconnect, get status or send messages to clients. This is all done at a similar fashion as with SimpleLambdaClient. You suppliy it with a apiId, region, stage & you're able to interact with connections.

### Example

The following example utilize dynamo-db to create a connection pool and on each `ping` that it receives it scans over all connected clients and respond with `pong`. Nothing fancy - just to showcase that it works.

**handler/websocket.php**
```php
<?php

use Bref\Context\Context;
use Bref\Event\ApiGateway\WebsocketEvent;
use Bref\Event\ApiGateway\WebsocketHandler;
use Bref\Event\Http\HttpResponse;
use Bref\Websocket\SimpleWebsocketClient;
use AsyncAws\DynamoDb\Input\PutItemInput;
use AsyncAws\DynamoDb\ValueObject\AttributeValue;
use AsyncAws\DynamoDb\DynamoDbClient;

class MyHandler extends WebsocketHandler
{
    public function handleWebsocket(WebsocketEvent $event, Context $context): HttpResponse
    {
        $dynamoDb = new DynamoDbClient();

        switch ($event->getEventType()) {
            case 'CONNECT':
                $dynamoDb->putItem(new PutItemInput([
                    'TableName' => 'connectionPool',
                    'Item' => [
                        'connectionId' => new AttributeValue(['S' => $event->getConnectionId()]),
                        'apiId' => new AttributeValue(['S' => $event->getApiId()]),
                        'region' => new AttributeValue(['S' => $event->getRegion()]),
                        'stage' => new AttributeValue(['S' => $event->getStage()]),
                    ],
                ]));

                return new HttpResponse('connect');

            case 'DISCONNECT':
                $dynamoDb->deleteItem([
                    'TableName' => 'connectionPool',
                    'Key' => [
                        'connectionId' => [
                            'S' => $event->getConnectionId(),
                        ],
                    ]
                ]);

                return new HttpResponse('disconnect');

            default:
                if ($event->getBody() === 'ping') {
                    foreach ($dynamoDb->scan([
                        'TableName' => 'connectionPool',
                    ])->getItems() as $item) {
                        $connectionId = $item['connectionId']->getS();

                        $client = SimpleWebsocketClient::create(
                            $item['apiId']->getS(),
                            $item['region']->getS(),
                            $item['stage']->getS()
                        );

                        $client->message($connectionId, 'pong');

                        $status = $client->status($connectionId);

                        echo json_encode($status->toArray());
                    }
                }

                return new HttpResponse('message');
        }
    }
}

return new MyHandler();
```

**serverless.yml**
```yaml
service: test-apigateway
provider:
  name: aws
  region: eu-west-1
  stage: dev
  runtime: provided
  iamRoleStatements:
    - Effect: Allow
      Action:
        - dynamodb:Scan
        - dynamodb:PutItem
        - dynamodb:DeleteItem
      Resource: "*"
plugins:
  - ./vendor/bref/bref
resources:
  Resources:
    connectionPool:
      Type: AWS::DynamoDB::Table
      Properties:
        TableName: connectionPool
        AttributeDefinitions:
          - AttributeName: connectionId
            AttributeType: S
          - AttributeName: apiId
            AttributeType: S
          - AttributeName: region
            AttributeType: S
          - AttributeName: stage
            AttributeType: S
        KeySchema:
          - AttributeName: connectionId
            KeyType: HASH
        ProvisionedThroughput:
          ReadCapacityUnits: 1
          WriteCapacityUnits: 1
functions:
  websocket:
    handler: handler/websocket.php
    layers:
      - ${bref:layer.php-74}
    events:
      - websocket: $connect
      - websocket: $disconnect
      - websocket: $default
```

**JS (just paste in browser developer console)**
```javascript
let socket = new WebSocket('wss://xyz.execute-api.eu-west-1.amazonaws.com/dev');

socket.onopen = function(e) {
  socket.send("ping");
};

socket.onmessage = function(event) {
    console.log('[message] Data received from server:', event);
};
```

### What's left?

- [x] Tests - I'm unsure of what tests to create for this. A lot of this require actual API communication to take place - which I'm unsure of if we can even test? I was however tempted to add a simple test, testing the WebsocketClientStatus class. Please share your thoughts here.
- [x] Fix any code style issues in PR


src/ApiGateway/WebsocketClient.php
```diff
@@ -0,0 +1,25 @@
+<?php declare(strict_types=1);
+
+namespace Bref\ApiGateway;
+
+use AsyncAws\Core\AbstractApi;
+use AsyncAws\Core\Request;
+use AsyncAws\Core\Response;
+
+class WebsocketClient extends AbstractApi
```
mnapoli, line 9: From what I understand this class necessary because there is no native support in Async AWS?

If that's indeed it, could you:

- add a block comment to explain that
- add a `@internal` tag in the docblock (to say that it shouldn't be used directly by users)

src/ApiGateway/SimpleWebsocketClient.php
```diff
@@ -0,0 +1,62 @@
+<?php declare(strict_types=1);
+
+namespace Bref\ApiGateway;
```
mnapoli, line 3: Sorry to nitpick but could be use a `Bref\Websocket` namespace?

I think it would make more sense if later we move this to a separate package, it would probably named `bref/websocket` (it would be specific to websocket).

src/ApiGateway/SimpleWebsocketClient.php
```diff
@@ -0,0 +1,62 @@
...
+ */
+final class SimpleWebsocketClient
+{
+    /** @var string */
+    private $stage;
+
+    /** @var WebsocketClient */
+    private $client;
+
+    public function __construct(string $apiId, string $region, string $stage, int $timeout = 60)
```
mnapoli, line 20:
```suggestion
    public function __construct(string $apiId, string $region, string $stage, int $timeout = 10)
```

Default timeouts at 60s are usually problematic because the timeout for HTTP requests is 30 seconds on Lambda. That means that we never know where the code was actually hanged. Because of that, and also because having users wait 60 seconds when a service is down is a bit much, I always advocate for shorter timeouts.

src/ApiGateway/SimpleWebsocketClient.php
```diff
@@ -0,0 +1,62 @@
...
+                'timeout' => $timeout,
+            ])
+        );
+    }
+
+    public function disconnect(string $connectionId): bool
+    {
+        return $this->client->process(
+            $this->request('DELETE', sprintf('/%s/@connections/%s', $this->stage, $connectionId))
+        )->getStatusCode() === 200;
```
mnapoli, line 39: Instead of returning booleans would it would make more sense to throw exceptions? Said another way: if the status code is `!== 200` does that mean that this is an exceptional error? Or is that a normal case?

Exceptions would seem more consistent with any other PHP class (where the default behavior works, else there's an exception).

src/ApiGateway/WebsocketClientStatus.php
```diff
@@ -0,0 +1,56 @@
...
+    private $sourceIp;
+
+    /** @var string */
+    private $userAgent;
+
+    /** @var string */
+    private $connectedAt;
+
+    /** @var string */
+    private $lastActiveAt;
```
mnapoli, line 17: Should these 2 variables be `DateTimeImmutable` instances?

mnapoli (commented):
This is really great, thanks a lot!

I've opened a few minor discussions, but the structure of the PR is 👍 

Regarding tests, I think testing `SimpleWebsocketClient` would be great as it's where there is the most logic and where it would be possible to introduce a regression in the future.

To add test, the Symfony HTTP client has helpers: https://symfony.com/doc/current/http_client.html#testing-http-clients-and-responses

I would recommend the following:

- rename the current constructor to a `public static function create()`
- change the constructor to take a `HttpClient`

The benefits:

- easy to mock the HTTP client in tests
- easier later if we want to add another constructor, for example `createFromApiUrl(string)`

src/ApiGateway/WebsocketClientStatus.php
danniehansen: @mnapoli Right. That would make sense. I'll change this.

src/ApiGateway/SimpleWebsocketClient.php
danniehansen: @mnapoli Wops - you're right. Exceptions is already thrown in case of a unsuccessful disconnect anyway - so this is really unnecessary. I'll change this up.

src/ApiGateway/SimpleWebsocketClient.php
danniehansen: @mnapoli Sure - I'll change this

src/ApiGateway/WebsocketClient.php
danniehansen: @mnapoli Correct. I'll add some docblock explaining this.

danniehansen:
@mnapoli Thank you for taking your time reviewing this. Great feedback & suggestion with testing 👍 I'll do the changes here in a bit & report back once all is resolved.

danniehansen:
@mnapoli Not sure what is going on here. All unit tests are passing locally - but in CI it appears that no `MockResponse` is passed to `MockHttpClient`. Do you have any idea of what could be the cause here? Everything works perfectly locally and there all unit tests are passing.

danniehansen:
@mnapoli Looks like the issue was caused by me having ~/.aws credentials defined locally. I've added an option to define your own credentials provider in the constructor. This essentially resolves the issue as I can just use NullProvider during unit test. I'm not too sure of this implementation of having this in the constructor - but not sure of how else it should be implemented because we need that NullProvider.

Other than that it should all be ready :)

mnapoli:
Thank you! I've marked the constructor as `@internal` so that we have freedom to adjust it if necessary in the future.

PR #788, opened 2020-11-18
fix: missing amazon linux 2 lib

shouze:
This library was missing resulting of an immediate crash of Amazon Linux 2 deployed lambda.


mnapoli:
Thanks!

PR #791, opened 2020-11-20
Reorganize the documentation to separate "Bref for web apps" and "Bref for functions"

### LINKED ISSUE #641: Reorganizing the documentation to separate "Bref for web hosting" and "Bref for functions"
Following many exchanges and support requests over the months, as well as the fact that native events are better supported by Bref, it could be good to reorganize the documentation to separate:

- running PHP as usual (PHP-FPM, frameworks, etc.) on Lambda
- running PHP functions in the "AWS Lambda" way

In the first use case, we use Lambda for (scalable, cheap, effortless) **web hosting**.

In the second use case, we use Lambda to build **event-driven** serverless applications using **functions**.

Naming is hard, but here are possibilities for the 2 "main sections" of the documentation:

- **Bref for web hosting**
- **Bref for functions**

Thank you @GeeH for coming up with that!

The last one could also be *Bref for event-driven applications* but I find that more confusing, and it's a mouthful.

## Goals

- Users should more easily understand that there are 2 main use cases for Bref: web hosting and functions.
  They should know easily which one they want to use. They should also be aware that the other exists.
- AFAIK there are 2 main categories of users: those that know about Lambda and want to write functions in PHP, and those that don't and want to benefit from serverless.
  The first one should easily land on "Bref for functions". The second one should easily land on "Bref for web hosting".
  This is why I think "Web hosting" should be first in the list. It should be the default Bref use case, both in documentation and `bref init`. Those that know about Lambda will find their way for easily.
- "Web hosting" goes hand in hand with frameworks.
  That means that Laravel and Symfony should be clearly visible.
- When a user has a specific use case or problem in mind, they should find the solution easily via the menu
  That means that pages should be structured mostly around use cases (local development, deploying assets, using databases, etc.). The exception I see is the reference regarding `serverless.yml`, which is useful.
- Users should be able to follow the documentation linearly
  That means: ideally the documentation should build over what was explained previously. For example, we should not explain how to deploy before users have something to deploy.

## Proposal

Here is a 1st proposal, feedback is welcome:

CleanShot 2020-05-04 at 11 00 50

The design is not set in stone of course, this is a first attempt that iterates on the current design.

mnapoli:
Fixes #641 

Following many exchanges and support requests over the months, as well as the fact that native events are better supported by Bref, it could be good to reorganize the documentation to separate:

- running PHP as usual (PHP-FPM, frameworks, etc.) on Lambda
- running PHP functions in the "AWS Lambda" way

In the first use case, we use Lambda for (scalable, cheap, effortless) **web app hosting**.

In the second use case, we use Lambda to build **event-driven** serverless applications using **functions**.

That will lead to 2 "main sections" of the documentation:

- **Bref for web apps**
- **Bref for event-driven functions**

More details are available in #641 

Note: I've also switched the documentation from API Gateway v1 to v2's HTTP API's to address #309.

Here is a preview:

image


docs/function/handlers.md
```diff
@@ -0,0 +1,342 @@
...
+
+class Handler extends WebsocketHandler
+{
+    public function handleWebsocket(WebsocketEvent $event, Context $context): HttpResponse
+    {
+        $route = $event->getRouteKey();
+        $eventType = $event->getEventType();
+        $body = $event->getBody();
+
+        return new HttpResponse('ok');
```
mnapoli, line 212: @danniehansen does that look correct to you? I'm adding a really small example for websockets.

danniehansen: @mnapoli 👌Looks correct yes

mnapoli: Thanks!

PR #792, opened 2020-11-20
Feature/config local run

Fabex:
For some reasons, a project could not have is serverless.yaml in the project root folder but in a subfolder.
So I add a new argument `--config=` to the local run command in order to config the path of the serverless.yaml file

### DIFF
--- a/docs/local-development.md
+++ b/docs/local-development.md
@@ -63,6 +63,12 @@ $ vendor/bin/bref local --handler=my-function.php --file=event.json
 Hello Jane
 ```
 
+Or if your `serverless.yml` is not in root folder, use the `--config` parameter instead:
+```bash
+$ vendor/bin/bref local hello --config=path/to/your/serverless.yml --file=event.json
+Hello world
+```
+
 ## HTTP applications
 
 If you want to keep things simple, you can run your PHP application like you did without Bref. For example with your favorite framework:
--- a/docs/local-development.md
+++ b/docs/local-development.md
@@ -63,12 +63,6 @@ $ vendor/bin/bref local --handler=my-function.php --file=event.json
 Hello Jane
 ```
 
-Or if your `serverless.yml` is not in root folder, use the `--config` parameter instead:
-```bash
-$ vendor/bin/bref local hello --config=path/to/your/serverless.yml --file=event.json
-Hello world
-```
-
 ## HTTP applications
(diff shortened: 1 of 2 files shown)


src/Console/Command/Local.php
```diff
@@ -77,13 +81,14 @@ public function __invoke(?string $function, ?string $data, ?string $file, ?strin
         return 0;
     }
 
-    private function handlerFromServerlessYml(string $function): string
+    private function handlerFromServerlessYml(string $function, ?string $config): string
     {
-        if (! file_exists('serverless.yml')) {
+        $file = $config ?? 'serverless.yml';
+        if (! file_exists($file)) {
             throw new Exception("No `serverless.yml` file was found to resolve function $function.\nIf you do not use serverless.yml, pass the handler via the `--handler` option: vendor/bin/bref local --handler=file.php");
```
mnapoli, line 88:
```suggestion
            throw new Exception("No `serverless.yml` file was found to resolve function $function.\nIf you do not use serverless.yml, pass the handler via the `--handler` option: vendor/bin/bref local --handler=file.php\nIf your serverless.yml file is stored elsewhere, use the `--config` option: vendor/bin/bref local --config=foo/serverless.yml");
```

docs/local-development.md
```diff
@@ -63,6 +63,12 @@ $ vendor/bin/bref local --handler=my-function.php --file=event.json
 Hello Jane
 ```
 
+Or if your `serverless.yml` is not in root folder, use the `--config` parameter instead:
+```bash
+$ vendor/bin/bref local hello --config=path/to/your/serverless.yml --file=event.json
+Hello world
+```
```
mnapoli, line 70: I would suggest we keep the documentation simple and not mention the option: instead we suggest it in the error message. That way users will try, fail and be able to fix the command themselves.

mnapoli (commented):
Thanks! Just a suggestion regarding the docs and then we're good to merge!

src/Console/Command/Local.php
Fabex: yes you right

docs/local-development.md
Fabex: ok, no problem, I remove this from the doc

mnapoli:
Thanks!

PR #800, opened 2020-11-25
feature: Deal with nested exception

tyx:
Currently, we display only first level of nested exception. 

If a previous exception exists, it would be nice to dump its trace also to cloudwatch

### DIFF
--- a/tests/Runtime/LambdaRuntimeTest.php
+++ b/tests/Runtime/LambdaRuntimeTest.php
@@ -81,6 +81,22 @@ class LambdaRuntimeTest extends TestCase
         $this->assertErrorInLogs('RuntimeException', 'This is an exception');
     }
 
+    public function test nested exceptions in the handler result in an invocation error()
+    {
+        $this->givenAnEvent(['Hello' => 'world!']);
+
+        $this->runtime->processNextEvent(function () {
+            throw new \RuntimeException('This is an exception', 0, new \RuntimeException('The previous exception.', 0, new \Exception('The original exception.')));
+        });
+
+        $this->assertInvocationErrorResult('RuntimeException', 'This is an exception');
+        $this->assertErrorInLogs('RuntimeException', 'This is an exception');
+        $this->assertPreviousErrorsInLogs([
+            ['errorClass' => 'RuntimeException', 'errorMessage' => 'The previous exception.'],
+            ['errorClass' => 'Exception', 'errorMessage' => 'The original exception.'],
+        ]);
+    }
+
     public function test an error is thrown if the runtime API returns a wrong response()
     {
         $this->expectExceptionMessage('Failed to fetch next Lambda invocation: The requested URL returned error: 404 Not Found');
@@ -391,6 +407,7 @@ ERROR;
         $this->assertNotEmpty($requestId);
 
         $invocationResult = json_decode($json, true);
+        unset($invocationResult['previous']);
         $this->assertSame([
             'errorType',
             'errorMessage',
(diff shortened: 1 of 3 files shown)


mnapoli:
Thanks, that's a very good point!

The thing however is that currently we try to follow the formatting of errors with the NodeJS runtime -> that allows us to be compatible with all the tooling that exists out there (e.g. monitoring SaaS, etc.).

If I understand correctly, the JSON stack trace is no longer a stack trace but a nested object containing multiple stack traces 🤔 

We need to check how Node (or other runtimes) handle chained exceptions like this, if you have any idea I'm up for it!

mnapoli:
For reference, [here is what a NodeJS error log line looks like](https://docs.aws.amazon.com/lambda/latest/dg/nodejs-exceptions.html):

```
2019-06-05T22:11:27.082Z        8bbbfb91-a3ff-4502-b1b7-cb8f6658de64    ERROR   Invoke Error    {"errorType":"ReferenceError","errorMessage":"x is not defined","stack":["ReferenceError: x is not defined","    at Runtime.exports.handler (/var/task/index.js:2:3)","    at Runtime.handleOnce (/var/runtime/Runtime.js:63:25)","    at process._tickCallback (internal/process/next_tick.js:68:7)"]}
```

If we format the JSON for readability (which CloudWatch does):

```
2019-06-05T22:11:27.082Z        8bbbfb91-a3ff-4502-b1b7-cb8f6658de64    ERROR   Invoke Error
{
  "errorType":"ReferenceError",
  "errorMessage":"x is not defined",
  "stack": [
    "ReferenceError: x is not defined",
    "    at Runtime.exports.handler (/var/task/index.js:2:3)",
    "    at Runtime.handleOnce (/var/runtime/Runtime.js:63:25)",
    "    at process._tickCallback (internal/process/next_tick.js:68:7)"
  ]
}
```

mnapoli:
It seems JS doesn't have that kind of feature (the "previous" exception thing): https://stackoverflow.com/questions/42754270/re-throwing-exception-in-nodejs-and-not-losing-stack-trace

Maybe what we could do is keep the original JSON shape and add extra information as extra keys, so that tools can still support the base fields?

Example:

```
2019-06-05T22:11:27.082Z        8bbbfb91-a3ff-4502-b1b7-cb8f6658de64    ERROR   Invoke Error
{
  "errorType":"WrappedException",
  "errorMessage":"x is not defined",
  "stack": [
    "...",
    "..."
  ],
  "previous": [
    {
      "errorType":"OriginalException",
      "errorMessage":"x is not defined",
      "stack": [
        "...",
        "..."
      ],
    },
    { ... }
  ],
}
```

WDYT?

tyx:
Ok for me ! I will update the PR

tests/Runtime/LambdaRuntimeTest.php
```diff
@@ -81,6 +81,19 @@ public function test exceptions in the handler result in an invocation 
...
 
+    public function test nested exceptions in the handler result in an invocation error()
+    {
+        $this->givenAnEvent(['Hello' => 'world!']);
+
+        $this->runtime->processNextEvent(function () {
+            throw new \RuntimeException('This is an exception', 0, new \Exception('The original exception.'));
+        });
+
+        $this->assertInvocationErrorResult('Exception', 'The original exception.');
```
mnapoli, line 92: We should probably report the wrapping exception (i.e. the one that was last trown), because it's the exception actually thrown by the app.

tyx: Yes you're right

tyx:
I updated the PR with the `previous` key added.

mnapoli:
Thanks!

PR #806, opened 2020-11-30
Local - Parse config tags, prevent from crashing

mheki:
- Adds an exception if both function and header are not provided to local invoke.
- Prevents from crashing when config contains tags

Hey guys,
I've just updated to bref 1.0.0 (well done BTW 🎉 ) and tried `vendor/bin/bref local`

First run was obviously wrong, because I didn't pass in a function name, but it ended up with fatal error instead of an exception:
```
vendor/bin/bref local
PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Bref\Console\Command\Local::handlerFromServerlessYml()
 must be of the type string, null given, 
called in vendor/bref/bref/src/Console/Command/Local.php on line 36
and defined in vendor/bref/bref/src/Console/Command/Local.php:84
```
Changed to:
Screenshot 2020-11-30 at 13 48 42

Then, when I got it right, I couldn't run it against my `serverless.yml` config file due to some `CloudFrontDistribution` config containing tags:
```
Origins:
            -   Id: AuthApiGateway
                DomainName: !Join ['.', [!Ref ApiGatewayRestApi, 'execute-api', !Ref AWS::Region, 'amazonaws.com']]
```
adding `Yaml::PARSE_CUSTOM_TAGS` helped.

### DIFF
--- a/src/Console/Command/Local.php
+++ b/src/Console/Command/Local.php
@@ -24,7 +24,6 @@ class Local
         if ($function === null && $handler === null) {
             throw new Exception('Please provide a function name or the --handler= option.');
         }
-        
         if ($function && $data && $handler) {
             throw new Exception('You cannot provide both a funtion name and the --handler= option.');
         }
--- a/src/Console/Command/Local.php
+++ b/src/Console/Command/Local.php
@@ -21,6 +21,10 @@ class Local
 
     public function __invoke(?string $function, ?string $data, ?string $file, ?string $handler, ?string $config, SymfonyStyle $io): int
     {
+        if ($function === null && $handler === null) {
+            throw new Exception('Please provide a function name or the --handler= option.');
+        }
+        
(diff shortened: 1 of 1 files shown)


src/Console/Command/Local.php
```diff
@@ -21,6 +21,10 @@ class Local
 
     public function __invoke(?string $function, ?string $data, ?string $file, ?string $handler, ?string $config, SymfonyStyle $io): int
     {
+        if ($function === null && $handler === null) {
+            throw new Exception('Please provide a function name or the --handler= option.');
+        }
+
```
mnapoli, line 27:
```suggestion

```

mnapoli (commented):
That's super cool!

I added a suggestion to fix the CI and then we're good to merge.

mnapoli:
Thanks!

Issue #818, opened 2020-12-08
API Gateway v2: multiple response headers / multiple cookies


hvipana:
**Description:**

I have upgraded my deployment to HttpApi (API-Gateway v2). After that I got only one cookie in response and the second  cookie `laravel_session` was missing.

<img width="228" alt="image" src="https://user-images.githubusercontent.com/55196856/101472068-5e73ad80-3948-11eb-9f40-2ce57f990eee.png">

Then I switched back to RestApi (v1) and I got both cookies in response.

<img width="203" alt="image" src="https://user-images.githubusercontent.com/55196856/101472232-9da1fe80-3948-11eb-8157-a8b45c2cf5df.png">

bref: 1.0.2
laravel: 8.4.4
runtime: provided.al2
layer.php-74-fpm

flemzord:
I have a same bug, I came back to APIG REST for the moment

mnapoli:
Hi,

Thanks for the report. Indeed, the API Gateway v2 format does not support multiple headers in the response. Or at least, that is what we have concluded over our research.

Here are useful links to work further on this:

- [The docs on API Gateway v1 and v2 format for requests and responses](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html)
- The current Bref code that doesn't support multi-headers in responses: [here](https://github.com/brefphp/bref/blob/9a46653388aa638629634194949339a1aefca430/src/Event/Http/HttpRequestEvent.php#L71-L78) and [here](https://github.com/brefphp/bref/blob/c3eb5ddeec66bac95cd7a3a1a661dddb15747cd4/src/Event/Http/HttpResponse.php#L38-L44)

If you want to contribute solving this:

- test/confirm whether it is possible to return multiple headers (override `false` with `true` [here](https://github.com/brefphp/bref/blob/9a46653388aa638629634194949339a1aefca430/src/Event/Http/HttpRequestEvent.php#L74)): if it works, we can work on changing that in Bref
- investigate if it's possible to find another solution, for example returning all cookies into one response header?

Workarounds:

- stay on REST API which uses the v1 format
- use HTTP API but force the v1 format:

  ```
  provider:
    httpApi:
      payload: '1.0'
  ```

josecl:
> Here are useful links to work further on this:
> 
[...]

PR #820, opened 2020-12-09
Support for multiple cookies in http payload version 2.0

hvipana:
Fix for #818

For http payload version 2.0 there is no `multiValueHeaders` field anymore. But to support multiple cookies, a new `cookies` field was introduced as described here:
https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html

This PR adds the `cookies` field to the response and removes the `Set-Cookie` header, that is now handled by API-Gateway v2.

### DIFF
--- a/src/Event/Http/HttpResponse.php
+++ b/src/Event/Http/HttpResponse.php
@@ -39,8 +39,13 @@ final class HttpResponse
                 // Make sure the values are always arrays
                 $headers[$name] = is_array($values) ? $values : [$values];
             } else {
-                // Make sure the values are never arrays
-                $headers[$name] = is_array($values) ? end($values) : $values;
+                if ($name === 'Set-Cookie') {
+                     // Support for multiple cookies in http payload version 2.0
+                    $cookies = is_array($values) ? $values : [$values];
+                } else {
+                    // Make sure the values are never arrays
+                    $headers[$name] = is_array($values) ? end($values) : $values;
+                }
             }
         }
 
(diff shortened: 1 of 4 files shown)


src/Event/Http/HttpResponse.php
```diff
@@ -53,11 +58,17 @@ public function toApiGatewayFormat(bool $multiHeaders = false): array
...
+        $response = [
             'isBase64Encoded' => $base64Encoding,
             'statusCode' => $this->statusCode,
             $headersKey => $headers,
             'body' => $base64Encoding ? base64_encode($this->body) : $this->body,
         ];
+
+        if (isset($cookies)) {
+            $response['cookies'] = $cookies;
+        }
```
mnapoli, line 70: At this point I'm afraid the `cookies` key might be added to the v1.0 format too, and the `Set-Cookies` header will be removed (from v1.0 responses) 🤔

tests/Handler/FpmHandlerTest.php
```diff
@@ -1018,9 +1018,12 @@ public function test response with multivalue headers()
...
     {
-        $cookieHeader = $this->get('cookies.php')['headers']['Set-Cookie'];
-
-        self::assertEquals('MyCookie=MyValue; expires=Fri, 12-Jan-2018 08:32:03 GMT; Max-Age=0; path=/hello/; domain=example.com; secure; HttpOnly', $cookieHeader);
+        $cookieHeader = $this->get('cookies.php', [
+            'version' => '1.0',
+            'httpMethod' => 'GET',
+        ])['cookies'];
+        self::assertEquals('MyCookie=FirstValue; expires=Fri, 12-Jan-2018 08:32:03 GMT; Max-Age=0; path=/hello/; domain=example.com; secure; HttpOnly', $cookieHeader[0]);
+        self::assertEquals('MyCookie=MyValue; expires=Fri, 12-Jan-2018 08:32:03 GMT; Max-Age=0; path=/hello/; domain=example.com; secure; HttpOnly', $cookieHeader[1]);
```
mnapoli, line 1026: Could you explain this change?

mnapoli:
Closing until we have some news.

Anyone else feel free to pick up this PR again in a new pull request.

mnapoli:
FYI I've opened #841 as an alternative.

PR #830, opened 2020-12-31
use new amazon container images

shouze:
During 2020 re:Invent, Amazon has anounced support of docker images for lambdas. They've also [published new base container images](https://gallery.ecr.aws/lambda/provided). This pull request suggests to use corresponding images to build bref images.

I've tested on some apps and it's ok. The only thing is that in the new images the user is root and there's no more filesystem protection (previous images had only /var/task and /tmp as read/write paths, others were read-only).

Big advantage by the way is that resulting docker images are lightweight compared to currently produced ones.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,7 +61,7 @@ changelog:
 
 # http://amazon-linux-2-packages.bref.sh/
 amazonlinux-package-list:
-	docker run --rm -it amazonlinux:2 yum list --quiet --color=never > index.html
+	docker run --rm -it --entrypoint= public.ecr.aws/lambda/provided:al2 yum list --quiet --color=never > index.html
 	aws s3 cp index.html s3://amazon-linux-2-packages.bref.sh/ --content-type=text/plain
 	rm index.html
 
--- a/runtime/Makefile
+++ b/runtime/Makefile
@@ -13,7 +13,15 @@ export/php%.zip: docker-images
 	PHP_VERSION=$$(echo $@ | cut -d'/' -f 2 | cut -d'.' -f 1);\
 	rm -f $@;\
 	cd export ; \
-	docker run --rm --entrypoint= bref/$$PHP_VERSION:latest sh -c "cd /opt && zip -qq -y -r - {*,.[!.]*} > /tmp/file.zip && cat /tmp/file.zip -" > $$PHP_VERSION.zip ;
+	set -e ; \
+	rm -rf opt ; \
+	CID=$$(docker create --entrypoint=scratch bref/$$PHP_VERSION:latest) ; \
+	docker cp $${CID}:/opt . ; \
+	docker rm $${CID} ; \
+	cd opt ; \
+	zip -qq -y -r - {*,.[!.]*} > ../$$PHP_VERSION.zip
+	cd ../ ; \
+	rm -rf opt ;
 
 # The console runtime
 export/console.zip: layers/console/bootstrap
--- a/runtime/layers/function/Dockerfile
+++ b/runtime/layers/function/Dockerfile
@@ -8,6 +8,6 @@ COPY php.ini /opt/bref/etc/php/conf.d/bref.ini
 # Remove PHP-FPM
 RUN rm /opt/bref/sbin/php-fpm /opt/bin/php-fpm
 
-# Build the final image from the lambci image that is close to the production environment
-FROM lambci/lambda:build-provided.al2
+# Build the final image from the amazon image that is close to the production environment
+FROM public.ecr.aws/lambda/provided:al2
 COPY --from=0 /opt /opt
--- a/docs/environment/php.md
+++ b/docs/environment/php.md
@@ -149,15 +149,15 @@ FROM bref/build-php-74
 
 RUN curl -A "Docker" -o /tmp/blackfire.so -L -s "https://packages.blackfire.io/binaries/blackfire-php/1.42.0/blackfire-php-linux_amd64-php-74.so"
 
-# Build the final image from the lambci image that is close to the production environment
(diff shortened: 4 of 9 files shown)


runtime/Makefile
```diff
@@ -13,7 +13,15 @@ export/php%.zip: docker-images
 	PHP_VERSION=$$(echo $@ | cut -d'/' -f 2 | cut -d'.' -f 1);\
 	rm -f $@;\
 	cd export ; \
-	docker run --rm --entrypoint= bref/$$PHP_VERSION:latest sh -c "cd /opt && zip -qq -y -r - {*,.[!.]*} > /tmp/file.zip && cat /tmp/file.zip -" > $$PHP_VERSION.zip ;
```
shouze, line 16: zip is not included anymore so like in [brefphp/extra-php-extensions](https://github.com/brefphp/extra-php-extensions) we have to use another way to extract /opt filesystem from docker image.

mnapoli (approved):
Thanks!

I was able to build and test (quick tests) too.

@Nyholm do you recon some changes would be necessary in bref extra extensions afterwards?

Nyholm:
I don’t think there needs to be any changes in the extra extensions. As long as the paths to php has not changed. 

If I understand correctly, using this image will never make my application to error with “cannot write to cache”. The cache will not be stored between requests, but it will not error, right?

shouze:
> I don’t think there needs to be any changes in the extra extensions. As long as the paths to php has not changed.

yes it just works.

> If I understand correctly, using this image will never make my application to error with “cannot write to cache”. The cache will not be stored between requests, but it will not error, right?

yes that's it, no more errors when you try to write files somewhere else than in /tmp.

Write errors when you try to write to cache are very specific of `lambci/lambda` images, even `amazonlinux` image don't have this kind of protection. This is mostly because of this [Dockerfile instruction](https://github.com/lambci/docker-lambda/blob/b25f2695dec37f4b25c71f33ae3166075df38b6a/provided.al2/run/Dockerfile#L11). So we can imagine adding a default `bref` user for example and making it the default runtime user of the produced docker image, WDYT?

Nyholm:
I’m impressed how much you know about this :)

> So we can imagine adding a default bref user for example and making it the default runtime user of the produced docker image,

I don’t quite follow, could you explain why we would like to add a Bref user? What would be the benefit?

runtime/layers/function/Dockerfile
```diff
@@ -8,6 +8,6 @@ COPY php.ini /opt/bref/etc/php/conf.d/bref.ini
 # Remove PHP-FPM
 RUN rm /opt/bref/sbin/php-fpm /opt/bin/php-fpm
 
-# Build the final image from the lambci image that is close to the production environment
-FROM lambci/lambda:build-provided.al2
+# Build the final image from the amazon image that is close to the production environment
+FROM public.ecr.aws/lambda/provided:al2
 COPY --from=0 /opt /opt
```
shouze, line 13: @Nyholm to keep BC regarding filesystem write proection we could add the 2 following instructions into this Dockerfile, bu also in `runtime/layers/fpm/Dockerfile`
```suggestion
COPY --from=0 /opt /opt
RUN echo "bref:x:1000:1000:Bref:/:/sbin/nologin" >> /etc/passwd \
  && chown bref:1000 /tmp && \
  chmod 700 /tmp
USER bref
```

Nyholm: No, I think it is a good thing we dont have the write protection

shouze: Ok perfect, looks also good to me to keep this PR as is.

shouze: so @mnapoli I guess we can move forward and merge :angel:

mnapoli: Yep I'm intending to merge this, just took some holidays this week :P I'm avoiding non-urgent work this week, will do that next week.

Thanks for all the clarification! I also agree (for now) about not replicating the filesystem restrictions: most apps will write to the project directory (if they ever write anything) which is mounted, so users have control over making read-only that volume.

shouze: Thx - nothing urgent - enjoy holliday time!

shouze: I'll make another PR to upgrade to php 8.0.1 as it fixes an important tokenizer issue that would help using php-cs-fixer with php8 see https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/5415#issue-549599811.

mnapoli:
Let's finally do this, thank you for your patience!

shouze:
@mnapoli if you're ok let me upgrade php versions too prior to tag any release :wink:

Makefile
```diff
@@ -61,7 +61,7 @@ changelog:
 
 # http://amazon-linux-2-packages.bref.sh/
 amazonlinux-package-list:
-	docker run --rm -it amazonlinux:2 yum list --quiet --color=never > index.html
+	docker run --rm -it --entrypoint= public.ecr.aws/lambda/provided:al2 yum list --quiet --color=never > index.html
```
Nyholm, line 64: Should it really be an extra space here?

mnapoli:
@shouze 👍

Makefile
Nyholm: https://github.com/brefphp/bref/pull/836

mnapoli: I think so yes, this is to disable the `entrypoint`. The next argument is the Docker image, unless I'm mistaken.

shouze: @Nyholm yes we need the extra space in... we can put `--entrypoint=scratch` for example like in extra extensions if it make things a little more readable/clear.

PR #841, opened 2021-01-18
Support multiple cookies with API Gateway v2

### LINKED ISSUE #818: API Gateway v2: multiple response headers / multiple cookies

mnapoli:
Fixes #818 

Replaces #820 

This PR allows apps to set multiple cookies in HTTP responses when using API Gateway v2.

Previously this wasn't possible because v2 does not support "multi-value headers". But they added a specific `cookies` field which we can use to set multiple cookies.

### DIFF
--- a/src/Event/Http/FpmHandler.php
+++ b/src/Event/Http/FpmHandler.php
@@ -126,7 +126,7 @@ final class FpmHandler extends HttpHandler
             throw new FastCgiCommunicationFailed;
         }
 
-        $responseHeaders = $this->getResponseHeaders($response, $event->hasMultiHeader());
+        $responseHeaders = $this->getResponseHeaders($response);
 
         // Extract the status code
         if (isset($responseHeaders['status'])) {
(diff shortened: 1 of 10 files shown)


src/Event/Http/FpmHandler.php
```diff
@@ -280,17 +278,8 @@ private function waitUntilStopped(int $pid): void
...
-        $responseHeaders = $response->getHeaders();
-        if (! $isMultiHeader) {
-            // If we are not in "multi-header" mode, we must keep the last value only
-            // and cast it to string
-            foreach ($responseHeaders as $key => $value) {
-                $responseHeaders[$key] = end($value);
-            }
-        }
-
-        return array_change_key_case($responseHeaders, CASE_LOWER);
```
mnapoli, line 294: Headers are no returned unchanged because we normalize them later in any case. That simplifies the code.

src/Event/Http/HttpHandler.php
```diff
@@ -24,6 +24,10 @@ public function handle($event, Context $context): array
 
         $response = $this->handleRequest($httpEvent, $context);
 
+        if ($httpEvent->isFormatV2()) {
+            return $response->toApiGatewayFormatV2();
```
mnapoli, line 28: I've added a completely separate method to avoid too many `if`s and branches.

PR #843, opened 2021-01-20
Fixed docker entrypoint in documentation

Chtiadrien:
The `docker run` command is missing an overwrite of the entrypoint to work.


mnapoli:
🚀

Issue #847, opened 2021-01-23
seperateVendor option does not work if no environment variables are used


svnldwg:
**Description:**

If there are no environment variable defined in the serverless.yml `provider` block, the option `separateVendor` in the `custom.bref` block is not working. Deployment fails with this error message:

>  Type Error ---------------------------------------------
  TypeError: Cannot set property 'BREF_DOWNLOAD_VENDOR' of undefined
      at ServerlessPlugin.createVendorZip (/var/task/vendor/bref/bref/index.js:120:75)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:97:5)

**How to reproduce:**

Set the `seperateVendor` option and do not define any environment variables in the `provider` block.:

```
provider:
    name: aws
    stage: ${opt:stage, 'dev'}
    region: ${env:AWS_DEFAULT_REGION}
    runtime: provided.al2

// ...

custom:
    bref:
        separateVendor: true

svnldwg:
I think I know how this can be fixed and will open a pull-request for this bug shortly.

PR #848, opened 2021-01-23
initialize provider environment variables if there are none

svnldwg:
fixes bug #847 which causes deployment fail when no environment variables are defined and separateVendor option is used


mnapoli:
Thank you!

PR #853, opened 2021-01-29
Updated PDO-Sqlite link to go to the correct page.

GregHolmes:
On the Bref docs page: https://bref.sh/docs/environment/php.html#extensions-installed-and-enabled
Under `Extensions Installed and enabled`, the pdo_sqlite is throwing a 404, this is because it's a reference in the manual rather than a book page. I've corrected this to point to the correct URL.


mnapoli:
Thanks!

PR #855, opened 2021-02-04
Improve running PHP on Lambda with Docker

mnapoli:
- Improve the existing Docker images for running on Lambda
- Add `php-xx-console` Docker images for running CLI commands on Lambda

Usage example:

```Dockerfile
FROM bref/php-80-fpm

ADD . $LAMBDA_TASK_ROOT

CMD [ "index.php" ]
```

For console commands:

```Dockerfile
FROM bref/php-80-console

ADD . $LAMBDA_TASK_ROOT

CMD [ "bin/console" ]
```


soukicz:
This took me couple hours :)) Could you please update this article? https://mnapoli.fr/aws-lambda-php-docker-containers/

This one and the one on AWS blog are the only two I was able to find how to deploy bref docker image and both are with old syntax that doesn't work after this update.

(but the new syntax is definitely better!)

mnapoli:
For those interested I've published a quick gist here: https://twitter.com/matthieunapoli/status/1374369291135373316

I'll try to find time to update the article too.

PR #857, opened 2021-02-06
[docker] Sync fpm-dev and fpm settings

Maks3w:
These settings are available in the "fpm" image so they should replicated in the "fpm-dev" image too.


mnapoli:
Thank you!

PR #859, opened 2021-02-08
[LambdaRuntime] Use PHP >7.3 JSON errors

Maks3w:
Replace legacy JSON error management with the modern JsonException

Note: I've removed the phrase "Here is the original JSON error: '%s'" in favor of chain exception Exception::$previous


mnapoli:
Hi, I'm not sure why we would do that?

The downside I see is that an exception would add some overhead, but at this place this might not be really problematic. But I don't see a reason why change the existing code?

Maks3w:
ok. np

PR #865, opened 2021-02-14
Added Lambda and API Gateway request IDs to Cloudwatch logs when there's a PHP-FPM issue

aran112000:


src/Event/Http/FpmHandler.php
```diff
@@ -79,7 +84,7 @@ public function start(): void
         $this->fpm->setTimeout(null);
         $this->fpm->start(function ($type, $output): void {
             // Send any PHP-FPM log to CloudWatch
-            echo $output;
+            $this->logToCloudwatch($output);
```
mnapoli, line 87: I think we shouldn't edit this part: these are the logs coming from the PHP application (for the most part at least).

src/Event/Http/FpmHandler.php
```diff
@@ -282,4 +290,33 @@ private function getResponseHeaders(ProvidesResponseData $response): array
...
+     * Returns an array with AWS Request IDs for the current Lambda invocation and API Gateway (where applicable)
+     *
+     * @return string[]|array
+     */
+    private function getLogContext(): array
+    {
+        $logContext = [];
+
+        if ($this->lambdaRequestId) {
+            $logContext['lambdaRequestId'] = $this->lambdaRequestId;
```
mnapoli, line 313: We could use the same terminology as existing Lambda CloudWatch logs:

```suggestion
            $logContext['requestId'] = $this->lambdaRequestId;
```

image

mnapoli (commented):
Looks good to me, added some comments

src/Event/Http/FpmHandler.php
aran112000: They weren't previously going to CW due to no line ending from my tests, hence adding this based on the comment that was already there above indicating it should have been. Happy to drop through if you don't feel this is appropriate :)

src/Event/Http/FpmHandler.php
aran112000: My thinking here was that as there are potentially two different request IDs in the log here (if invoked via API Gateway) that it would make sense to be more explicit to avoid confusion, but if you'd rather remain with the existing terminology, I've no issue with that

src/Event/Http/FpmHandler.php
mnapoli: Yes this is intentional: we are forwarding the PHP logs as-is. If we were to add a line return, we would end up with 2 line returns whenever a PHP logger logs for example.

And in any case I don't think we should add the JSON part on every line of PHP logs (too much of a change for users).

src/Event/Http/FpmHandler.php
mnapoli: Right, I understand but yep I think it's better to rather stay in line with AWS naming.

src/Event/Http/FpmHandler.php
```diff
@@ -108,16 +113,19 @@ public function __destruct()
      */
     public function handleRequest(HttpRequestEvent $event, Context $context): HttpResponse
     {
+        $this->lambdaRequestId = $context->getAwsRequestId();
```
GrahamCampbell, line 116: Why are we mutating the handler state?

GrahamCampbell (changes requested):

src/Event/Http/FpmHandler.php
mnapoli: Yes, this is exactly where I stumbled when working on this locally.

I don't really want to store state in the class, but then that means passing it everywhere.

This led me to realize that in most cases there is no context associated to the log lines, since we boot PHP-FPM _before_ any request comes in.

In the end, I'm not 100% sure about the approach, and #895 may be able to remove the source issue.

GrahamCampbell: Yeh, I agree #895 or something like it, is really the best option here.

PR #871, opened 2021-02-27
Document S3 assets deployment for Symfony

t-richard:
This PR is following some discussion we had on Slack regarding the improvement of the Symfony section of Bref docs.

This is mostly pasted from the same part of the [Laravel section](https://bref.sh/docs/frameworks/laravel.html#assets), adapted for Symfony and Webpack Encore.

### DIFF
--- a/docs/frameworks/symfony.md
+++ b/docs/frameworks/symfony.md
@@ -137,16 +137,7 @@ To learn more about all this, read the [environment variables documentation](/do
 
 To deploy Symfony websites, we need assets to be served by AWS S3. Setting up the S3 bucket is already explained in the [websites documentation](../websites.md#hosting-static-files-with-s3). This section provides additional instructions specific to Symfony assets and Webpack Encore.
 
-First, you can compile assets for production in the `public` directory, then synchronize that directory to a S3 bucket:
-
-```bash
-php bin/console assets:install --env prod
-# if using Webpack Encore, additionally run
-yarn encore production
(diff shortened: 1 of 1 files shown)


t-richard:
Marked it as draft so that I can test the last code example about Webpack Encore config. Not using Encore this way so I'm unsure if that piece of config is correct for what we do here.

I looked at https://symfony.com/doc/current/frontend/encore/cdn.html but it's unclear to me what config we must have to generate a manifest saying

```
"build/dashboard.js": "https://<bucket-name>.s3.amazonaws.com/build/dashboard.js"
```

mnapoli (commented):
This is AWESOME!

I'll let you test that to make sure that it works, but it's absolutely perfect! Thank you!

mnapoli:
> but it's unclear to me what config we must have to generate a manifest saying

Oh sorry about that part I haven't used Encore + S3 so I'm not really sure.

t-richard:
Me neither, that's why I'll have to test this out 🙂

t-richard:
Ok, the config was wrong :sweat_smile: 

I've updated the code sample.

This is now generating a manifest with lines like 
```
"build/app.js": "https://test-encore.s3.amazonaws.com/build/app.7da5d4f5.js"
```

Which is what we want because we deploy the whole public folder.

There's an example of this in the Encore's code comments https://github.com/symfony/webpack-encore/blob/main/index.js#L84

mnapoli:
🚀  awesome thank you!

PR #872, opened 2021-02-27
Document trusted_proxies & trusted_headers for Symfony

t-richard:
This PR documents proxy config for Symfony.

This is needed for Symfony to generate URLs properly (eg. using the correct scheme which is problematic because API Gateway is only available through https).

More details here https://symfony.com/doc/current/deployment/proxies.html

### DIFF
--- a/docs/frameworks/symfony.md
+++ b/docs/frameworks/symfony.md
@@ -133,6 +133,32 @@ The secrets (e.g. database passwords) must however not be committed in this file
 
 To learn more about all this, read the [environment variables documentation](/docs/environment/variables.md).
 
+## Trust API Gateway
+
+When hosting your site on Lambda, API Gateway will act as a proxy between the client and your function. 
+
+By default, Symfony doesn't trust proxies for security reasons, but it's safe to do it when using API Gateway and
+Lambda. 
+
+This is needed because otherwise, Symfony will not be able to generate URLs properly.
+
+You should add the following lines to `config/packages/framework.yaml`
+
+```yaml
+framework:
+    # trust the remote address because API Gateway has no fixed IP or CIDR range that we can target
+    trusted_proxies: '127.0.0.1,REMOTE_ADDR'
+    # trust "X-Forwarded-*" headers coming from API Gateway
+    trusted_headers: ['x-forwarded-for', 'x-forwarded-proto', 'x-forwarded-port']
+```
+
+Note that API Gateway doesn't set the `X-Forwarded-Host` header, so we don't trust it by default. 
+You should only whitelist this header if you set it manually, for example in your CloudFront configuration.
+
+> Be careful with these settings if your app will not be executed only in a Lambda environment.
+ 
+You can get more details in the [Symfony documentation](https://symfony.com/doc/current/deployment/proxies.html).
+
 ## Assets
 
 To deploy Symfony websites, we need assets to be served by AWS S3. Setting up the S3 bucket is already explained in the [websites documentation](../websites.md#hosting-static-files-with-s3). This section provides additional instructions specific to Symfony assets and Webpack Encore.
--- a/docs/frameworks/symfony.md
+++ b/docs/frameworks/symfony.md
@@ -153,11 +153,19 @@ framework:
 ```
 
 Note that API Gateway doesn't set the `X-Forwarded-Host` header, so we don't trust it by default. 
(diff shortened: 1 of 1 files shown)


docs/frameworks/symfony.md
```diff
@@ -133,6 +133,29 @@ The secrets (e.g. database passwords) must however not be committed in this file
...
+
+This is needed because otherwise, Symfony will not be able to generate URLs properly.
+
+You should add the following lines to `config/packages/framework.yaml`
+
+```yaml
+framework:
+    # trust the remote address because API Gateway has no fixed IP or CIDR range that we can target
+    trusted_proxies: '127.0.0.1,REMOTE_ADDR'
+    # trust "X-Forwarded-*" headers coming from API Gateway (note that it doesn't send X-Forwarded-Host)
```
mnapoli, line 151: API Gateway doesn't send that header, but it's possible to send it manually when using CloudFront (which is something recommended in Bref's documentation, also you can see an example [here](https://github.com/brefphp/bref/issues/535#issuecomment-577158098)).

So I would recommend allowing `X-Forwarded-Host` too to be forward-compatible, in case the user wants to use CloudFront and sets up that header.

t-richard: IMHO this is a bad idea to enable it by default because we're not sure it will be configured and that can be a security problem.

From https://symfony.com/doc/current/deployment/proxies.html#solution-settrustedproxies
> Enabling the Request::HEADER_X_FORWARDED_HOST option exposes the application to [HTTP Host header attacks](https://www.skeletonscribe.net/2013/05/practical-http-host-header-attacks.html). Make sure the proxy really sends an x-forwarded-host header.

So this could expose the application to host poisoning attacks if they don't configure CloudFront to send that header. But maybe we should add a note to say that it should be enabled if CloudFront is configured properly ?

mnapoli: Very good point.

> But maybe we should add a note to say that it should be enabled if CloudFront is configured properly ?

👍  let's go with "disabled by default" (in the docs). Maybe a note like this:

> If you set the `X-Forwarded-Host` manually, for example in CloudFront configuration, you can also whitelist that header.

t-richard: Updated to include X-Forwarded-Host recommendations !

chekalsky:
I would also add disclaimer about CloudFront — this solution AFAIK won't work with it — Symfony will return CloudFront IP on getClientIp call. [Here](https://symfony.com/doc/current/deployment/proxies.html#but-what-if-the-ip-of-my-reverse-proxy-changes-constantly) you can find more on that (at the bottom).

t-richard:
That's really annoying indeed, but I feel like getting the client IP is not something you want on every app and making it to work properly is hard.

It sure is a problem but I don't think we can document a clean approach to this problem in this PR. I will add a disclaimer to make users aware of this issue as suggested by @chekalsky.

More details about my findings on this issue :arrow_down: 

@chekalsky shared his solution on Slack which is basically calling [an API provided by AWS](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/LocationsOfEdgeServers.html) listing all edge IPs and setting trusted proxies based on thoses IPs (and additionally storing these IPs in cache for performance).

This is also the approach taken by this Laravel middleware https://github.com/jdavidbakr/CloudfrontProxies/blob/master/src/CloudfrontProxies.php#L31

This looks cumbersome to document and is also not perfect because IPs could change while cached and it also means you need an external cache storage like Redis which implies additional infrastructure costs and maintenance.

Another approach would be the one offered by AWS [here](https://aws.amazon.com/fr/blogs/security/automatically-update-security-groups-for-amazon-cloudfront-ip-ranges-using-aws-lambda/)

> An Amazon Simple Notification Service (Amazon SNS) message is sent to a topic whenever the AWS IP ranges change. Enabling you to build an event-driven, serverless solution that updates the IP ranges for your security groups, as needed by using a Lambda function that is triggered in response to the SNS notification.

This approach avoid serving an expired range of IPs from cache and the additional infrastructure burden and would fit better with Bref because it relies on SNS events and Serverless function. But its harder to make our Symfony website Lambda aware of the change (ATM I don't have a solution to do it). 

LMKWYT :slightly_smiling_face:

t-richard:
Updated to add a disclaimer about Cloudfront and IP adresses

mnapoli:
Really nice work, thank you!

PR #874, opened 2021-03-01
Document using buildDir in Kernel.php

t-richard:
Symfony 5.2 introduced `Kernel::getBuildDir()` for read-only cache that can be prewarmed (like the container).

This is my attempt to document its usage with Bref.

I think it could also replace the `prepareCacheDir` of the [symfony-bridge](https://github.com/brefphp/symfony-bridge/blob/master/src/BrefKernel.php#L64) but I might be missing something. Hope you can give me some more information about it. 

Just to give you an insight on the performance boost for cold starts, here are my benchmark on a `symfony/skeleton` app (using `ab -c 20 -n 20`).

Using the currently documented approach
```
Percentage of the requests served within a certain time (ms)
  50%   1549
  66%   1641
  75%   1732
  80%   1771
  90%   1853
  95%   1860
  98%   1860
  99%   1860
 100%   1860
```

Using the config added by this PR
```
Percentage of the requests served within a certain time (ms)
  50%    960
  66%    972
  75%    984
  80%    986
  90%    998
  95%   1147
  98%   1147
  99%   1147
 100%   1147
```

Let me know what you think :slightly_smiling_face:

### DIFF
--- a/docs/frameworks/symfony.md
+++ b/docs/frameworks/symfony.md
@@ -37,6 +37,8 @@ package:
         - node_modules/**
         - tests/**
         - var/**
+    include:
+        - var/cache/prod # allows to deploy a pre-warmed container
 
 functions:
     website:
@@ -56,7 +58,11 @@ functions:
 
 We still have a few modifications to do on the application to make it compatible with AWS Lambda.
 
-Since [the filesystem is readonly](/docs/environment/storage.md) except for `/tmp` we need to customize where the cache and the logs are stored in the `src/Kernel.php` file. This is done by adding 2 new methods to the `Kernel` class:
+Since [the filesystem is readonly](/docs/environment/storage.md) except for `/tmp` we need to customize where the cache and the logs are stored in the `src/Kernel.php` file. This is done by adding 3 new methods to the `Kernel` class. 
+We still want our [build directory](https://symfony.com/doc/current/reference/configuration/kernel.html#build-directory) 
+to be readonly so we make it use the default `var/cache` directory.
+
+> Note : having the build directory readonly on Lambda means you have to pre-warm the cache as part of your deploy process
 
 ```php
     public function getLogDir()
@@ -78,9 +84,20 @@ Since [the filesystem is readonly](/docs/environment/storage.md) except for `/tm
 
         return parent::getCacheDir();
     }
+
+    public function getBuildDir(): string
+    {
+        return $this->getProjectDir().'/var/cache/'.$this->environment;
+    }
 ```
 
 ## Deploy
+
+Pre-warm the Symfony cache before deploying
+
+```bash
+php bin/console cache:clear --env=prod
+```
 
 The application is now ready to be deployed. Follow [the deployment guide](/docs/deploy.md).
 
--- a/docs/frameworks/symfony.md
+++ b/docs/frameworks/symfony.md
@@ -58,11 +58,7 @@ functions:
 
 We still have a few modifications to do on the application to make it compatible with AWS Lambda.
 
-Since [the filesystem is readonly](/docs/environment/storage.md) except for `/tmp` we need to customize where the cache and the logs are stored in the `src/Kernel.php` file. This is done by adding 3 new methods to the `Kernel` class. 
-We still want our [build directory](https://symfony.com/doc/current/reference/configuration/kernel.html#build-directory) 
-to be readonly so we make it use the default `var/cache` directory.
-
-> Note : having the build directory readonly on Lambda means you have to pre-warm the cache as part of your deploy process
+Since [the filesystem is readonly](/docs/environment/storage.md) except for `/tmp` we need to customize where the cache and the logs are stored in the `src/Kernel.php` file. This is done by adding 2 new methods to the `Kernel` class:
 
 ```php
     public function getLogDir()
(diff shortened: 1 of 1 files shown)


mnapoli:
Awesome, looks really good!

I'm wondering about those not on Symfony 5.2: will that new documentation break something for them, or could it confuse them?

I don't think it will break anything, because if they implement the new `getBuildDir()` method it will simply be ignored. They will deploy the compiled cache, but this cache will not be used. So in the end, maybe a simple note could be added to the docs? Let me know what you think.

t-richard:
> I don't think it will break anything, because if they implement the new getBuildDir() method it will simply be ignored. They will deploy the compiled cache, but this cache will not be used.

Yes I think this is what would happen. So the current phrasing might be misleading for them. Maybe the least confusing approach would be to split this in two paragraphs and clearly say that this will work only starting from Symfony 5.2

The downside of this is that we dont have something to offer to users using an older version for deploying a warmed cache except what is done in the symfony-bridge. But maybe that's OK, because using Serverless/Bref is a choice generally made at the beginning of a new project, and most projects will start with a Symfony 5.2+ version from now on.

t-richard:
I tested further this approach and it seems this is not working as expected. 

When the container is compiled using `cache:clear` (or `cache:warmup`), the `cache_dir` and `build_dir` values will be "hardcoded" in the container. Because of this, when building the container locally (or in CI), `LAMBDA_TASK_ROOT` will not be available and the value in the container will be 

```
'kernel.cache_dir' => (\dirname(__DIR__, 3).'/cache/prod'),
```

if running `LAMBDA_TASK_ROOT=test bin/console cache:warmup --env=prod` then the value will be

```
'kernel.cache_dir' => '/tmp/cache/prod',
```

Which is what is expected, anything else works as expected.

@mnapoli what do you think would be the best approach to tackle this ? It seems weird to me to suggest running the command with `LAMBDA_TASK_ROOT` set in the environment.

docs/frameworks/symfony.md
```diff
@@ -56,7 +58,11 @@ functions:
 
 We still have a few modifications to do on the application to make it compatible with AWS Lambda.
 
-Since [the filesystem is readonly](/docs/environment/storage.md) except for `/tmp` we need to customize where the cache and the logs are stored in the `src/Kernel.php` file. This is done by adding 2 new methods to the `Kernel` class:
+Since [the filesystem is readonly](/docs/environment/storage.md) except for `/tmp` we need to customize where the cache and the logs are stored in the `src/Kernel.php` file. This is done by adding 3 new methods to the `Kernel` class.
```
Nyholm, line 61: Just say 2 new methods.

docs/frameworks/symfony.md
```diff
@@ -56,7 +58,11 @@ functions:
 
 We still have a few modifications to do on the application to make it compatible with AWS Lambda.
 
-Since [the filesystem is readonly](/docs/environment/storage.md) except for `/tmp` we need to customize where the cache and the logs are stored in the `src/Kernel.php` file. This is done by adding 2 new methods to the `Kernel` class:
+Since [the filesystem is readonly](/docs/environment/storage.md) except for `/tmp` we need to customize where the cache and the logs are stored in the `src/Kernel.php` file. This is done by adding 3 new methods to the `Kernel` class. 
+We still want our [build directory](https://symfony.com/doc/current/reference/configuration/kernel.html#build-directory)
```
Nyholm, line 62: Maybe something like: 
```suggestion
To be able to deploy our application with a warm cache, we need to specify the [build directory](https://symfony.com/doc/current/reference/configuration/kernel.html#build-directory) to be separate from the cache directory.
```

Then explain that you need to warm the cache in the deploy process.

Nyholm (changes requested):
Thank you. 

Here are some rough suggestions. They are not perfect and they probably need to be reworded. But I think it makes sense to say: 

"Add this 2 methods to move the cache and logs. Do optimisations to have a warm cache by adding a third method".

Nyholm:
> @mnapoli what do you think would be the best approach to tackle this ? It seems weird to me to suggest running the command with LAMBDA_TASK_ROOT set in the environment.

I think this is the best approach. Let's add `LAMBDA_TASK_ROOT` to the "warm up" command.

t-richard:
Thanks for your review and feedback @Nyholm :pray: 

I'll try to work on this evening :slightly_smiling_face:

docs/frameworks/symfony.md
mnapoli: But it's actually 3 right? (it includes the "getBuildDir" method)

mnapoli: Oh nevermind I finally got what you were trying to say:

- change 2 methods
- **if** you want optimization, do this extra thing

Makes sense, +1 on your suggested changes

mnapoli:
> I think this is the best approach. Let's add LAMBDA_TASK_ROOT to the "warm up" command.

+1

It is far from ideal IMO because it's very confusing, but now we have a clear path to creating a simple and efficient "Symfony Bridge" (base Kernel class that does all that job automatically).

So I'd rather not waste time on improving this solution: let's document how to make it work.

Next step would be to update https://github.com/brefphp/symfony-bridge to automatically do the cache/build dir thing properly.

t-richard:
Thanks both for your feedback.

I've update the PR !

mnapoli:
Thanks!

chekalsky:
@mnapoli @t-richard this approach with BuildDir doesn't work when built on macOS (and in other circumstances):
- `getCacheDir()` from our Kernel returns `/tmp/cache/prod` as it should
- Symfony [does](https://github.com/symfony/symfony/blob/5.x/src/Symfony/Component/HttpKernel/Kernel.php#L612) `realpath()` over it
- It becomes `/private/tmp/cache/prod` on macOS (because `/tmp` is a symlink)
- Nothing works! 💥

Nyholm:
Hm.. Interesting. I'll try to dig around more on this within the next few days.

PR #887, opened 2021-03-18
Test build docker images on PRs

Nyholm:
This build is quite slow, but it is only executed when we make changes to files in `runtime/`.


mnapoli:
Oh nice, TIL about `paths` filtering.

Nyholm:
=)

Thank you for merging

PR #889, opened 2021-03-20
Adding Layer to support Symfony Runtime component

Nyholm:
This may be premature, but I wrote the code and pushed a layer here: `arn:aws:lambda:eu-central-1:403367587399:layer:bref-symfony-runtime:3`

Here is some information about code that I think should eventually move into Bref itself: https://github.com/php-runtime/bref
Here is a small Symfony example application: https://github.com/Nyholm/bref-runtime-demo

See this PR more as a discussion rather than "Merge this now". 

With this PR you are able to run any PHP application on Bref by: 

```
composer install runtime/bref
```

```diff
 # serverless.yml

 functions:
     web:
         handler: public/index.php
         timeout: 10
         layers:
-            - ${bref:layer.php-80}
+            - ${bref:layer.php-80-runtime}
```

Benefits includes: 
- We can solve the timeout problem (started in #895)
- Automatically handle "trusted proxies" in Symfony and Laravel applications. ([source](https://github.com/php-runtime/bref/blob/0.2.0/src/SymfonyRequestBridge.php#L25))
- No need for special layer for FPM, Console or FPM-dev. Ie less choices.


Nyholm:
FYI 1: I added https://github.com/brefphp/extra-php-extensions/pull/231

FYI 2: I'll do a big presentation about this on Symfony Online in about 4 weeks.

mnapoli:
Hey @Nyholm, adding that in bref-extra sounds like a good solution for now! At least the layer is available/usable, which is great to gauge usefulness and adoption.

mnapoli:
🎉  Symfony runtime is now implemented and supported in https://github.com/brefphp/symfony-bridge

I've released a beta version, there should be a stable in the coming weeks!

PR #890, opened 2021-03-22
Support custom http warming

GrahamCampbell:
When warming a Laravel HTTP handler, one would want to "bootstrap" the kernel as part of the warming process. This is possible by totally overriding the handle method.

### DIFF
--- a/src/Event/Http/HttpHandler.php
+++ b/src/Event/Http/HttpHandler.php
@@ -8,8 +8,11 @@ use Bref\Event\Handler;
 abstract class HttpHandler implements Handler
 {
     abstract public function handleRequest(HttpRequestEvent $event, Context $context): HttpResponse;
-    
-    public function handleWarmer(HttpRequestEvent $event, Context $context): void
+
+    /**
+     * @param mixed $event The raw event data.
+     */
+    public function handleWarmer($event, Context $context): void
     {
         // Delay the response to ensure concurrent invocation
         // See https://github.com/brefphp/bref/pull/734
--- a/src/Event/Http/HttpHandler.php
+++ b/src/Event/Http/HttpHandler.php
@@ -8,15 +8,21 @@ use Bref\Event\Handler;
 abstract class HttpHandler implements Handler
 {
     abstract public function handleRequest(HttpRequestEvent $event, Context $context): HttpResponse;
+    
+    public function handleWarmer(HttpRequestEvent $event, Context $context): void
+    {
+        // Delay the response to ensure concurrent invocation
+        // See https://github.com/brefphp/bref/pull/734
+        usleep(10000); // 10ms
+    }
 
     /** {@inheritDoc} */
     public function handle($event, Context $context): array
     {
         // See https://bref.sh/docs/runtimes/http.html#cold-starts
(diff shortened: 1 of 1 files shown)


deleugpn:
Just out of curiosity, why would you want to bootstrap the framework?

GrahamCampbell:
So that it doesn't have to boot when the first real request comes in - it's already booted. I was thinking something like:

```php
    public function handleWarmer($event, Context $context): void
    {
        // We ensure the warming process takes at least 10ms to ensure concurrent
        // invocation. See https://github.com/brefphp/bref/pull/734 for more.
        $startTime = hrtime(true) / 1000;

        $this->kernel->bootstrap();

        $bootTime = (hrtime(true) / 1000) - $startTime;

        if ($bootTime < 10000) {
            usleep((int) (10000 - $bootTime));
        }
    }
```

GrahamCampbell:
NB That code would be in https://github.com/php-runtime/runtime/blob/main/src/bref/src/LaravelHttpHandler.php.

dhrrgn:
I'm confused why this is needed? 

The same thing can be accomplished as-is by overriding `handle`:

```php
    public function handle($event, Context $context): array
    {
        if ($event['warmer'] ?? false) {
            // perform custom warmer stuff
        }

        return parent::handle($event, $context);
    }
```

This still sleeps for the requisite 10ms in the base handler, and lets you do some special warmup.

GrahamCampbell:
Because your code sleeps for too long now. See the example code in my upper post.

GrahamCampbell:
10ms is a non-trivial amount of time to waste, when my handlers only takes 3ms to run...

dhrrgn:
My point still stands, you can just return early then:

```php
public function handle($event, Context $context): array
{
    if ($event['warmer'] ?? false) {
        // We ensure the warming process takes at least 10ms to ensure concurrent
        // invocation. See https://github.com/brefphp/bref/pull/734 for more.
        $startTime = hrtime(true) / 1000;

        $this->kernel->bootstrap();

        $bootTime = (hrtime(true) / 1000) - $startTime;

        if ($bootTime < 10000) {
            usleep((int) (10000 - $bootTime));
        }
        
        return ['Lambda is warm'];
    }

    return parent::handle($event, $context);
}
```

GrahamCampbell:
That's a kinda hacky solution, though. It leaks the logic of checking `$event['warmer']`.

mnapoli:
Hi! I think I understand @GrahamCampbell that you want something a bit more "officially supported".

At the moment, I think the use case is thin: warming up the kernel like you do is an interesting idea, but I'm not sure it will provide any benefit in real scenarios.

Until we are certain that this brings noticeable benefits, I think it's risky to add a new API (that we'll need to support) to Bref for that use case.

@dhrrgn's solution is a very good alternative: the `warmer` key is actually up to you to define in `serverless.yml`, so if you have custom logic to handle it, you can name it however you want. It doesn't leak anything IMO.

```php
public function handle($event, Context $context): array
{
    if ($event['any-key-you-want'] ?? false) {
        // do whatever you want
        return [...];
    }

    return parent::handle($event, $context);
}
```

As always, let's revisit this once we have a clear use case and clear benefits. If you see performance differences, I'm definitely interested!

Issue #873, opened 2021-02-28
PHP lambda timing out


jmuraca:
Not sure if related to other time out tickets here so raising a new issue.

I'm running php-74-fpm on lambda and with a 25 second timeout, 2048 memory, and seeing a very consistent stream of timeouts at 25seconds. The bulk of the calls to my API are very quick (<1 second) and the function code is relatively tight/small. I've tried to trace it through the code or reduce database connections, but it seems to be getting close to an environment issue. 

![image](https://user-images.githubusercontent.com/1421873/109425100-839e1800-799b-11eb-89ef-1707cf3d16a7.png)

![image](https://user-images.githubusercontent.com/1421873/109425251-438b6500-799c-11eb-9f46-2d56e5260a7a.png)

Is there any way I can turn on additional logging or how can I help to provide more feedback?

marcusirgens:
Do you have any dependencies downstream that might be timing out?

jmuraca:
I have an RDS database connected and Cloudwatch logging.  I reduced database calls and the database seems to have extra capacity. 

I can't replicate time outs on the staging or local environments. I'm not sure if that's a good or bad thing though. It could be that having many more requests on product is causing something else to timeout or be blocked, but I'm also trying to isolate if having more invocations of bref means the issue happens more frequently.

jmuraca:
Is there a way I could correlate a run of the bref instance with the application runtime instance? That way I could step through application logs to see if it actually ends or hangs at a point, compared to the bref instance timing out.

matthewsuan:
Try to enable xray tracing on your lambda function. I usually do it via:

```
provider:
  tracing:
    lambda: true
```

In Xray, you can see the entire invocation lifecycle and the cause the of timeout.

mnapoli:
Could it be that you have a piece of code that calls an API that times out?

Usually that's something like Sentry/bugsnag (which would timeout when there's an error to report). Related: does your VPC has a NAT Gateway?

jmuraca:
Turned on Xray for the past day. I didn't find it particularly useful since I'm building with Bref, single API Gateway function, into a single PHP instance with a Router internal. I haven't looked if/how this should be broken up yet as was hoping to move this to Lambda from EC2's
![image](https://user-images.githubusercontent.com/1421873/109565664-a57cc500-7a97-11eb-86f0-0bcdbd7e8dc9.png)

[...]

Issue #862, opened 2021-02-10
Handle Lambda timeouts and PHP-FPM crashes better


aran112000:
One of our Lambda functions running in eu-west-2 started getting intermittant errors logging in Cloudwatch stating:
"Error: signal: terminated Extension.Crash"

When this occurs, none of our application code is executed and the request just fails after ~2 seconds with API Gateway then returning a 500 error on the affected requests.

Our function only runs 3 layers and has done so for months, the issues started for 2 hours on 2021-02-08, then again around the same time in the evening last night 2021-02-09:
 - ${bref:layer.php-74-fpm}
 - ${bref:extra.memcached-php-74}
 - ${bref:extra.redis-php-74}

For clarity, there have been no changes to the application or environment by ourselves during these times and we have no time-based logic that would account for the issues starting in the evening for a few hours two days running.

Not sure if this is a Bref / Bref Extra issue or an AWS one, the Lambda team have investigated and believe it is to do with the layers, so any help/guideance to track things down further would be greatly appreciated.

evoaran:
We've managed to track down the issue to PHP-FPM having issues as can be seen from an extract from the logs below, has anyone encountered this before? For us, we've had it the last two nights only then it seems to eventually resolve itself but affecting over 1k invocations sporadically.

![image](https://user-images.githubusercontent.com/12773460/107547706-a4a5f100-6bc5-11eb-911e-e061a1819396.png)

matthewsuan:
@evoaran we are experiencing this issue as well. It fixes itself i think when lambda spawn another micro container instance.

mnapoli:
Hi! I haven't forgotten about this, I'm just short of ideas on what to do here. FYI there's also #842 that could be related, it may be worth giving it a look.

> it seems to eventually resolve itself but affecting over 1k invocations sporadically.

Right, as @matthewsuan mentioned this might be a case of the Lambda instance that is in a "broken" state is replaced by another Lambda instance.

Now the questions would be:

- what causes Lambda/PHP-FPM to enter a broken state?
- how can we improve Bref so that we never enter a broken state? (i.e. OK to have a failed request, but not the whole container)

Judging from #842 and past experience, could the initial failures be triggered by a large file upload or large HTTP response?

aran112000:
Hi @mnapoli, @matthewsuan,

After combing through the logs again this evening, I think I've found the cause that starts it each time - a Lambda timeout:
> Task timed out after 28.53 seconds

[...]

Issue #55, opened 2018-08-16
Gracefully stop the PHP process before max execution time is reached


t-geindre:
If our lambda reach the maximum execution time, it gets shut down. Since we use an external log system, we don't get any log in that case and we still have to check on cloudwatch what happened.

We could use `set_time_limit()` or `ini_set()` with the `max_execution_time` directive. But the issue with those functions is that you can't take control back when the limit is reached, the PHP process is immediatly stopped. The other issue is that systems calls are not taken in account (eg `sleep()`, `file_get_contents()`, etc.).

So I'm trying to find a solution to handle that case.

I had the following idea: we could configure a specific timeout  (obviously shorter than the lambda max execution time) which, when it's reached, trigger a signal sending to the PHP process (eg `SIGTERM`).

This could let some time to the PHP process to gracefully shut down and do whatever is required during the time left (eg log what happened).

And this should obviously be handled in the parent process, the javascript one (and I'm not sure to be skilled enough in javascript to achive that).

This is just an idea, any other idea is very welcome.

t-geindre:
I made a simple POC to validate my idea.

Here a simple PHP script which handle signals and runs forever:
```php
<?php
declare(ticks = 1);

class SigtermException extends \Exception {}

pcntl_signal(SIGTERM, function () {
	throw new SigtermException('SIGTERM received');
});


set_exception_handler(function ($exception) {
	// Log exception as usual
	echo "Exception thrown: " , $exception->getMessage(), "\n";
	// Is it's a Sigterm one, stop execution
	if ($exception instanceof SigtermException) {
		exit(1);
	}
});


for(;;){
	echo "Running\n";
	usleep(100000);
}
```

And here is the javascript handler which runs the PHP script:
```javascript
const spawn = require('child_process').spawn;

var script = spawn('php', ['test.php']);
var timeoutHandler = setTimeout(() => script.kill('SIGTERM'), 2000);

script.on('close', (code) => {
	console.log('Process closed with code: ' + code);
	clearTimeout(timeoutHandler);
});

script.stdout.on('data', function(data) {
    console.log(data.toString());
});

script.stderr.on('data', function(data) {
    console.log('[STDERR] ' + data.toString());
});
```

And finally the execution output:
```bash
$ nodejs handler.js
Running

Running

[...]

Running

Exception thrown: SIGTERM received

Process closed with code: 1
```

So it seems to be working as expected and would allows us to to gracefully shut down the PHP process before the lambda shuts down.

[...]

PR #895, opened 2021-03-25
POC handle timeouts

Nyholm:
This PR builds on the work started by @t-geindre in #55. 

If we are 1 second away from Lambda shutting down, then thrown an exception to let the application shutdown and (more importantly) flush the logs. 

There are two things Im not super happy with: 
1. This does not work for PHP-FPM since it the application run in a child process
2. I get a segmentation fault when SIGALRM is triggers while the application is waiting for IO

The question is: do we want to continue down this path?

It would solve issues like #873 and #862.


deleugpn:
> I get a segmentation fault when SIGALRM is triggers while the application is waiting for IO

Isn't every connection timeout an IO operation that PHP is waiting?

> This does not work for PHP-FPM since it the application run in a child process

This seems to me like a strong enough limitation that may indicate it's not worth it to support it? It does bring a nice idea to the table though: what if we offer a Debugging Layer where you can write a basic function invocation and run the code that's getting timeout from it so that you might get a better idea where it's stopping?
By making it an explicit debugging layer, we don't need to guarantee production-ready quality and perhaps we can even try to include default debugging statements that will get logged?

mnapoli:
Hey, I wonder if something similar could be used in the PHP-FPM worker (the process handling the request)?

That could be done as a Symfony/Laravel package?

I'm not familiar enough with signals like that to know if that could work.

Also another solution is to find a way for logs to be streamed correctly. Or else, find a way to force shorter timeouts on all connections in all frameworks, but I have no idea if that's even remotely possible.

Nyholm:
I reorganized things. 

I've updated `LambdaRuntime` to get a bool parameter to enable throwing `TimeoutException`. This is enabled in the bootstrap files in `php-xx` and console. It is not enabled in the `php-fpm`. For FPM the user needs to update their application by adding `Bref\Timeout\Timeout::enable()`. This will read the environment variables and set the timeout. 

I think this is a good way forward. I would like to get another review before I add tests and docs.

Nyholm:
You may also control this behavior with `BREF_TIMEOUT`. 
Value 0 is "auto". Value "-1" is disabled. Positive values sets the timeout.

src/Runtime/LambdaRuntime.php
```diff
@@ -42,19 +43,33 @@ final class LambdaRuntime
...
     private $invoker;
 
-    public static function fromEnvironmentVariable(): self
+    /** @var int seconds */
+    private $timeout;
+
+    public static function fromEnvironmentVariable(?int $timeout = null): self
     {
-        return new self((string) getenv('AWS_LAMBDA_RUNTIME_API'));
+        return new self((string) getenv('AWS_LAMBDA_RUNTIME_API'), $timeout ?? (int) getenv('BREF_TIMEOUT'));
```
Nyholm, line 51: If the BREF_TIMEOUT env var does not exist, the `getenv()` method will return `false`. And `(int) false` will be 0.

src/Runtime/LambdaRuntime.php
```diff
@@ -96,6 +111,13 @@ public function processNextEvent($handler): void
         [$event, $context] = $this->waitNextInvocation();
         \assert($context instanceof Context);
 
+        if ($this->timeout > 0) {
+            Timeout::timoutAfter($this->timeout);
+        } elseif ($this->timeout === 0 && 0 < $context->getRemainingTimeInMillis()) {
+            // Throw exception one second before Lambda pulls the plug.
+            Timeout::timoutAfter((int) floor($context->getRemainingTimeInMillis() / 1000) - 1);
```
GrahamCampbell, line 118: This could be negative if `$context->getRemainingTimeInMillis()` is between 1 and 999.

GrahamCampbell (changes requested):

tests/Timeout/TimeoutTest.php
```diff
@@ -0,0 +1,101 @@
...
+        Timeout::reset();
+        parent::tearDown();
+    }
+
+    public function testEnableWithoutContext()
+    {
+        $this->expectException(\LogicException::class);
+        Timeout::enable();
+        $timeout = pcntl_alarm(0);
+        $this->assertGreaterThan(20, $timeout, 'If not');
```
Nyholm, line 36: These two lines can be removed

tests/Timeout/TimeoutTest.php
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    public function testEnableWithBrefTimeoutInactive()
+    {
+        $_SERVER['BREF_TIMEOUT'] = -1;
+        $_SERVER['LAMBDA_INVOCATION_CONTEXT'] = json_encode(['deadlineMs' => (time() + 30) * 1000]);
+
+        Timeout::enable();
+        $timeout = pcntl_alarm(10);
```
Nyholm, line 45: Change to 0

src/Runtime/LambdaRuntime.php
Nyholm: Thank you. I solved this with `max(1, ...)`

Nyholm:
A PHP-FPM user would have to write something like: 

```php
if (isset($_SERVER['LAMBDA_TASK_ROOT'])) {
    Bref\Timeout\Timeout::enable();
}
```

Preferably as soon as possible, ie `Kernel::handle()` or `index.php`.

src/Runtime/LambdaRuntime.php
Nyholm: This is covered with tests

src/Timeout/Timeout.php
```diff
@@ -0,0 +1,92 @@
...
+
+        self::$initialized = true;
+
+        return true;
+    }
+
+    /**
+     * Set a timer to throw an exception.
+     */
+    public static function timoutAfter(int $seconds): void
```
mnapoli, line 77: Small typo:

```diff
- timoutAfter
+ timeoutAfter
```

Nyholm: Thank you

src/Runtime/LambdaRuntime.php
```diff
@@ -42,19 +43,33 @@ final class LambdaRuntime
     /** @var Invoker */
     private $invoker;
 
-    public static function fromEnvironmentVariable(): self
+    /** @var int seconds */
+    private $timeout;
+
+    public static function fromEnvironmentVariable(?int $timeout = null): self
```
mnapoli, line 49: I'm not sure we need to make the timeout configurable. After all, since it's based on the Lambda timeout users can customize the timeout via Lambda's configuration.

How about a boolean: enabled (true/false)?

Nyholm: I did a boolean at first. But that did not allow me to always use a Lambda timeout of 28 seconds and just "play" with `BREF_TIMEOUT`. 

The edge case is that my "shutdown process" may take 5 seconds.

src/Timeout/Timeout.php
```diff
@@ -0,0 +1,92 @@
...
+        throw new \LogicException('Could not find value for bref timeout. Are we running on Lambda?');
+    }
+
+    /**
+     * Setup custom handler for SIGTERM. One need to call Timeout::timoutAfter()
+     * to make an exception to be thrown.
+     *
+     * @return bool true if successful.
+     */
+    public static function init(): bool
```
mnapoli, line 53: `init()` is called already by `timeoutAfter()`, how about making it private?

That would simplify the API of the `Timeout` class, and simplify the code in `LambdaRuntime`.

Nyholm: But then the LambdaRuntime would not know if the timeout failed to setup. Currently `LambdaRuntime` disables the "exception throwing" if init fails. Ie, it will not try to init over and over again. 

https://github.com/brefphp/bref/pull/895/files#diff-042bba61d17e1d04b45e10c2440aab63bc6e32d5ca42c57fa67d5e9d56512f2cR69-R72

Making `init()` private would fill up the logs with potential error messages. 

....
Though.. one could argue that pcntl is always available and you would "only" get one error message per request.

Nyholm: I did remove it.. the patch did not have much effect on `LambdaRuntime` 

https://github.com/brefphp/bref/pull/895/commits/52f8d65f665f7f0be6ee5b0de13a70694466c1e4

src/Runtime/LambdaRuntime.php
mnapoli: Not sure I see what you mean by "shutdown process"?

> But that did not allow me to always use a Lambda timeout of 28 seconds and just "play" with BREF_TIMEOUT.

I don't think that's something we need to support. If I understand correctly, that would mean the timeout is entirely controlled by the Lambda timeout, which sounds fine to me. Less settings, less code paths, less options is better. At least unless we have a clear use case for allowing 2 different controls (Lambda timeout + env variable)?

mnapoli: Forgot about the fact that at first it's good to make the feature opt-in (i.e. it's a bit "experimental" until we enable it by default).

So how about:

- `BREF_TIMEOUT` undefined -> no Bref timeout
- `BREF_TIMEOUT=1` -> automatic Bref timeout

If all works as expected, we can them remove that feature flag and enable it by default in the future?

I just don't see a use case for a custom timeout on top of the already customisable Lambda timeout.

Nyholm: > Not sure I see what you mean by "shutdown process"?

Whenever an exception is thrown. My application will flush the logs, maybe write something to the DB and maybe send an email to admin. This could take 3 seconds. This is my "shutdown process". If we don't allow to customise the time to when we thrown an exception, then I cannot use this feature. 

-------

> Less settings, less code paths, less options is better.

We still need a way to opt-out from this feature, right? If so, we still need an environment variable. 

------

Sure let's do it as a feature flag to opt-in to this feature in version 1.3. But I would do 

* `BREF_TIMEOUT` undefined => no Bref timeout
* `BREF_TIMEOUT=0` => automatic Bref timeout
* `BREF_TIMEOUT=-1` => no Bref timeout

Then in version 1.x
* `BREF_TIMEOUT` undefined => automatic Bref timeout

mnapoli:
Thanks, following our discussions I've opened https://github.com/Nyholm/bref/pull/3 to remove the option of configuring the duration timeout via an environment variable.

Nyholm:
The support PR is merged. This PR is now ready for a final rounds of review.

Nyholm (commented):
Im :+1:

We don't mention that the logs from php-fpm also will be flushed, but that is a technical detail I think it is okey to skip.

src/Timeout/Timeout.php
```diff
@@ -66,7 +67,23 @@ private static function init(): void
...
         pcntl_signal(SIGALRM, function (): void {
-            throw new LambdaTimeout('Maximum AWS Lambda execution time reached');
+            if (Timeout::$stackTrace !== null) {
+                // we already thrown an exception.
+                error_log('Lambda timed out');
+                error_log((new LambdaTimeout())->getTraceAsString());
+                error_log('Original stack trace');
+                error_log(Timeout::$stackTrace);
+                Timeout::$stackTrace = null;
+                exit(1);
```
mnapoli, line 77: Should we force a `500` status code before exit?

That would only be taken into account in the FPM layer, so that wouldn't really impact the `php-XX` layer.

src/Timeout/Timeout.php
```diff
@@ -0,0 +1,103 @@
...
+    /**
+     * @internal
+     */
+    public static function enable(int $remainingTimeInMillis): void
+    {
+        self::init();
+
+        $remainingTimeInSeconds = (int) floor($remainingTimeInMillis / 1000);
+
+        // The script will timeout 1 second before the remaining time
```
mnapoli, line 42:
```suggestion
        // The script will timeout 2 seconds before the remaining time
```

src/Timeout/Timeout.php
Nyholm: How? 
And wouldn't that effect my `AWS::ApiGateway::GatewayResponse` that triggers on timeouts or 500?

mnapoli:
```php
http_response_code(500);
```

> And wouldn't that effect my AWS::ApiGateway::GatewayResponse that triggers on timeouts or 500?

🤔 not sure about that, maybe. But in any case, the same would happen if your app returns a pretty 500 error page.

Does that make sense? What I mean is that, when using FPM, there would be no difference between:

- Symfony catches any exception and displays an error page with a 500 status code
- Bref timeout kicks in and forces a 500 status code

In both cases, API Gateway would behave the same way.

Nyholm: Oh, no that is not true. 

> Symfony catches any exception and displays an error page with a 500 status code

This will show you a nice error page. API Gateway does nothing

> Bref timeout kicks in and forces a 500 status code

No content(HTML) is provided. I would like the custom error page in API Gateway to be used. 

-------

Im not sure if `http_response_code(500)` would make any difference. 


-------

Also note. We are not sure it is a HTTP context. It could be any other event.. So Im not sure about sending a HTTP status code.

src/Timeout/Timeout.php
```diff
@@ -0,0 +1,108 @@
...
+        pcntl_async_signals(true);
+        // Setup a handler for SIGALRM that throws an exception
+        // This will interrupt any running PHP code, including `sleep()` or code stuck waiting for I/O.
+        pcntl_signal(SIGALRM, function (): void {
+            if (Timeout::$stackTrace !== null) {
+                // we already thrown an exception, do a harder exit.
+                error_log('Lambda timed out');
+                error_log((new LambdaTimeout)->getTraceAsString());
+                error_log('Original stack trace');
+                error_log(Timeout::$stackTrace);
```
mnapoli, line 79: For the record, I'm testing and notice that `error_log` adds noise in FPM:

```
NOTICE: PHP message: Lambda timed out
NOTICE: PHP message: #0 /var/task/demo/http.php(16): Bref\Timeout\Timeout::Bref\Timeout\{closure}()
#1 {main}
NOTICE: PHP message: Original stack trace
NOTICE: PHP message: #0 /var/task/demo/http.php(9): Bref\Timeout\Timeout::Bref\Timeout\{closure}()
#1 {main}
```

I'm changing that to:

```php
                $stderr = fopen('php://stderr', 'w');
                fwrite($stderr, 'Lambda timed out' . PHP_EOL);
                fwrite($stderr, (new LambdaTimeout)->getTraceAsString() . PHP_EOL);
                fwrite($stderr, 'Original stack trace' . PHP_EOL);
                fwrite($stderr, Timeout::$stackTrace . PHP_EOL);
                fclose($stderr);
```

Which gives a cleaner log:

```
Lambda timed out
#0 /var/task/demo/http.php(16): Bref\Timeout\Timeout::Bref\Timeout\{closure}()
#1 {main}
Original stack trace
#0 /var/task/demo/http.php(9): Bref\Timeout\Timeout::Bref\Timeout\{closure}()
#1 {main}
```

No need to update the PR

Nyholm: Oh, I didnt know. 

Thank you for testing.

mnapoli:
OK following discussion through Slack, we were happy with the PR.

However I tested and I'm getting weird results in FPM too:

- 1st request: the timeouts work as expected
- following requests: `pcntl_alarm(...)` is triggered, but our custom handler isn't called. That causes pending I/O (or sleep) to be interrupted, but without any exception being triggered

I've been testing locally with PHP-FPM to pinpoint this, but my guess currently is that pcntl signals may not work well with FPM and its worker pool. Maybe we first set a signal handler on the process, and it's not correctly cleaned up by FPM on the next request cycle?

I tried cleaning up manually the handlers, no changes. I tried putting the cleanup in a `__destruct()` to be sure it's executed, no changes. I'm not sure what to do anymore…

How to reproduce:

- `docker-compose up`
- open http://localhost:8000/?sleep=1
- play with `demo/http.php` in the `sleep` block (e.g. catch the exception)

E.g. in my case:

```php
if (isset($_GET['sleep'])) {
    echo 'Sleeping';
    try {
        sleep(6);
    } catch (Throwable $e) {
        echo 'CAUGHT!';
    }
    echo 'Sleeping';
    sleep(2);
}
```

- first request: I see `CAUGHT!`
- next requests: the catch is not executed, yet the `sleep()` calls are interrupted (without errors/warnings)

mnapoli:
Since this wasn't working with FPM I went with a different route: #1133 

This PR might still be worth following for the Function layer, so I'll leave it open.

PR #897, opened 2021-03-25
Dont use guzzlehttp/guzzle

Nyholm:
We download `guzzlehttp/guzzle` because we need a PSR7 implementation. We already have a PSR7 implementation and a HTTP client installed. We dont need to install a legacy version of Guzzle. 

-----

I know it looks like Im super bias and just want to push to promote `nyholm/psr7`. But I am also maintainer of `guzzlehttp/guzzle` and `guzzlehttp/psr7`.


Nyholm:
Oh, Sorry, I didnt how much we were using it in the tests.

Issue #899, opened 2021-03-27
Calendar extension


semsphy:
Hello @mnapoli,

Thank you very much this great project. I would like to request to enable the `calendar extension` by default.
If you don't mind I will create a PR to this request because it just adds `--enable-calendar` in PHP build.

Thanks,

PR #900, opened 2021-03-27
Enabled Calendar

semsphy:
Issue #899 

Enabled calendar by default


GrahamCampbell:
Shouldn't this be done as an extension layer in https://github.com/brefphp/extra-php-extensions.

semsphy:
> Shouldn't this be done as an extension layer in https://github.com/brefphp/extra-php-extensions.

Ok, I'm checking now. Thank @GrahamCampbell .

PR #903, opened 2021-03-31
Send console output to CloudWatch in real time

tomcant:
We recently noticed a problem whereby output from our console function (stdout) wasn't appearing in CloudWatch. It turned out to be related to the process reaching our configured Lambda timeout.

Consider the following example:

```php
<?php

echo "Line 1\n";

sleep(10);

echo "Line 2\n";
```

For a function that uses the console layer with a timeout of 10 seconds, the above code produces the following output in CloudWatch (notice no echo):

Screenshot 2021-03-31 at 15 48 57

With a 15 second timeout we see the following (containing the echo's):

Screenshot 2021-03-31 at 15 47 49

This PR changes the way output is handled so it is sent to CloudWatch in real time. With the same example script above, and timeouts of 10 and 15 seconds respectively, we now produce the following outputs in CloudWatch (notice the timestamps, which show the 10 second sleep between each echo):

Screenshot 2021-03-31 at 15 43 24

Screenshot 2021-03-31 at 15 44 00

### DIFF
--- a/runtime/layers/console/bootstrap
+++ b/runtime/layers/console/bootstrap
@@ -3,6 +3,7 @@
 
 use Bref\Context\Context;
 use Bref\Runtime\LambdaRuntime;
+use Symfony\Component\Process\Process;
 
 ini_set('display_errors', '1');
 error_reporting(E_ALL);
@@ -43,20 +44,14 @@ while (true) {
             $cliOptions = '';
         }
 
-        // We redirect stderr to stdout so that all the output is collected
-        $command = sprintf(
-            'LAMBDA_INVOCATION_CONTEXT=%s /opt/bin/php %s %s 2>&1',
-            escapeshellarg(json_encode($context)),
-            $handlerFile,
-            $cliOptions
-        );
+        $command = sprintf('/opt/bin/php %s %s 2>&1', $handlerFile, $cliOptions);
+        $process = Process::fromShellCommandline($command, null, ['LAMBDA_INVOCATION_CONTEXT' => json_encode($context)]);
 
-        exec($command, $output, $exitCode);
+        $process->run(function ($type, $buffer) {
+            echo $buffer;
+        });
 
-        $output = implode("\n", $output);
-
-        // Echo the output so that it is written to CloudWatch logs
-        echo $output;
+        $exitCode = $process->getExitCode();
 
         if ($exitCode > 0) {
             throw new Exception('The command exited with a non-zero status code: ' . $exitCode);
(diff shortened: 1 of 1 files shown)


runtime/layers/console/bootstrap
```diff
@@ -51,12 +51,26 @@ while (true) {
...
+
+        do {
+            $status = proc_get_status($process);
 
-        $output = implode("\n", $output);
+            while (false !== $data = fgets($pipes[1])) {
+                echo $data;
+                $output .= $data;
+            }
+        } while ($status['running']);
```
mnapoli, line 70: Thanks for the PR!

This will run the process in a loop, which will consume a lot of CPU resources, is that correct?

tomcant: Thanks for the speedy response.

Good point, as it stands this would just loop and consume resources.

There may be something we can do with `stream_select()` to wait a short time for changes on stdout before going round the loop again (potentially based on `deadlineMs`). Or alternatively we may be able to cause `fgets()` to block with `socket_set_blocking()`, which would mean we don't loop until there is data to read (or the process ends, need to experiment).

I can experiment with these ideas and come back to you.

mnapoli: Thanks for the details!

Maybe it could be worth looking into the Symfony Process component? I believe it has the ability to do that already, and we already require this package in composer.json.

tomcant: Great idea! It looks like the Process component makes this a lot simpler. Would be good to get your thoughts on this. Cheers

mnapoli:
Thank you!

dhrrgn:
@mnapoli @tomcant Unless I am missing something, this introduced a regression/bug. Symfony Process has a default timeout of 60 seconds. So, even if the Lambda timeout is longer the php process will timeout. 

A simple fix would be to just send `null` as the last param to disable it completely on [lime 48](https://github.com/brefphp/bref/blob/154681413f65de17124cf14fa3c12b2bf868f6f5/runtime/layers/console/bootstrap#L48):

```php
$process = Process::fromShellCommandline($command, null, ['LAMBDA_INVOCATION_CONTEXT' => json_encode($context)], null);
```

Alternatively, since we have the `Context`, we could do a little math and set the timeout to the lambda timeout minus 1 second (so it handles timeouts nicely instead of AWS just killing the invocation).

tomcant:
Good spot @dhrrgn, thanks! I think your suggestion of timing out 1s before Lambda makes sense. I've put together a PR to show the problem and a potential solution in https://github.com/brefphp/bref/pull/919. Would be great to know what people think. Thanks again.

mnapoli:
Thanks, I merged your PR and released [1.2.5](https://github.com/brefphp/bref/releases/tag/1.2.5).

PR #915, opened 2021-04-17
[Docs] Update AWS CLI S3 sync command to v2

DvDty:
Just a simple change so the documentation points to the updated aws cli v2 docs.


mnapoli:
Thank you!

PR #916, opened 2021-04-17
[Docs] Update Laravel documentation references to automatically point to latest version

DvDty:
Currently the references are pointing to multiple versions (5.8/7/8) which may cause confusion.
Additionally those references won't need to get updated in the future.


mnapoli:
🙏  thank you so much, love this change!

PR #919, opened 2021-05-02
Fix regression introduced in #903 relating to process timeouts

tomcant:
In #903 we started using the Symfony Process component to execute the handler in the console context. This fixed an issue whereby output from the sub-process wouldn't get flushed to CloudWatch if the Lambda timeout was exceeded. However, as [pointed out](https://github.com/brefphp/bref/pull/903#issuecomment-830712762) by @dhrrgn, this introduced another problem whereby if the sub-process doesn't finish before the default timeout in Symfony Process (60s), it would be terminated, even if the Lambda timeout hasn't been reached.

As suggested in the linked comment above, this change updates the way we invoke the Process component to specify a timeout of 1 second less than the Lambda timeout. Aligning the timeouts in this way means the sub-process will always either complete gracefully (i.e. before the timeout) or the Process component will terminate it and we'll see a thrown exception, rather than the Lambda timing out and terminating for us.

To demo the problem, here's the CloudWatch output for a Lambda with a 65s timeout and a PHP handler that sleeps for 61s:

Screenshot 2021-05-02 at 09 40 03

As we can see, even though the Lambda timeout is greater than the sleep, the Process component terminates at 60s.

With the fix in place, here's the output for the exact same scenario:

Screenshot 2021-05-02 at 09 20 10

Here's the output for a Lambda with a 10s timeout and an 11s sleep (exception thrown 1s before the Lambda timeout):

Screenshot 2021-05-02 at 09 26 47

And just for completeness, this is from a Lambda with a 10s timeout and a 5s sleep:

Screenshot 2021-05-02 at 09 32 33

### DIFF
--- a/runtime/layers/console/bootstrap
+++ b/runtime/layers/console/bootstrap
@@ -44,8 +44,9 @@ while (true) {
             $cliOptions = '';
         }
 
+        $timeout = max(1, $context->getRemainingTimeInMillis() / 1000 - 1);
         $command = sprintf('/opt/bin/php %s %s 2>&1', $handlerFile, $cliOptions);
-        $process = Process::fromShellCommandline($command, null, ['LAMBDA_INVOCATION_CONTEXT' => json_encode($context)]);
+        $process = Process::fromShellCommandline($command, null, ['LAMBDA_INVOCATION_CONTEXT' => json_encode($context)], null, $timeout);
 
         $process->run(function ($type, $buffer) {
             echo $buffer;


runtime/layers/console/bootstrap
```diff
@@ -44,8 +44,9 @@ while (true) {
             $cliOptions = '';
         }
 
+        $timeout = $context->getRemainingTimeInMillis() / 1000 - 1;
```
dhrrgn, line 47: We may want to ensure this is always > 0, incase the timeout is 1 second, this will end up setting the timeout to zero.
```suggestion
        // Ensure the timeout is at least 1 second
        $timeout = max(1, $context->getRemainingTimeInMillis() / 1000 - 1);
```

tomcant: Good point, or even worse, the timeout passed to Symfony would be negative and it looks like the component throws in that case. Good catch 👍🏻

mnapoli:
Thank you for the quick fix!

PR #923, opened 2021-05-04
Update Symfony documentation

ker0x:
Due to recent changes in the Symfony Bridge package and my personal experience using Bref with Symfony, I have updated the documentation accordingly.

All feedbacks and improvments are welcome 😃!

When this PR will be reviewed and approved, a new released of the the `symfony-bridge` should be made because currently the bridge don't require bref to be install.

### DIFF
--- a/docs/frameworks/symfony.md
+++ b/docs/frameworks/symfony.md
@@ -6,90 +6,112 @@ introduction: Learn how to deploy serverless Symfony applications on AWS Lambda 
 
 This guide helps you run Symfony applications on AWS Lambda using Bref. These instructions are kept up-to-date to be compatible with the latest Symfony version.
 
-A demo application is available on GitHub at [github.com/mnapoli/bref-symfony-demo](https://github.com/mnapoli/bref-symfony-demo).
+Multiple demo applications are available on GitHub at [github.com/brefphp/examples/Symfony](https://github.com/brefphp/examples/tree/master/Symfony).
 
 ## Setup
 
-Assuming we're inside an existing Symfony project, let's install Bref via Composer:
+First, make sure you have followed the [Installation guide](../installation.md) to create an AWS account and install the necessary tools.
+
+Next, in an existing Symfony project, install Bref and the [Symfony Bridge package](https://github.com/brefphp/symfony-bridge).
 
 ```
-composer require bref/bref
+composer require bref/symfony-bridge
 ```
 
-Now we can create a `serverless.yml` configuration file (at the root of the project) optimized for Symfony:
+If you are using [Symfony Flex](https://flex.symfony.com/), you're done. Flex will automatically run
+the [bref/symfony-bridge recipe](https://github.com/symfony/recipes-contrib/tree/master/bref/symfony-bridge/0.1) which will perform the following tasks:
+
+- Create a `serverless.yml` configuration file optimized for Symfony.
+- Add the `.serverless` folder to the `.gitignore` file.
+
+Otherwise, you can create the `serverless.yml` file manually at the root of the project, with the following configuration:
 
 ```yaml
-service: bref-demo-symfony
+# serverless.yml
+
+service: symfony
 
 provider:
-    name: aws
-    region: us-east-1
-    runtime: provided.al2
-    environment:
-        # Symfony environment variables
-        APP_ENV: prod
+  name: aws
+  # The AWS region in which to deploy (us-east-1 is the default)
+  region: us-east-1
+  # The stage of the application, e.g. dev, production, staging… ('dev' is the default)
+  stage: dev
+  runtime: provided.al2
+  environment:
+    # Symfony environment variables
+    APP_ENV: prod
 
 plugins:
-    - ./vendor/bref/bref
-
-package:
-    exclude:
-        - node_modules/**
-        - tests/**
-        - var/**
-    include:
-        - var/cache/prod/** # allows to deploy a pre-warmed container
+  - ./vendor/bref/bref
 
 functions:
-    website:
-        handler: public/index.php
-        timeout: 28 # in seconds (API Gateway has a timeout of 29 seconds)
-        layers:
-            - ${bref:layer.php-74-fpm}
-        events:
-            - httpApi: '*'
-    console:
-        handler: bin/console
-        timeout: 120 # in seconds
-        layers:
-            - ${bref:layer.php-74} # PHP
-            - ${bref:layer.console} # The "console" layer
+  # This function runs the Symfony website/API
+  web:
+    handler: public/index.php
+    timeout: 28 # in seconds (API Gateway has a timeout of 29 seconds)
+    layers:
+      - ${bref:layer.php-80-fpm}
+    events:
+      - httpApi: '*'
+  # This function let us run console commands in Lambda
+  console:
+    handler: bin/console
+    timeout: 120 # in seconds
+    layers:
+      - ${bref:layer.php-80} # PHP
+      - ${bref:layer.console} # The "console" layer
+
+package:
+  patterns:
+    # Excluded files and folders for deployment
+    - '!node_modules/**'
+    - '!tests/**'
+    - '!var/**'
+    # If you want to include files and folders that are part of excluded folders,
+    # add them at the end
+    - 'var/cache/prod/**'
 ```
 
-We still have a few modifications to do on the application to make it compatible with AWS Lambda.
+You still have a few modifications to do on the application to make it compatible with AWS Lambda.
 
-Since [the filesystem is readonly](/docs/environment/storage.md) except for `/tmp` we need to customize where the cache and the logs are stored in the `src/Kernel.php` file. This is done by adding 2 new methods to the `Kernel` class:
+Since [the filesystem is readonly](/docs/environment/storage.md) except for `/tmp` you need to customize where the cache and logs are stored in
+the `src/Kernel.php` file. This is done by extending the default `Kernel` class create by Symfony with the `Kernel` class provide by the bridge:
 
-```php
-    public function getLogDir()
-    {
-        // When on the lambda only /tmp is writeable
-        if (isset($_SERVER['LAMBDA_TASK_ROOT'])) {
-            return '/tmp/log/';
-        }
+```diff
+// src/Kernel.php
 
-        return parent::getLogDir();
-    }
+namespace App;
 
-    public function getCacheDir()
-    {
-        // When on the lambda only /tmp is writeable
-        if (isset($_SERVER['LAMBDA_TASK_ROOT'])) {
-            return '/tmp/cache/'.$this->environment;
-        }
++ use Bref\SymfonyBridge\BrefKernel;
+use Symfony\Bundle\FrameworkBundle\Kernel\MicroKernelTrait;
+use Symfony\Component\Config\Loader\LoaderInterface;
+use Symfony\Component\Config\Resource\FileResource;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+-use Symfony\Component\HttpKernel\Kernel as BaseKernel;
+use Symfony\Component\Routing\RouteCollectionBuilder;
 
-        return parent::getCacheDir();
-    }
+- class Kernel extends BaseKernel
++ class Kernel extends BrefKernel
+{
+    // ...
 ```
 
-We can also separate the [build directory](https://symfony.com/doc/current/reference/configuration/kernel.html#build-directory)
+You can also separate the [build directory](https://symfony.com/doc/current/reference/configuration/kernel.html#build-directory)
 from the cache directory to be able to [deploy a pre-warmed cache](#Deploy) and thus improve load time.
 
 ```php
+// src/Kernel.php
+
+class Kernel extends BrefKernel
+{
+    // ...
+    
     public function getBuildDir(): string
     {
         return $this->getProjectDir().'/var/cache/'.$this->environment;
     }
+}
 ```
 
 ## Deploy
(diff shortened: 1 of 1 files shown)


docs/frameworks/symfony.md
```diff
@@ -6,90 +6,112 @@ introduction: Learn how to deploy serverless Symfony applications on AWS Lambda
...
 ```
 
-Now we can create a `serverless.yml` configuration file (at the root of the project) optimized for Symfony:
+If you are using [Symfony Flex](https://flex.symfony.com/), you're done. Flex will automatically run
+the [bref/symfony-bridge recipe](https://github.com/symfony/recipes-contrib/tree/master/bref/symfony-bridge/0.1) which will perform the following tasks:
+
+- Create a `serverless.yml` configuration file optimized for Symfony.
+- Add the `.serverless` folder to the `.gitignore` file.
+
+Otherwise, you can create the `serverless.yml` file manually at the root of the project, with the following configuration:
```
t-richard, line 27: Just an idea : Maybe we could link to the recipes repositiory for those not using Flex, that would avoid synching the recipe and the docs every time we need to do a change ?

Most projects will use Flex as choosing Bref is rarely something you do on a legacy/not updated project.

WDYT ?

docs/frameworks/symfony.md
```diff
@@ -6,90 +6,112 @@ introduction: Learn how to deploy serverless Symfony applications on AWS Lambda
...
+    # add them at the end
+    - 'var/cache/prod/**'
 ```
 
-We still have a few modifications to do on the application to make it compatible with AWS Lambda.
+You still have a few modifications to do on the application to make it compatible with AWS Lambda.
 
-Since [the filesystem is readonly](/docs/environment/storage.md) except for `/tmp` we need to customize where the cache and the logs are stored in the `src/Kernel.php` file. This is done by adding 2 new methods to the `Kernel` class:
+Since [the filesystem is readonly](/docs/environment/storage.md) except for `/tmp` you need to customize where the cache and logs are stored in
+the `src/Kernel.php` file. This is done by extending the default `Kernel` class create by Symfony with the `Kernel` class provide by the bridge:
```
t-richard, line 79:
```suggestion
Since [the filesystem is readonly](/docs/environment/storage.md) except for `/tmp` we need to customize where the cache and logs are stored 
in the `src/Kernel.php` file. This is done automatically by the bridge, you just need to use the `BrefKernel` class provided by the bridge instead of the default `BaseKernel`
```

docs/frameworks/symfony.md
```diff
@@ -6,90 +6,112 @@ introduction: Learn how to deploy serverless Symfony applications on AWS Lambda
...
-        return parent::getCacheDir();
-    }
+- class Kernel extends BaseKernel
++ class Kernel extends BrefKernel
+{
+    // ...
 ```
 
-We can also separate the [build directory](https://symfony.com/doc/current/reference/configuration/kernel.html#build-directory)
+You can also separate the [build directory](https://symfony.com/doc/current/reference/configuration/kernel.html#build-directory)
```
t-richard, line 100: I've tested this several times and this works really bad with the bridge for now (see for example https://github.com/brefphp/symfony-bridge/pull/36 and https://github.com/symfony/symfony/issues/40827)

I would delete this for now as it causes more problems than it solves and the bridge is more performant on projects where I tested both approaches.

docs/frameworks/symfony.md
```diff
@@ -6,90 +6,112 @@ introduction: Learn how to deploy serverless Symfony applications on AWS Lambda
...
 from the cache directory to be able to [deploy a pre-warmed cache](#Deploy) and thus improve load time.
 
 ```php
+// src/Kernel.php
+
+class Kernel extends BrefKernel
+{
+    // ...
+    
     public function getBuildDir(): string
```
t-richard, line 110: and remove this too

docs/frameworks/symfony.md
```diff
@@ -105,47 +127,39 @@ The application is now ready to be deployed. Follow [the deployment guide](/docs
...
-As you may have noticed, we define a function of type "console" in `serverless.yml`. That function is using the [Console runtime](/docs/runtimes/console.md), which lets us run the Symfony Console on AWS Lambda.
+As you may have noticed, we define a function of type "console" in `serverless.yml`. That function is using the [Console runtime](/docs/runtimes/console.md),
+which lets us run the Symfony Console on AWS Lambda.
 
 To use it follow [the "Console" guide](/docs/runtimes/console.md).
 
 ## Logs
 
-While overriding the log's location in the `Kernel` class was necessary for Symfony to run correctly, by default Symfony logs in `stderr`. That is great because Bref [automatically forwards `stderr` to AWS CloudWatch](/docs/environment/logs.md).
+While overriding the log's location in the `Kernel` class was necessary for Symfony to run correctly, by default Symfony logs in `stderr`. That is great because
```
t-richard, line 137: We don't need to say this as it is internal to the Bridge

```suggestion
By default Symfony logs in `stderr`. That is great because
```

t-richard (changes requested):
@ker0x thank you very much for this.

TBH, I was planning to do it this week :smile: 

I think the bridge is now stable enough to make this move and encourage new comers to use it 

@mnapoli it would be nice to tag a new version of the bridge to get the latest fixes from https://github.com/brefphp/symfony-bridge/pull/37, https://github.com/brefphp/symfony-bridge/pull/39 and https://github.com/brefphp/symfony-bridge/pull/41  without having to require `dev-master` :slightly_smiling_face: 

Here is a first round of review !

docs/frameworks/symfony.md
ker0x: Indeed, this will be easier to maintain afterwards.

docs/frameworks/symfony.md
ker0x: Fixed

docs/frameworks/symfony.md
ker0x: Also fixed

docs/frameworks/symfony.md
ker0x: Indeed, I changed it

docs/frameworks/symfony.md
ker0x: I removed the configuration example and added an annotation to recipe with a direct link

docs/frameworks/symfony.md
```diff
@@ -202,35 +157,48 @@ framework:
...
+```yaml
+# serverless.yml
 
-```html
-<script src="{{ asset('js/app.js') }}"></script>
+package:
+  pattern:
+    - '!assets/**'
+    - '!public/build/**'
+    - 'public/build/manifest.json'
```
t-richard, line 191: maybe add `!public/bundles` that may be generated by `bin/console assets:install`

docs/frameworks/symfony.md
```diff
@@ -202,35 +157,48 @@ framework:
...
 ```
 
-If your templates reference some assets via direct path, you should edit them to use the `asset()` helper:
+### Assets in templates
+
+For the above configuration to work, assets must be referenced in templates via the `asset()` helper:
 
-```html
+```diff
 - <img src="/images/logo.png">
```
t-richard, line 200: Shouldn't it be the following ?

```suggestion
- <img src="/build/images/logo.png">
```

docs/frameworks/symfony.md
```diff
@@ -6,90 +6,49 @@ introduction: Learn how to deploy serverless Symfony applications on AWS Lambda
...
+If you are using [Symfony Flex](https://flex.symfony.com/), you're done. Flex will automatically run
+the [bref/symfony-bridge recipe](https://github.com/symfony/recipes-contrib/tree/master/bref/symfony-bridge/0.1) which will perform the following tasks:
 
-We still have a few modifications to do on the application to make it compatible with AWS Lambda.
+- Create a `serverless.yml` configuration file optimized for Symfony.
+- Add the `.serverless` folder to the `.gitignore` file.
 
-Since [the filesystem is readonly](/docs/environment/storage.md) except for `/tmp` we need to customize where the cache and the logs are stored in the `src/Kernel.php` file. This is done by adding 2 new methods to the `Kernel` class:
+> Otherwise, you can create the `serverless.yml` file manually at the root of the project. Take a look
+at the [default configuration](https://github.com/symfony/recipes-contrib/blob/master/bref/symfony-bridge/0.1/serverless.yaml) provide by the recipe.
```
t-richard, line 28:
```suggestion
at the [default configuration](https://github.com/symfony/recipes-contrib/blob/master/bref/symfony-bridge/0.1/serverless.yaml) provided by the recipe.
```

t-richard (changes requested):
Here are some other small things I noticed :slightly_smiling_face:

docs/frameworks/symfony.md
ker0x: After thought, I think we can totally remove this part because Symfony recommends to use the Asset component to display images anyway: https://symfony.com/doc/current/templates.html#linking-to-css-javascript-and-image-assets

t-richard: Well there are two ways to see this :
1. this is common knowledge that you should use `asset()` so no need to document it
2. some users might not know this and this small bit of docs will help them and avoid support from the Bref community for these users 

From my personal experience, there are some users that don't know about this and just use absolute paths in their templates.

So I would keep it.

It's also the same in the laravel docs https://bref.sh/docs/frameworks/laravel.html#assets-in-templates

But maybe linking to that Symfony doc for details about this could be a good idea.

ker0x: It makes sense and remains consistent with the Laravel doc indeed, PR updated 😉

docs/frameworks/symfony.md
```diff
@@ -6,146 +6,95 @@ introduction: Learn how to deploy serverless Symfony applications on AWS Lambda
...
 ```
-composer require bref/bref
+composer require bref/symfony-bridge
 ```
 
-Now we can create a `serverless.yml` configuration file (at the root of the project) optimized for Symfony:
-
-```yaml
-service: bref-demo-symfony
+If you are using [Symfony Flex](https://flex.symfony.com/),it will automatically run
```
t-richard, line 21: missing space :smile: 

```suggestion
If you are using [Symfony Flex](https://flex.symfony.com/), it will automatically run
```

t-richard (approved):
Just a minor comment but all seem good to me :heavy_check_mark: 

Thanks @ker0x

t-richard:
@mnapoli @deleugpn is there something preventing the merge of this PR ? Except a new release in bref/symfony-bridge ?

Let me know if I can be of any help in moving this forward (cc @ker0x)

deleugpn:
My knowledge in Symfony is completely `null` and since this PR is a documentation-only, I can't be much of help. It would be nice if more Symfony users could weight in on these changes.

docs/frameworks/symfony.md
```diff
@@ -202,33 +156,47 @@ framework:
...
+Finally, you need to update the `serverless.yml` file to exclude the `assets`, `public/build` and `public/bundles` directories from deployment:
 
-For the above configuration to work, assets must be referenced in templates via the `asset()` helper:
+```yaml
+# serverless.yml
 
-```html
-<script src="{{ asset('js/app.js') }}"></script>
+package:
+  pattern:
```
t-richard, line 187: it is patterns with an `s`

t-richard:
@deleugpn I understand that you are not knowledgable with Symfony.

I'll try to ask on Slack if Symfony users could check this PR and review/approve it.

docs/frameworks/symfony.md
ker0x: it's fixed, thanks @t-richard

docs/frameworks/symfony.md
```diff
@@ -154,46 +103,51 @@ To learn more about all this, read the [environment variables documentation](/do
...
 ```yaml
+# config/packages/framework.yaml
+
 framework:
-    # trust the remote address because API Gateway has no fixed IP or CIDR range that we can target
-    trusted_proxies: '127.0.0.1,REMOTE_ADDR'
-    # trust "X-Forwarded-*" headers coming from API Gateway
-    trusted_headers: ['x-forwarded-for', 'x-forwarded-proto', 'x-forwarded-port']
+  # trust the remote address because API Gateway has no fixed IP or CIDR range that we can target
+  trusted_proxies: '127.0.0.1,REMOTE_ADDR'
```
Nyholm, line 119: Isn't `$_SERVER['REMOTE_ADDR']` always 127.0.0.1 on Lambda?

docs/frameworks/symfony.md
```diff
@@ -154,46 +103,51 @@ To learn more about all this, read the [environment variables documentation](/do
...
 You can get more details in the [Symfony documentation](https://symfony.com/doc/current/deployment/proxies.html).
 
 ### Getting the user IP
 
-**When using CloudFront** on top of API Gateway, you will not be able to retrieve the client IP address and you will
-instead get one of Cloudfront's IP when calling `Request::getClientIp()`. If you really need this, you will need
-to whitelist [every CloudFront IP](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/LocationsOfEdgeServers.html)
+**When using CloudFront** on top of API Gateway, you will not be able to retrieve the client IP address, and you will instead get one of Cloudfront's IP when
+calling `Request::getClientIp()`. If you really need this, you will need to
+whitelist [every CloudFront IP](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/LocationsOfEdgeServers.html)
```
Nyholm, line 136: I know you did not change this. But this is not really true. 

You can just update your index.php file with: 

```php
// Get user IP:
if (isset($_SERVER['LAMBDA_REQUEST_CONTEXT'])) {
    $context = json_decode($_SERVER['LAMBDA_REQUEST_CONTEXT'], true);
    $_SERVER['HTTP_X_FORWARDED_FOR'] = $context['identity']['sourceIp'] ?? '';
}
```

docs/frameworks/symfony.md
```diff
@@ -202,33 +156,47 @@ framework:
...
-Finally, you can compile assets for production in the `public` directory, then synchronize that directory to a S3 bucket:
+Then, you can compile assets for production in the `public` directory, and synchronize that directory to a S3 bucket:
 
 ```bash
 php bin/console assets:install --env prod
 # if using Webpack Encore, additionally run
 yarn encore production
-aws s3 sync public/ s3://<bucket-name>/ --delete --exclude index.php
+aws s3 sync public/ s3://<bucket-name>/ \
+  --delete \
```
Nyholm, line 175: If you delete stuff, it would mean that there is some downtime for your application. (ie no assets)

docs/frameworks/symfony.md
```diff
@@ -202,33 +156,47 @@ framework:
...
+  --exclude index.php \
+  --exclude public/build/manifest.json \
+  --exclude public/build/entrypoint.json
```
Nyholm, line 178: Also, one should configure cace control (But only if one use unique filenames, ie hashes)

```
--cache-control max-age=31449600,immutable
```

docs/frameworks/symfony.md
```diff
@@ -154,46 +103,51 @@ To learn more about all this, read the [environment variables documentation](/do
...
+First, you need to [tell Symfony](https://symfony.com/doc/current/reference/configuration/framework.html#base-urls) to use the S3 URL as the assets base URL,
+instead of your app domain in production.
 
 ```yaml
-# config/packages/prod/framework.yaml
+# config/packages/prod/assets.yaml
+
 framework:
   assets:
     base_urls: 'https://<bucket-name>.s3.amazonaws.com'
```
Nyholm, line 153: Should we really care about S3? Why not do cloudfront directly?

Nyholm (commented):
Thank you. 

I've added some thoughts. Feel free to skip, ignore or add to a follow-up PR.

docs/frameworks/symfony.md
t-richard: I think it is https://github.com/brefphp/bref/blob/master/src/Event/Http/FpmHandler.php#L184

So we could change it to only keep the IP

```suggestion
  trusted_proxies: '127.0.0.1'
```

docs/frameworks/symfony.md
t-richard: There was some discussions about this on Slack and it seems to work fine but we weren't able to say if this is working in all scenarios. And I wasn't able to find AWS docs that states clearly what they put in it.

If you are sure about that, I'm a :100: to document it :+1:

docs/frameworks/symfony.md
Nyholm: :+1:

docs/frameworks/symfony.md
t-richard: True, I think the solution is to sync twice. First without `--delete` and then with the `--delete` option.

Or do you have another solution ?

docs/frameworks/symfony.md
t-richard: File hasing is enabled by default with Webpack Encore. But I'm not sure it's true for third party bundles providing assets (like EasyAdmin or Sonata for example) moved by `bin/console assets:install` in `bundles`.

docs/frameworks/symfony.md
Nyholm: I'll just leave the old files :)

docs/frameworks/symfony.md
t-richard: This is simpler to setup for beginners and if you put Cloudfront in front of your app, then this is not needed I think.

t-richard:
@Nyholm thank you very much appreciated. 

I think we can address the trusted proxies IP here, and I'll open a follow-up PR for assets and the source IP thing because they could need more work to get them right. 

This would allow us to move this PR forward and merge it.

mnapoli:
Thank you all!

t-richard:
Thank you @mnapoli and @Nyholm for your time on that matter.

And thank you @ker0x for this PR, the contributions on the recipe and the bridge :heart: 

We'll be able to build upon that. I'll hopefully come up with some new PRs soon.

PR #926, opened 2021-05-05
Fix #925 to handle excluding vendor dir in package patterns with sepa…

### LINKED ISSUE #925: Handle patterns instead of exclude in serverless.yml when using separate vendor option
**Description:**

When using the separate vendor option, the vendor directory is added to the package.exclude array in the serverless.yml, if it's not there yet. package.exclude is deprecated and it is now recommended to use package.patterns. Not defining the package.exclude key will result in an error.

This bug can easily be fixed, I will open a pull request shortly.

**How to reproduce:**

```
custom:
    bref:
        separateVendor: true

package:
    patterns:
        - '!node_modules/**'
```

svnldwg:
…rateVendor option

### DIFF
--- a/index.js
+++ b/index.js
@@ -103,8 +103,12 @@ class ServerlessPlugin {
         // we will add it here as we do not want the vendor folder in our
         // lambda archive file.
         let excludes = this.serverless.service.package.exclude;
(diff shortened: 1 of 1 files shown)


mnapoli:
Thanks!

Issue #875, opened 2021-03-04
Symlink to bref/lib64/libgpg-error.so broken in some layers


logicalor:
**Description:**

Development Environment is Ubuntu 20.04, Serverless 2.28

Sample layers:

- arn:aws:lambda:ap-southeast-2:209497400698:layer:php-74:17
- arn:aws:lambda:ap-southeast-2:209497400698:layer:php-73:31

**How to reproduce:**

Create a serverless config referring to these layers:

e.g.

```
  layers:
    - ${bref:layer.php-73}
```
Attempt to locally invoke serverless.

Serverless will fail to execute with the error

`Error: ENOENT: no such file or directory, stat '/home/{username}/.cache/serverless/invokeLocal/layers/php-73/31/bref/lib64/libgpg-error.so'`

Upon examining `/home/{username}/.cache/serverless/invokeLocal/layers/php-73/31/bref/lib64/` (eg. `ls -la`) one will see the symlink is broken:  `libgpg-error.so -> ../../lib64/libgpg-error.so.0.10.0`

Manually download and examine the layer in question - eg. .`arn:aws:lambda:ap-southeast-2:209497400698:layer:php-73:31` and examine the contents.  One will see that the symlink is broken in the layer package as well.

**Solution**

Fix the symlink in the layers - change from `../../lib64/libgpg-error.so.0.10.0` to `../lib64/libgpg-error.so.0.10.0` or `libgpg-error.so.0.10.0`

mnapoli:
Hi! I've since published a new release with new layers, could you update Bref, test and let us know if you still have the problem?

logicalor:
Hi again @mnapoli 

Sorry about the delayed response, and thanks for the update.  I tried again against the layer `arn:aws:lambda:ap-southeast-2:209497400698:layer:php-73:32`.

In this layer it appears the symlink doesn't exist at all.

`ls` on `32/bref/lib64` gives me

```
drwxrwxr-x 2 shaunj shaunj   4096 Apr 22 11:29 .
drwxrwxr-x 5 shaunj shaunj   4096 Apr 22 11:29 ..
lrwxrwxrwx 1 shaunj shaunj     22 Apr 22 11:29 libgpg-error.so.0 -> libgpg-error.so.0.10.0
-rwxr-xr-x 1 shaunj shaunj  19296 Apr 22 11:29 libgpg-error.so.0.10.0
[...]

PR #928, opened 2021-05-06
Fix missing libs

logicalor:
These modifications are intended to address issue #875  whereby compiled PHP layers are missing libraries, despite successful compilation and deployment.


mnapoli:
Thank you for the details!

To be honest, as I said in #875 since the problem only occurs when using `serverless local invoke`, I wouldn't take the risk of changing the build for live environments without understanding more the issue.

Instead I recommend using `vendor/bin/bref local` for local development.

Thanks for taking the time though.

Issue #930, opened 2021-05-10
Running locally using HttpRequestEvent fails with TypeError


sc0ttdav3y:
**Description:**

When running locally with a `HttpRequestEvent` handler, Bref fails with a TypeError because `$event` is null but the method parameter type expects an array.

```
[ERROR] Argument 1 passed to Bref\Event\Http\HttpRequestEvent::__construct() must be of the type array, null given, called in /vendor/bref/bref/src/Event/Http/HttpHandler.php on line 23   
```

Relevant stack trace:

```
Stack trace:
#0 /vendor/bref/bref/src/Event/Http/HttpHandler.php(23): Bref\Event\Http\HttpRequestEvent->__construct(NULL)
#1 /vendor/bref/bref/src/Runtime/Invoker.php(29): Bref\Event\Http\HttpHandler->handle(NULL, Object(Bref\Context\Context))
#2 /vendor/bref/bref/src/Console/Command/Local.php(74): Bref\Runtime\Invoker->invoke(Object(Bref\Event\Http\Psr15Handler), NULL, Object(Bref\Context\Context))
#3 [internal function]: Bref\Console\Command\Local->__invoke(NULL, NULL, NULL, Object(Example\ApiTestString), NULL, Object(Symfony\Component\Console\Style\SymfonyStyle))
...

```

**How to reproduce:**

1. Implement a HttpRequestHandler 
2. Call via CLI as below

`Example\MyCommand.php`:

```php
<?php

namespace Example;

use Nyholm\Psr7\Response;
use Psr\Http\Message\ResponseInterface;
use Psr\Http\Message\ServerRequestInterface;
use Psr\Http\Server\RequestHandlerInterface;

class MyCommand implements RequestHandlerInterface
{
    final public function handle(ServerRequestInterface $request): ResponseInterface
    {
        return new Response(200, [], json_encode(["message" => "Hi"]));
    }
}
```

`handler.php`:
```php
<?php

require __DIR__ . '/vendor/autoload.php';
return new \Example\MyCommand();

```

Call it via CLI:

```
vendor/bin/bref local --handler=handler.php
```

sc0ttdav3y:
I've submitted a PR that addresses this issue: https://github.com/brefphp/bref/pull/931

mnapoli:
Closing per the PR

PR #931, opened 2021-05-10
Fix: running locally using HttpRequestEvent fails with TypeError

sc0ttdav3y:
This PR fixes the issue described in https://github.com/brefphp/bref/issues/930.

The README docs [here](https://bref.sh/docs/function/local-development.html) describe local development, and provide the following command examples:

```
$ vendor/bin/bref local hello
$ vendor/bin/bref local hello '{"name": "Jane"}'
$ vendor/bin/bref local hello --file=event.json
```

The problem is that these don't work if the handlers use `HttpRequestEvent` (i.e. implementing `RequestHandlerInterface`). Currently Bref crashes with a TypeError because `$event` is null but is expected to be an array in `HtpRequestEvent`'s constructor. 

So to ensure the local development docs are honoured for these events, I've implemented a basic GET request in the case where `$event` is `null`. 

An alternative strategy would be to throw an exception in this place to avoid the TypeError. If this is preferred, some documentation on what's expected for running these types of handlers should be also written (and I'm out of my depth here).

Thanks, Scott.

### DIFF
--- a/src/Event/Http/HttpHandler.php
+++ b/src/Event/Http/HttpHandler.php
@@ -20,6 +20,11 @@ abstract class HttpHandler implements Handler
             return ['Lambda is warm'];
         }
 
+        // $event must be an array for HttpRequestEvent but might come in as null
+        if (!is_array($event)) {
+            $event = [];
+        }
+
         $httpEvent = new HttpRequestEvent($event);
 
         $response = $this->handleRequest($httpEvent, $context);
--- a/src/Event/Http/HttpHandler.php
+++ b/src/Event/Http/HttpHandler.php
@@ -25,7 +25,7 @@ abstract class HttpHandler implements Handler
             $event = [
                 'requestContext' => [
                     'http' => [
-                        'method' => 'GET'
+                        'method' => 'GET',
                     ],
                 ],
             ];
--- a/src/Event/Http/HttpHandler.php
+++ b/src/Event/Http/HttpHandler.php
@@ -20,9 +20,15 @@ abstract class HttpHandler implements Handler
             return ['Lambda is warm'];
         }
 
-        // $event must be an array for HttpRequestEvent but might come in as null
-        if (!is_array($event)) {
-            $event = [];
+        // If $event is null, set it up as a GET default
+        if ($event === null) {
+            $event = [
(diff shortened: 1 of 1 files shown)


sc0ttdav3y:
Hi @mnapoli, can I get some feedback on this PR please?

Following the documentation as a first-time user produces a PHP Fatal Error, so I expect this is a bug, but it looks like you've parked it for a future "enhancement"? Can you let me know your thoughts on this?

Thanks, Scott.

mnapoli:
Hi @sc0ttdav3y, sorry for not following up sooner.

In short, what you are experiencing is expected.

```
$ vendor/bin/bref local hello
$ vendor/bin/bref local hello '{"name": "Jane"}'
$ vendor/bin/bref local hello --file=event.json
```

Non of these invocations are valid: if you are writing an API Gateway application, you must pass an API Gateway event to Lambda. That's how API Gateway + Lambda works.

You can find an example here: https://github.com/brefphp/bref/blob/master/tests/Event/Http/Fixture/ag-v2-simple.json

I have 2 recommendations:

- You have decided to write low-level Lambda functions with the "Bref for Functions" approach:
    - Either you have a good reason for this, but that means you need to master these more advanced scenarios
    - Either you don't really know why, then I suggest you use "[Bref for web apps](https://bref.sh/docs/runtimes/http.html)" instead -> this provides a much more familiar environment, especially if your app is "HTTP-oriented". It's basically "PHP as usual", without anything too specific to Lambda.
- If you decide to continue using "Bref for Functions", maybe https://github.com/brefphp/dev-server can help. But again, this is an advanced scenario and there is limited support and documentation for this

sc0ttdav3y:
Thanks for getting back to me @mnapoli. 

I'm happy with the function approach, it makes sense for my use-case and I have learned more about the event JSON in the meantime. 

I still think bref crashing with a TypeError isn't a good way to leave this, and an exception would be more appropriate. Also some docs. I'm happy to help out, but please let me know if it's worth my time.

Thanks again.

mnapoli:
Yes I do agree that a better error message would be more helpful here, we actually discussed that with @deleugpn recently. If you have to send a PR for that feel free 👍

PR #934, opened 2021-05-16
Fix: bref init on PHP8

t-richard:
Here is an idea to fix https://github.com/brefphp/bref/discussions/933

ATM running `vendor/bin/bref init` with PHP8 will result in a composer errror when deployed on Lambda
because the layer used by default is `74-fpm`

Here I check for the php version and use that to determine which template should be used


template/8/http/index.php
```diff
@@ -0,0 +1 @@
+../../7/http/index.php
```
deleugpn, line 1: Is this a symbolic link? Could it have issues with Windows NTFS filesystem?

t-richard: This is indeed a symbolik and I didn't thought about windows compatibility to be honest.

Reading some things about this online, it appears it could cause issues.

Maybe the simplest solution is to duplicate these files, they are small enough and won't be changed often (as they are mostly meant to be used for demo and introduction to bref).

mnapoli: 👍  let's be safe then and duplicate (no idea about Windows too)

mnapoli (commented):
Thank you for doing this, that's awesome!

t-richard:
Updated to duplicate instead of the symlinks

deleugpn:
I have been thinking about this and although I think it's great to fix the issue, I'm not so sure about the approach. Given that this is a basic initialization, I fail to see the usefulness of duplicating the entire folder e.g. the PHP files will very likely be the same throughout multiple versions because of how basic they are.

I was wondering whether a simple search and replace on the serverless template could be a simpler solution? 

@mnapoli what do you think?

mnapoli:
You know what, I foresee this being a problem with PHP 8.1 indeed 🤔  And currently PHP 7.3 is still supported, so right now we would have the same problem with that version too.

Replacing the version number would be a more scalable approach indeed 👍  Sorry @t-richard I made you duplicate those files.

t-richard:
I agree, I will take a look at this later this week and see what I can come up with :slightly_smiling_face: 

> And currently PHP 7.3 is still supported, so right now we would have the same problem with that version too.

PHP 7.3 would not be a problem because the 7.4 layer would be used and there should be no compatibility issue (nor composer error) when running PHP 7.3 code on PHP 7.4. Same for running 8.0 code on 8.1

template/7/function/serverless.yml
```diff
@@ -17,5 +17,5 @@ functions:
 
 # Exclude files from deployment
 package:
-    exclude:
-        - 'tests/**'
+    pattern:
```
deleugpn, line 20: I was looking into this change and it appears that `pattern` is an invalid key. The correct key would be `patterns`, perhaps?

```
Serverless:   at 'package': unrecognized property 'pattern'
```

t-richard: Yes indeed 🤦
I'll look into this soon™️

Didn't had the time to work on this again and test it

deleugpn:
I just sent https://github.com/brefphp/bref/pull/937 as an alternative solution using search/replace.

PR #936, opened 2021-05-24
Improve error message when invoking Http Event with invalid payload

deleugpn:
#930 reports an issue when using `bref local hello` with an empty payload and an Http lambda. This PR will make the message a bit clearer and consistent with the other invocation events.

Note that PR #931 mentioned 3 invocation styles (e.g. `vendor/bin/bref local hello '{"name": "Jane"}'` and `vendor/bin/bref local hello --file=event.json`) but those don't have the same issue as they already fall under the appropriate error message.

```

 [ERROR] This handler expected to be invoked with a API Gateway or ALB event. Instead, the handler was invoked with
         invalid event data: Array
         (
             [name] => value
         )
```

ping @sc0ttdav3y

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -1,23 +1,17 @@
 <?xml version="1.0" encoding="utf-8"?>
-<phpunit colors="true"
-         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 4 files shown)


mnapoli (approved):
Awesome!

PR #937, opened 2021-05-24
Automatically detect user PHP Version for Lambda Layer definition

deleugpn:
This is a pull request built on top of the proposed changes on #934.

The original issue is that people running PHP 8 are now getting a serverless template that relies on the PHP 7.4 layer. Since layers follow a `MAJORMINOR` pattern, we can identify the PHP user version and replace it on the template.

Replacing seems like a better strategy than duplicating the entire file. This strategy also avoids weird issues in case the user is running e.g. PHP 7.3 and having his/her code deployed on PHP 7.4.

For simplicity, I also removed the "folder" replication with a hardcoded filename since only `index.php` and `serverless.yml` are being copied over. We may revisit this if we need to add flexibility back by copying different set of files.


mnapoli (approved):
Looking good!

PR #938, opened 2021-06-03
Bumped PHP versions

peldax:
Hello!

Here comes the regular PR which bumps the PHP versions to latest. :+1:

### DIFF
--- a/runtime/base/php-73.Dockerfile
+++ b/runtime/base/php-73.Dockerfile
@@ -13,7 +13,7 @@
 
 FROM bref/tmp/step-1/build-environment as build-environment
 
-ENV VERSION_PHP=7.3.28
+ENV VERSION_PHP=7.3.29
 
 ENV PHP_BUILD_DIR=${BUILD_DIR}/php
 RUN set -xe; \
--- a/runtime/base/php-74.Dockerfile
+++ b/runtime/base/php-74.Dockerfile
@@ -36,7 +36,7 @@ RUN set -xe; \
     make install
 
 
-ENV VERSION_PHP=7.4.20
+ENV VERSION_PHP=7.4.21
 
 
 ENV PHP_BUILD_DIR=${BUILD_DIR}/php
--- a/runtime/base/php-74.Dockerfile
+++ b/runtime/base/php-74.Dockerfile
@@ -36,7 +36,7 @@ RUN set -xe; \
     make install
 
 
-ENV VERSION_PHP=7.4.19
+ENV VERSION_PHP=7.4.20
 
 
 ENV PHP_BUILD_DIR=${BUILD_DIR}/php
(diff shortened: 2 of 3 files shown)


deleugpn:
Hi @peldax, thanks for the PR! I just want to give you a heads up that this PR might stay open for a little while. The reason for that is because it takes about 3 hours to compile all PHP versions and publish the new layers. If there isn't any major change on the patch release, sometimes we tend to skip one or two.
Besides that, we'll also going to be bundling multiple PHP versions on a single release instead of having to go through the same process multiple time for PHP 7.3, 7.4 and 8.0 individually.

peldax:
Sure, no problem!

This is currently my third PR which bumps PHP versions and it was always meant to be something more like a notification. I am aware that increasing one number is nothing in comparison to compiling and publishing the layer to amazon :+1:.

peldax:
This PR will also include 7.3 and 8.0, but they haven't been officially released yet - only 7.4 at the moment.

peldax:
I included 8.0, but 7.3 has slower release cycle and probably wont release this month.

shouze:
> I included 8.0, but 7.3 has slower release cycle and probably wont release this month.

BTW for php 7.3 do we have to wait? EOF is for the end of this year... WDYT @mnapoli ? :thinking: 
image

peldax:
Currently there is nothing planned for PHP 7.3 (unless some security fix appears) - as noted by the changelog: https://github.com/php/php-src/blob/PHP-7.3/NEWS

But it might be worth waiting for next PHP 8.0.9 and PHP 7.4.21, which should release soon.

mnapoli:
TBH I don't really have an opinion in the sense that I don't intend to merge this PR immediately (rebuilding layers takes some effort)

peldax:
> TBH I don't really have an opinion in the sense that I don't intend to merge this PR immediately (rebuilding layers takes some effort)

@mnapoli is there possibly anything we can do to automate the process?

mnapoli:
Here's what I do when I do a release:

```sh
# set proper environment variables first

make runtimes

make layers.json

git commit + push
```

The `make runtimes` is the painful part, I have to do it in an EC2 instance that I boot specifically for that.

Maybe we can now use GitHub Actions and their [manual trigger](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow) (`make runtimes` shouldn't run automatically)? I'm not sure if this would run into a timeout (it can take 40 minutes or more IIRC)

If anyone wants to give it a try, here's what you could do:

- fork bref and create a branch
- enable GitHub actions in your repository
- feel free to disable the other github actions jobs for your tests (that may make things simpler)
- try to create a "manual trigger" job that runs `make runtimes` (you'll need to configure your AWS credentials: that will publish layers into your AWS account)

If it works, then feel free to open a pull request.

t-richard:
Note 7.3.29, 7.4.21 and 8.0.8 contain a security fix.

I think merging and releasing this should be considered.

mnapoli:
Good point, merging and releasing 👍

mnapoli:
Released, thank you all!

PR #942, opened 2021-06-04
Added expose_php with “Off“ value to disable the X-Powered-By header

Chtiadrien:
Hello 👋

Just got a pentest from one of my client on a project running with Bref. I got only one "vulnerability" on my list which is the "X-Powered-By" header. 

This PR disable it to avoid showing the PHP version currently in use.

I know I can overwrite it on my side, but I think it's better to overwrite it to enable it rather than to disable it. What do you think?

### DIFF
--- a/runtime/layers/fpm/php.ini
+++ b/runtime/layers/fpm/php.ini
@@ -48,3 +48,7 @@ upload_max_filesize=6M
 
 ; Enable the PDO MySQL extension
 extension=pdo_mysql
+
+; Disable the header "X-Powered-By" exposing the installed PHP version
+; e.g. X-Powered-By: PHP/8.0.0
+expose_php=Off


deleugpn:
I'm going to leave this one for @mnapoli to decide because it's a very subjective thing. 
From my perspective, most projects that will be dealing with pentest are advanced enough to always warrant a custom `php.ini` anyway, I highly doubt that disabling it by default will prevent you from creating a custom `ini` just for this, so it doesn't really improve DX. 
I don't think anyone would go out of their way to enable this, either. If there's a strong security concern to it, perhaps it could be addressed by PHP Internals by shipping it off by default, but reality is different and there will be controversial around whether this is security by obscurity or not related to security at all.

mnapoli:
Hi @Chtiadrien!

To be honest I say let's keep it.

As @deleugpn said: it can be disabled easily for those who want to, it is enabled by default in PHP (so we are not making a decision here). And personally I think that it benefits PHP indirectly (this is used by crawlers to report PHP's market share).

Chtiadrien:
Thank you for your time and your explanations 👍 

I'm closing the PR 😉

PR #957, opened 2021-06-18
Fix typo in Lift construct variable

t-richard:
Same as https://github.com/getlift/lift/pull/37


mnapoli:
Thank you! This is what I get from copy/pasting blindly ^^

PR #963, opened 2021-06-25
Deploying with Docker

deleugpn:
This is a draft to improve Bref's documentation on how to deploy Docker-based PHP applications to AWS Lambda. 

I'm looking for community feedback on the steps described and questions that this document should potentially answer.

### DIFF
--- a/docs/web-apps/docker.md
+++ b/docs/web-apps/docker.md
@@ -1,0 +1,87 @@
+---
+title: Docker
+current_menu: web-docker
+previous:
+    link: docs/web-apps/local-development.html
+    title: Local development for web apps
+---
+
+AWS Lambda support running a Docker Container as a handler
+for your function. This alternative allows for lambda functions
+with up to 10GB to be deployed, as opposed to 250MB for non-docker.
+Another reason to choose this approach is if you need more than 5
+PHP extensions. Installing/Enabling multiple extensions on Docker
+is easier and will avoid hitting the limit of 5 layers per function.
+
+## Docker Image
+
+Bref helps you deploy Docker to AWS Lambda by offering
+out-of-the-box base images that are package for the Lambda environment.
+Here is an example of a Docker image
+
+```Dockerfile
+FROM composer:2 as installation
+
+WORKDIR /app
+
+RUN composer require bref/bref
+
+FROM bref/php-80-fpm
+
+COPY --from=bref/extra-redis-php-80:0.10 /opt/bref-extra /opt/bref-extra
+
+COPY --from=bref/extra-gmp-php-80:0.10 /opt/bref-extra /opt/bref-extra
+COPY --from=bref/extra-gmp-php-80:0.10 /opt/bref/ /opt/bref
+
+COPY --from=installation /app/ /var/task/
+
+RUN echo "<?php echo 'Hello from Docker Bref!';" >> /var/task/public/index.php
+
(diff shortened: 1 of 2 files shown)


t-richard:
Seems like a nice addition. 

Here are some thoughts :
* why put it in the Web Apps docs ? Seems to apply to both event-driven functions and web apps but there isn't a common place for this in the current docs I think.
* I think I read that big images have a huge impact on cold starts. Maybe it's worth mentionning.

i also think it would it would be nice to include a comparison between Bref runtimes and custom Docker image to guide the users toward the right solution for them. IMHO it targets specific use-cases where you need more control over the runtime but should not be the go-to solution for everyone.

docs/web-apps/docker.md
```diff
@@ -0,0 +1,87 @@
...
+COPY --from=bref/extra-redis-php-80:0.10 /opt/bref-extra /opt/bref-extra
+
+COPY --from=bref/extra-gmp-php-80:0.10 /opt/bref-extra /opt/bref-extra
+COPY --from=bref/extra-gmp-php-80:0.10 /opt/bref/ /opt/bref
+
+COPY --from=installation /app/ /var/task/
+
+RUN echo "<?php echo 'Hello from Docker Bref!';" >> /var/task/public/index.php
+
+CMD ["public/index.php"]
```
mnapoli, line 40: Really cool!

I would suggest splitting the example like this:

- first, the docs should start with the simplest example possible -> that example helps reader determine whether they get it or not. I think the 1st example could fit in maybe 3 lines
- then, extra examples that address different problems: 1 example on "how can I run composer install in docker?", 1 example on "how can I install extra extensions"

I would also document how to deploy an existing project, not one that is created (no `composer require bref` + `echo`). Because I expect that at this point, people reading that documentation will want to upload `.` (their existing project) to Lambda.

mnapoli: Also just to make it more explicit:

```suggestion
# Configure the handler file (the entrypoint that receives all HTTP requests)
CMD ["public/index.php"]
```

mnapoli:
> why put it in the Web Apps docs ? Seems to apply to both event-driven functions and web apps but there isn't a common place for this in the current docs I think.

That's true, but I think that is a good call: Docker images will only be useful for:

- those that hit the max size, and these projects usually use a full framework like Laravel or Symfony
- those that have a legacy project and Docker helps "normalize" the environment -> these projects are usually not "function-first"

So it would make sense to me not to consider deploying "functions" using containers as a main use case.

> I think I read that big images have a huge impact on cold starts. Maybe it's worth mentioning.

Oh it turns out it's the opposite: containers have a flat cold start. It's always ~250ms, no matter the size.

> i also think it would it would be nice to include a comparison between Bref runtimes and custom Docker image to guide the users toward the right solution for them. IMHO it targets specific use-cases where you need more control over the runtime but should not be the go-to solution for everyone.

Yes, very good point. It would make sense to say explicitly: "use Docker as a last resort". Docker shouldn't be the default.

deleugpn:
I know that the cold start is probably something worth documenting, but testing performance is something hard, with AWS Lambda and cold start it's even harder. Aside from that open area (performance) is there anything that would be worth mentioning when running Docker on Lambda?

mnapoli:
🎉  thank you @deleugpn! 

I'll merge and build the website

PR #978, opened 2021-07-22
Build PHP Binary with AWS CodeBuild

deleugpn:
This pull request sets the foundation to get us closer to automatically building new layers. Once this is merged, the process will still rely on the following steps:

- A pull request bringing the new PHP version to the docker files
- After merging, manually triggering AWS CodePipeline on the Layer Builder Account
- Once the Layer Builder Account finishes, we still need to manually run `make layers.json` and git commit + push
- Tag a new release

Future work may try to automate even further some of these processes, but for now we'll at least have a reproducible environment with AWS CodeBuild.

### Why CodeBuild?

The AWS Account that provides the layers are extremely closed down and secured. No Access Key or Secret Key needs to exists or be exposed. With AWS CodeBuild on a Builder Account we may `assume-role` into the AWS Account that provides the layers just with the least privilege access to publish new layers.

### How to install?

- [x] First the `builder.yaml` template should be provisioned on the Builder Account. This will set the IAM role which the layer account should allow
- [x] Go to CodeStar and connect to GitHub for access to the `brefphp/bref` repository
- [x] Then in the Layer Account the `publisher.yaml` template should be provisioned.

Now we can hit `release changes` on the AWS CodePipeline created and new layers will be published on the Layer Account.

TODO: change the branch name to `master` once it's merged.


runtime/pipeline/builder.yaml
```diff
@@ -0,0 +1,164 @@
...
+          - Name: DOCKER_BUILDKIT
+            Value: 1
+        Image: aws/codebuild/amazonlinux2-x86_64-standard:3.0
+        PrivilegedMode: true
+        Type: LINUX_CONTAINER
+      ServiceRole: !Ref CodeBuildRole
+      Source:
+        Type: CODEPIPELINE
+        BuildSpec: runtime/pipeline/buildspec.yaml
+      TimeoutInMinutes: 120
```
deleugpn, line 36: Maybe we can decrease this. It's usually taking less than 30 minutes to finish.

runtime/pipeline/publisher.yaml
```diff
@@ -0,0 +1,27 @@
...
+      Policies:
+        - PolicyName: LayerPublisherRole
+          PolicyDocument:
+            Statement:
+              - Effect: Allow
+                Action:
+                  - lambda:PublishLayerVersion
+                  - lambda:AddLayerVersionPermission
+                Resource:
+                  - '*'
```
deleugpn, line 27: Does it matter if we allow any layer to be published or just an explicit list of layers?

runtime/pipeline/publisher.yaml
```diff
@@ -0,0 +1,27 @@
...
+Resources:
+  LayerPublisherRole:
+    Type: AWS::IAM::Role
+    Properties:
+      RoleName: bref-layer-publisher
+      AssumeRolePolicyDocument:
+        Statement:
+          Effect: Allow
+          Principal:
+            AWS: !Sub "arn:aws:iam::${BuilderAccount}:role/bref-php-binary-builder"
```
deleugpn, line 16: If this role doesn't already exists in the builder account, this will error out. The reason for that is because IAM will resolve this role uniquely. If the role gets deleted and recreated it still won't work anymore as a protection against permission escalation.

runtime/pipeline/buildspec.yaml
```diff
@@ -0,0 +1,24 @@
+version: 0.2
+
+phases:
+  install:
+    runtime-versions:
+      docker: 19
+      php: 7.3
```
deleugpn, line 7: We're building with the minimum supported bref version.

PR #979, opened 2021-07-27
Added --update-with-dependencies for Laravel installation

Chtiadrien:
While waiting https://github.com/aws/aws-sdk-php/issues/2264 to be addressed, we have to use the `--update-with-dependencies` option on `composer require` when installing the laravel-bridge package, otherwise we get a composer error on a fresh install.


mnapoli:
Thanks a lot!

Issue #901, opened 2021-03-29
Decrypt SSM at runtime


allan-simon:
Currently the procedure explained by the documentation has a major drawback in term of security, as anyone with access to the AWS console can see the decrypted variables into  the lambda configuration.  Instead the best practices is to decrypt at runtime (once during the init phase of the lambda)

Not being a symfony, and even less a bref expert, I'm wondering if it's something that can be directly done transparently by bref so that the other of the application do not need to know the environment variables where encrypted 

(currently in my EC2 deployement for example, the .env.local file  is encrypted with KMS and uncrypted during provisionning)

mnapoli:
Bref documents the basic way to get started. If a company has more advanced requirements regarding configuration variables, then decrypting secrets at runtime makes sense, but it's up to them.

I'm more than happy to merge a pull request that adds a paragraph [in this section](https://bref.sh/docs/environment/variables.html#secrets) mentioning that this is a possibility. Even a link to an example, a blog post, or a package.

But I don't think this should be baked in Bref. You would get the same problem with any other hosting solution, you'd have to do something to decrypt the secrets. What you mentioned with EC2 is not solved by PHP, so I think it shouldn't be solved by Bref.

Maybe a PHP package? Or a Symfony bundle? Maybe a serverless plugin? Maybe something else?

matthewsuan:
@mnapoli i how do we hook up a function or code during the init phase using Bref? I looked into `bootstrap` file but cant find anything there that we can hook up to.

allan-simon:
I understand your point and they make sense , however I would have one counter argument:

  * bref is made specifically to my understanding to be a AWS Lambda <-> PHP wrapper to make AWS Lambda an implementation detail, as such it does not seem that strange to me , that bref also abstract the secret management part,
 

however I have to admit I don't know yet how I would implement it on bref side in way that is not too much opiniated

@matthewsuan right now what I'm  doing is putting my code loading the environment variable in the index.php , just before booting the kernel (where the DotEnv files are read)

mnapoli:
> how do we hook up a function or code during the init phase using Bref? I looked into bootstrap file but cant find anything there that we can hook up to.

I get this question a lot and always answer the same thing :) -> how would you solve that in any other PHP project?

Here is what I would suggest: in `index.php` or your framework bootstrap fetch the secrets and decrypt them. Then cache them (on disk, in APCu, or anywhere you want). It's simple and decoupled from Lambda.

[...]

PR #982, opened 2021-08-03
Custom extra bootstrap step

paulhenri-l:
While using bref with Laravel I wanted to cache the framework's config 
but after looking a bit at the source of both bref and laravel/vapor-core 
I realized this can only be done while bootstraping the runtime.

Caching Laravel's configuration is one use case for this new feature but
it could also be used to pre load some variables from ssm into the environement
just as laravel/vapor-core does.

I am actually pretty sure there is a broad array of potential use cases that I
haven't though about. So all in all letting people run a bit of their code during
the bootstrap process seems like a good idea.

### DIFF
--- a/runtime/layers/console/bootstrap
+++ b/runtime/layers/console/bootstrap
@@ -26,6 +26,12 @@ if (getenv('BREF_DOWNLOAD_VENDOR')) {
     require $appRoot . '/vendor/autoload.php';
 }
 
+$customBootstrap = $appRoot . '/lambda_bootstrap.php';
+if (file_exists($customBootstrap)) {
+    /** @noinspection PhpIncludeInspection */
+    require_once $customBootstrap;
+}
+
 $lambdaRuntime = LambdaRuntime::fromEnvironmentVariable();
 
 $handlerFile = $appRoot . '/' . getenv('_HANDLER');
--- a/runtime/layers/fpm/bootstrap
+++ b/runtime/layers/fpm/bootstrap
@@ -25,6 +25,12 @@ if (getenv('BREF_DOWNLOAD_VENDOR')) {
     require $appRoot . '/vendor/autoload.php';
 }
 
+$customBootstrap = $appRoot . '/lambda_bootstrap.php';
+if (file_exists($customBootstrap)) {
+    /** @noinspection PhpIncludeInspection */
+    require_once $customBootstrap;
+}
+
 $lambdaRuntime = LambdaRuntime::fromEnvironmentVariable();
 
 $handlerFile = $appRoot . '/' . getenv('_HANDLER');
--- a/runtime/layers/function/bootstrap.php
+++ b/runtime/layers/function/bootstrap.php
@@ -24,6 +24,12 @@ if (getenv('BREF_DOWNLOAD_VENDOR')) {
     require $appRoot . '/vendor/autoload.php';
 }
 
+$customBootstrap = $appRoot . '/lambda_bootstrap.php';
+if (file_exists($customBootstrap)) {
+    /** @noinspection PhpIncludeInspection */
+    require_once $customBootstrap;
+}
+
 $lambdaRuntime = LambdaRuntime::fromEnvironmentVariable();
 
 $container = Bref::getContainer();


paulhenri-l:
For context here is how it's done in vapor

[https://github.com/laravel/vapor-core/blob/2.0/stubs/fpmRuntime.php#L58-L66]

deleugpn:
Hi @paulhenri-l! Thanks for the contribution! I'll wait for @mnapoli to have a look, but usually we have been resistent to bring this kind of change into Bref because it steer us away from being similar to native runtimes. E.g. if you were to write a NodeJS lambda function, you would not have this functionality natively and you would have to write it yourself. Bref philosofy falls into a similar perspective and tries to make PHP available in a compatible manner with other programming languages.
While having a pre-invocation step could be useful, it adds a maintenance complexity that derails from the purpose of Bref.

You are encouraged to do this on your own `bootstrap/app.php` file and/or `public/index.php`

paulhenri-l:
@deleugpn oh I see, that actually makes perfect sense 😄

I had though about `bootstrap/app.php` and `public/index.php` but doing that during bootstrap would have been a bit more efficient.

mnapoli:
Hi! Yes that sounds similar to https://github.com/brefphp/bref/issues/901#issuecomment-809486183

To add on top of Marco's response (which I completely agree with): if you need to cache things, why not take the usual approach:

- if the file doesn't exist, create it
- else, use it

No need to hook into internals of Bref with that right?

paulhenri-l:
@mnapoli that was my initial idea but I went and looked at how it's done in vapor.

In terms of performance their solution is a tad better. But I guess I can live with that 😄

t-richard:
Hare are my 2 cents about this, I think that this solution has 2 main benefits over doing it into your entrypoint (index.php).
At init time :
* Lambda have bursted CPU power so this is more efficient, performant and could help reduce cold starts
* this boostrap script would be executed on warmer scheduled events or if using provisionned concurrency

> this kind of change into Bref because it steer us away from being similar to native runtimes

I think this is already the case, but it's a good thing. Bref is in this unique position to be able to provide more than just a runtime. And it's already doing it with for example : 
* ability to download the vendor directory with the breftoolbox and separateVendor. The Node runtime will not download node_modules for you
* native handling of warmer events
* running FPM in Lambda
* providing a console layer
* providing built-in extensions and support for `ini` files (and even more with the extra extensions)

All of that is unique to Bref and PHP, and adoption would certainly be lower if Bref wasn't doing these things.

deleugpn:
The problem is that there seems to be a desire to mix two different behaviors. Let's look at AWS Lambda in general:

```
<?php

$bootstrap = MyBootstrappingProcess::start();

return function (array $event) {
    // My lambda code
}
```

You get some code to run once for each warm-up lambda instance and you can do this on any AWS Lambda native language.

----

```
<?php

if (! is_file('/tmp/my-bootstrapping)) {
    $bootstrap = MyBootstrappingProcess::start();

    touch('/tmp/my-bootstrapping');
}

// index.php logic
```

This is the model behind FPM. No matter where you're running FPM (EC2, ECS, Digital Ocean, AWS Lambda), if you're running PHP on a Web Hosting solution, you can do this.

The foundation of the issue is that there seems to be an expectation that Bref will bridge the gap between then two and we're not really sure if we should. 

- If you choose the PHP Layer, you're closer to AWS Lambda. 
- If you choose the FPM Layer, you're closer to Web Hosting.

There isn't currently an option that will let you do what the PHP Layer does while using the FPM Layer. Should we have a special file that let people bridge the gap between Web Hosting and AWS Lambda? I don't know.

paulhenri-l:
@deleugpn In the end in ended up adding the init script inside the bootstrap.php file and I am using apcu to check wether initialization has already been done or not.

I think this solution works well and actually agree with you. Bref should remain as simple as possible and only focus on providing a php runtime. Other languages do not provide such features and there's no reason bref should be an exception. It makes the whole experience simpler.

The only thing that is a bit inconvenient is that the warm event does not trigger my script. But I replaced it with a fake API Gateway event so that it will trigger my init script.

PR #980, opened 2021-07-29
PHP 8.1 runtimes

mykiwi:
This pull request adds [PHP 8.1](https://www.php.net/archive/2021.php#2021-08-19-1) runtimes (PHP 8.1.0beta3)
Heavily inspired from #678

### To test it:

Name | Version | Compatible runtimes | Version ARN
-- | -- | -- | --
php-81 | 3 | provided.al2 | arn:aws:lambda:us-east-1:670541007614:layer:php-81:3
php-81-fpm | 3 | provided.al2 | arn:aws:lambda:us-east-1:670541007614:layer:php-81-fpm:3

PR #991, opened 2021-08-06
Test the codebase against php 8.1

t-richard:
As PHP 8.1 is coming and support is already considered in https://github.com/brefphp/bref/pull/980, I think its needed to run tests against 8.1 in the CI to ensure that the code doesn't break and eventually spot something, be it in Bref, php or a downstream dependency.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,6 +43,8 @@ jobs:
                     # sudo apt-get install -y php$version-fpm
                     whereis php
                     which php
+                    whereis php-fpm
+                    php-fpm -v
                     sudo cp /usr/sbin/php-fpm$version /usr/bin/php-fpm # copy to /usr/bin
                     echo 'expose_php=1' | sudo tee -a /etc/php/$version/fpm/php.ini
             -   name: Install dependencies
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,9 @@ jobs:
                 env:
                     version: ${{ matrix.php }}
                 run: |
-                    sudo apt-get install -y php$version-fpm
+                    # sudo apt-get install -y php$version-fpm
+                    whereis php
+                    which php
                     sudo cp /usr/sbin/php-fpm$version /usr/bin/php-fpm # copy to /usr/bin
                     echo 'expose_php=1' | sudo tee -a /etc/php/$version/fpm/php.ini
             -   name: Install dependencies
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,6 @@ jobs:
         runs-on: ubuntu-latest
         timeout-minutes: 15
         strategy:
-            fail-fast: false
             matrix:
                 php: [ '8.0', '7.4', '7.3' ]
                 dependency-version: [ '' ]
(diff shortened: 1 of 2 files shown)


t-richard:
Darn, FPM 8.1 is not available via apt-get yet. I'll try to figure this out

mnapoli:
Hey just FYI in case that helps, I had trouble with FPM too when adding tests for PHP 8 beta: https://github.com/shivammathur/setup-php/issues/280#issuecomment-703562986

t-richard:
@mnapoli thanks, I saw that, this issue is linked in the ci.yml file. Also just saw that @mykiwi had the same issue with the CI in #980 and ended up reverting the changes. 

As far as I can tell, FPM is installed with 8.1, just have to find the right way to make both co-exist.

t-richard:
@mnapoli the workflow is fixed for 8.1 but the tests are failing because `full_path` was introduced in `$_FILES` so the assert is not an exact match at https://github.com/brefphp/bref/blob/master/tests/Handler/FpmHandlerTest.php#L735 

See the php-src PR https://github.com/php/php-src/pull/6917

I've provided a fix for it, let me know if you think that's ok to it this way.

deleugpn (approved):

tests/Handler/FpmHandlerTest.php
```diff
@@ -1117,7 +1117,7 @@ private function assertGlobalVariables(array $event, array $expectedGlobalVariab
         // Test global variables that never change (simplifies all the tests)
         $response = $this->assertCommonServerVariables($response, $expectedGlobalVariables);
 
-        self::assertEquals($expectedGlobalVariables, $response);
+        self::assertArraySubset($expectedGlobalVariables, $response);
```
mnapoli, line 1120: Regarding your comment about this: we should aim to keep this as strict as possible. In the past, I went through a lot of hoops to keep this `assertEquals` (as you can see with the weird `assertCommonServerVariables()` above).

The reason is to have 100% confidence about the environment variables available in FPM.

I would be fine with a little hack, e.g. adding `unset()` (to remove the extra key) above with a comment explaining that it's specifically for PHP 8.1 (or anything like that). That would allow to keep the strictness here.

t-richard: I understand, I'll look into it.

t-richard: @mnapoli @deleugpn fixed to tests to keep the assertion strict

.github/workflows/ci.yml
```diff
@@ -18,9 +18,12 @@ jobs:
             matrix:
                 php: [ '8.0', '7.4', '7.3' ]
                 dependency-version: [ '' ]
+                platform-reqs: [ '' ]
                 include:
                     -   php: '7.3'
                         dependency-version: '--prefer-lowest'
+                    -   php: '8.1'
+                        platform-reqs: '--ignore-platform-req=php' # see https://github.com/phpspec/prophecy/pull/533#issuecomment-891606803
```
deleugpn, line 26: Is this still relevant now that the PR has been merged?

t-richard: The PR is merged but prophecy is not released and it won't be until php8.1 issues are fixed. I think it can be merged as is and removed when prophecy is compatible (it will be around RC1 apparently).

tests/Handler/FpmHandlerTest.php
mnapoli: Thank you!

mnapoli (approved):
LGTM, thank you @t-richard!

deleugpn:
Thank you @t-richard

PR #996, opened 2021-08-10
Support for events without size attribute

martijngastkemper:
I stumbled upon an undocumented inconsistency in the AWS API. The `s3:ObjectRemoved:Delete` event should have a `size` attribute. [The documentation says so](https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-content-structure.html).

I've send a message to AWS about this, but it seems to be an old problem. [This Stackoverflow question](https://stackoverflow.com/questions/48546482/how-to-get-file-size-from-aws-s3-objectremoveddelete-event) from January 2018 mentions the same issue.

But this inconsistency breaks Bref, because it requires a `size` attribute in `S3Record->getObject()`. This solution is a bit hacky, but making `BucketObject->size` nullable or remove it for this kind of events breaks backward compatibility.

Hopefully we can find a solution for this annoying problem.

### DIFF
--- a/src/Event/S3/S3Record.php
+++ b/src/Event/S3/S3Record.php
@@ -41,7 +41,7 @@ final class S3Record
     public function getObject(): BucketObject
     {
         $bucket = $this->record['s3']['object'];
-        return new BucketObject($bucket['key'], $bucket['size'], $bucket['versionId'] ?? null);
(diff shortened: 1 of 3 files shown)


deleugpn (approved):
LGTM

mnapoli (approved):
LGTM too!

Issue #997, opened 2021-08-20
No option for Console application


opensussex:
When I run `vendor/bin/bref init` the only options I get are

```
 What kind of lambda do you want to create? (you will be able to add more functions later by editing `serverless.yml`) [Web application]:
  [0] Web application
  [1] Event-driven function
```
Doesn't appear to be an option for Console applications (as described in the documentation here - https://bref.sh/docs/runtimes/console.html) 

I'm not sure what version of bref (latest I suppose) as when I do `vendor/bin/bref -V` I get `Deploy serverless PHP applications
`

t-richard:
It's the case in the latest version of Bref. (see [this](https://github.com/brefphp/bref/blob/db47a692a5d73665982e92361933aaed42848e72/bref#L47))

It was removed on purpose in https://github.com/brefphp/bref/pull/791, I think these docs splipped through.

If your app is console only, you may have to create the `serverless.yml` by hand or use the `Event-driven function` option in the init command and add the layer as pointed out in the docs.

Even if your use case seems like a totally valid one, I think the `console` part is mainly intented to be used along PHP frameworks such as Laravel (artisan) or Symfony (bin/console) and that's why it is now part of the `Bref for Web Apps` section of the docs.

opensussex:
Thanks for the reply @t-richard and what you're saying makes sense.

PR #998, opened 2021-08-20
Remove init command for console apps in the docs

t-richard:
As reported in https://github.com/brefphp/bref/issues/997, the `init` command doesn't offer the option to create a `Console Application` anymore.

If there is no plan to reintroduce it, this should be removed to avoid confusion.


mnapoli:
Thanks!

PR #1007, opened 2021-08-29
feat(runtimes): build and publish PHP 8.1 docker images

shouze:
:whale: will allow to test locally bref with php 8.1 when the docker images will get published :tada: 
:point_right: Related to https://github.com/brefphp/bref/pull/980/files#r695552502
:warning: For now xdebug & blackfire are not included because no php 8.1 version available yet.

### DIFF
--- a/runtime/layers/fpm-dev/Dockerfile
+++ b/runtime/layers/fpm-dev/Dockerfile
@@ -1,4 +1,16 @@
 ARG PHP_VERSION
+FROM bref/php-${PHP_VERSION}-fpm as without_extensions
+
+# Override the config so that PHP-FPM listens on port 9000
+COPY php-fpm.conf /opt/bref/etc/php-fpm.conf
+
+EXPOSE 9001
+
+# Clear the parent entrypoint
+ENTRYPOINT []
+
+ENV PHP_INI_SCAN_DIR="/opt/bref/etc/php/conf.d:/var/task/php/conf.d:/var/task/php/conf.dev.d"
+
 FROM bref/build-php-$PHP_VERSION as build_extensions
 
 RUN pecl install xdebug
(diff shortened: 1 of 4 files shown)


runtime/layers/fpm-dev/Dockerfile
```diff
@@ -13,9 +13,8 @@ USER root
 COPY --from=build_extensions /tmp/*.so /tmp/
 RUN cp /tmp/*.so $(php -r "echo ini_get('extension_dir');")
 
-FROM bref/php-${PHP_VERSION}-fpm
+FROM bref/php-${PHP_VERSION}-fpm as without_extensions
```
shouze, line 16: Will help for other new php versions upgrade as most of the time blackfire or xdebug are not ready during beta releases.

runtime/layers/fpm-dev/Dockerfile
```diff
@@ -29,3 +28,7 @@ ENV PHP_INI_SCAN_DIR="/opt/bref/etc/php/conf.d:/var/task/php/conf.d:/var/task/ph
 # Run PHP-FPM
 # opcache.validate_timestamps=1 : cancels the flag in the base configuration so that files are reloaded
 CMD /opt/bin/php-fpm --nodaemonize --fpm-config /opt/bref/etc/php-fpm.conf -d opcache.validate_timestamps=1 --force-stderr
+
+FROM without_extensions
```
mnapoli, line 32: Why the new `FROM` here? We would lose the EXPOSE, ENV and everything defined above?

mnapoli:
Thanks! I added a minor feedback, and it seems the build is not running in GitHub Actions

runtime/layers/fpm-dev/Dockerfile
shouze: I've checked, `ENV`, `EXPOSE`, `CMD` & `ENTRYPOINT` will be inherited from the above, the only difference is that we won't copy extensions (xdebug, blackfire).

shouze:
> Thanks! I added a minor feedback, and it seems the build is not running in GitHub Actions

Thanks, everything should be good now. Seems that some versions of docker build previous targets so I moved without_extensions one at first position and fixed the CI

runtime/layers/fpm-dev/Dockerfile
mnapoli: Ohh I think I see it now, thank you!

mnapoli:
Thank you!

Just to be sure: no need for new layers here, right?

shouze:
> Thank you!
> 
> 
> 
> Just to be sure: no need for new layers here, right?

New layers have been published in previous PR so no, only need to publish docker images.

shouze:
BTW a new tag should be required regarding GitHub action
https://github.com/brefphp/bref/blob/master/.github/workflows/deploy-docker.yml

mnapoli:
👍  the new release should have done it by now

shouze:
>  the new release should have done it by now

Thanks a lot... but I have bad news, an upfix seems necessary https://github.com/brefphp/bref/pull/1008

mnapoli:
For the record I'm re-enabling the extensions in #1114

PR #1008, opened 2021-08-29
fix(runtimes): docker file command of fpm-dev was misplaced

shouze:
@mnapoli I'm confused, my last #1007 refactor was incorrect :disappointed:

### DIFF
--- a/runtime/layers/fpm-dev/Dockerfile
+++ b/runtime/layers/fpm-dev/Dockerfile
@@ -32,3 +32,6 @@ RUN cp /tmp/*.so $(php -r "echo ini_get('extension_dir');")
 FROM without_extensions
 
 COPY --from=build_dev  /opt /opt
+
+# Override the config so that PHP-FPM listens on port 9000
+COPY php-fpm.conf /opt/bref/etc/php-fpm.conf
--- a/runtime/layers/fpm-dev/Dockerfile
+++ b/runtime/layers/fpm-dev/Dockerfile
@@ -4,12 +4,16 @@ FROM bref/php-${PHP_VERSION}-fpm as without_extensions
 # Override the config so that PHP-FPM listens on port 9000
 COPY php-fpm.conf /opt/bref/etc/php-fpm.conf
 
+ENV PHP_INI_SCAN_DIR="/opt/bref/etc/php/conf.d:/var/task/php/conf.d:/var/task/php/conf.dev.d"
+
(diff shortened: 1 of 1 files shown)


runtime/layers/fpm-dev/Dockerfile
```diff
@@ -25,10 +29,6 @@ USER root
 COPY --from=build_extensions /tmp/*.so /tmp/
 RUN cp /tmp/*.so $(php -r "echo ini_get('extension_dir');")
 
-# Run PHP-FPM
-# opcache.validate_timestamps=1 : cancels the flag in the base configuration so that files are reloaded
-CMD /opt/bin/php-fpm --nodaemonize --fpm-config /opt/bref/etc/php-fpm.conf -d opcache.validate_timestamps=1 --force-stderr
```
shouze, line 30: This was not the right place

t-richard:
> I'm confused, my last #1007 refactor was incorrect 😞

Hey that's okay, you're doing great 💪 ❤️

shouze:
I've already prepared extension's PR, we have a chance to get ready for first php 8.1 rc relase, septembre 2nd!
https://github.com/brefphp/extra-php-extensions/pull/261

runtime/layers/fpm-dev/Dockerfile
```diff
@@ -25,10 +29,9 @@ USER root
...
-
 FROM without_extensions
 
 COPY --from=build_dev  /opt /opt
```
shouze, line 34: This is a bit sad to copy `php-fpm.conf` twice, I just wonder if we cannot refine this one to:
```suggestion
COPY --from=build_dev  /opt/bref/lib/php/extensions/ /opt/bref/lib/php/extensions/
```

This would allow to remove the COPY instruction below, WDYT?

mnapoli: TBH I have no problem with a duplicate line or something that isn't 100% perfect for a temporary solution (when PHP 8.1 and xdebug support are here I guess this can be all simplified).

mnapoli:
No problem @shouze, thanks for the follow up PR!

Should I merge (since you opened up a question inline)?

shouze:
> No problem @shouze, thanks for the follow up PR!
> 
> Should I merge (since you opened up a question inline)?

Ok so you can merge, sorry again for the previous mistake :pray:

PR #1009, opened 2021-09-01
Document event/context for PSR-15 handlers

rdarcy1:
Add documentation for how to get the event and context objects within a PSR-15 handler. I couldn't see it documented already and found it by looking through PRs – related to #543.

### DIFF
--- a/docs/function/handlers.md
+++ b/docs/function/handlers.md
@@ -225,12 +225,13 @@ $id = $request->getAttribute('id');
 ```
 
 ### Event and context
+
 The HTTP request event and context are available as attributes on the request:
 ```php
-// Instance of Bref\Event\Http\HttpRequestEvent
+/** @var $event Bref\Event\Http\HttpRequestEvent */
 $event = $request->getAttribute('lambda-event'); 
 
-// Instance of Bref\Context\Context
+/** @var $context Bref\Context\Context */
 $context = $request->getAttribute('lambda-context');
 ```
 
--- a/docs/function/handlers.md
+++ b/docs/function/handlers.md
@@ -224,7 +224,9 @@ Note that path parameters (e.g. `{id}` in the example above) are available as re
 $id = $request->getAttribute('id');
 ```
 
-### Event and context
+[Full reference of HTTP events in `serverless.yml`](https://www.serverless.com/framework/docs/providers/aws/events/http-api/).
+
+### Lambda event and context
 
 The API Gateway event and Lambda context are available as attributes on the request:
 ```php
@@ -239,8 +241,6 @@ If you're looking for the request context array, for example when using a [Lambd
 ```php
 $requestContext = $request->getAttribute('lambda-event')->getRequestContext(); 
 ```
-
-[Full reference of HTTP events in `serverless.yml`](https://www.serverless.com/framework/docs/providers/aws/events/http-api/).
 
 ## Websocket events
 
--- a/docs/function/handlers.md
+++ b/docs/function/handlers.md
@@ -226,7 +226,7 @@ $id = $request->getAttribute('id');
 
 ### Event and context
 
-The HTTP request event and context are available as attributes on the request:
+The API Gateway event and Lambda context are available as attributes on the request:
 ```php
 /** @var $event Bref\Event\Http\HttpRequestEvent */
 $event = $request->getAttribute('lambda-event'); 
--- a/docs/function/handlers.md
+++ b/docs/function/handlers.md
@@ -225,7 +225,7 @@ $id = $request->getAttribute('id');
 ```
 
 ### Event and context
-The HTTP request event and Lambda context array are available as attributes on the request:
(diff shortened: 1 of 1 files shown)


docs/function/handlers.md
```diff
@@ -224,6 +224,21 @@ Note that path parameters (e.g. `{id}` in the example above) are available as re
 $id = $request->getAttribute('id');
 ```
 
+### Event and context
```
Nyholm, line 227: Please add a empty line after the heading

docs/function/handlers.md
```diff
@@ -224,6 +224,21 @@ Note that path parameters (e.g. `{id}` in the example above) are available as re
...
+The HTTP request event and context are available as attributes on the request:
+```php
+// Instance of Bref\Event\Http\HttpRequestEvent
```
Nyholm, line 230: Is it better to use PHP config?

```php
/** @var $event Bref\Event\Http\HttpRequestEvent */
```

docs/function/handlers.md
```diff
@@ -224,6 +224,21 @@ Note that path parameters (e.g. `{id}` in the example above) are available as re
...
+$event = $request->getAttribute('lambda-event'); 
+
+// Instance of Bref\Context\Context
+$context = $request->getAttribute('lambda-context');
+```
+
+If you're looking for the request context array, for example when using a [Lambda authorizer](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-lambda-authorizer.html#http-api-lambda-authorizer.payload-format-response):
```
Nyholm, line 237: In the Symfony docs there is a rule to not use "you" or "we". Maybe rewrite this paragraph to something like: 

> The request context array is accessible in a similar way. The context array is useful with the Lambda authorizer.

Nyholm (commented):
Thank you. 

I just had some minor style suggestions

docs/function/handlers.md
t-richard: I think it makes more sense :+1:

docs/function/handlers.md
```diff
@@ -224,6 +224,21 @@ Note that path parameters (e.g. `{id}` in the example above) are available as re
 $id = $request->getAttribute('id');
 ```
 
+### Event and context
+The HTTP request event and context are available as attributes on the request:
+```php
+// Instance of Bref\Event\Http\HttpRequestEvent
+$event = $request->getAttribute('lambda-event'); 
+
+// Instance of Bref\Context\Context
```
t-richard, line 233: Following the comment above 
```suggestion
/** @var $context Bref\Context\Context */
```

docs/function/handlers.md
rdarcy1: Thanks for the feedback. The Bref documentation does use the second person "you" throughout (e.g. [here](https://bref.sh/docs/function/handlers.html#autoloading) and [here](https://bref.sh/docs/function/local-development.html)) so I think it's in keeping with the style.

mnapoli: Yes, let's keep it in line with Bref practices instead of Symfony practices.

docs/function/handlers.md
```diff
@@ -224,6 +224,22 @@ Note that path parameters (e.g. `{id}` in the example above) are available as re
...
+/** @var $context Bref\Context\Context */
+$context = $request->getAttribute('lambda-context');
+```
+
+If you're looking for the request context array, for example when using a [Lambda authorizer](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-lambda-authorizer.html#http-api-lambda-authorizer.payload-format-response):
+```php
+$requestContext = $request->getAttribute('lambda-event')->getRequestContext(); 
+```
+
 [Full reference of HTTP events in `serverless.yml`](https://www.serverless.com/framework/docs/providers/aws/events/http-api/).
```
mnapoli, line 243: Can this line be moved above the new section? It seems to me like it belongs to what's above, not to "Lambda event and context"

docs/function/handlers.md
```diff
@@ -224,6 +224,22 @@ Note that path parameters (e.g. `{id}` in the example above) are available as re
 $id = $request->getAttribute('id');
 ```
 
+### Event and context
+
+The HTTP request event and context are available as attributes on the request:
```
mnapoli, line 229:
```suggestion
The API Gateway event and Lambda context are available as attributes on the request:
```

(because it's not the HTTP request, it's the low-level API Gateway event)

docs/function/handlers.md
mnapoli:
```suggestion
### Lambda event and context
```

mnapoli (commented):
Minor feedback and then for me it's good to merge, thank you!

rdarcy1:
Thanks @mnapoli, changes made.

mnapoli:
Perfect thank you!

PR #1010, opened 2021-09-03
Update php8.1 to use RC1

t-richard:
The first release candidate is out, let's test it !


mnapoli:
👍  thanks, let's do it!

PR #1013, opened 2021-09-05
Let LambdaRuntime::processNextEvent() return boolean

Nyholm:
I would like to know if a handler succeeds or not. Since the LambdaRuntime is echo out data on error, we might run into issues when keeping the process alive for multiple requests. 

See https://github.com/php-runtime/bref/blob/0.1.1/src/BrefRunner.php#L34

A simple fix would be to stop the `while(true)` on error.

### DIFF
--- a/src/Runtime/LambdaRuntime.php
+++ b/src/Runtime/LambdaRuntime.php
@@ -89,9 +89,10 @@ final class LambdaRuntime
      *     $lambdaRuntime->processNextEvent(function ($event, Context $context) {
      *         return 'Hello ' . $event['name'] . '. We have ' . $context->getRemainingTimeInMillis()/1000 . ' seconds left';
      *     });
+     * @return bool true if event was successfully handled
      * @throws Exception
      */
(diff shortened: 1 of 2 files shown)


t-richard (approved):
Looks good to me. Since this class is internal and we're moving from `void` to `bool`, it should not have an impact.

mnapoli:
+1 I'd be curious to know more about what would happen next.

Because this class is internal at the moment, if we make it public API then we need to have a strong & defined use case for that right? (else you'd have no BC promise)

Nyholm:
> +1 I'd be curious to know more about what would happen next.

Haha. I thought I could sneak in this change first before I made an argument for creating a public API. 

Because creating a public API requires some more thinking and discussions. But yes, I would like to make this class non-internal at some point in the future.

mnapoli:
Following discussions on Slack, we realized there is a sense to add this to Bref because we may want to use that feature to improve the "on-error" behavior.

Follow-up PR: #1026

PR #1014, opened 2021-09-06
[Docs] Document server-side construct in Symfony docs

t-richard:
This is a first shot at documenting the Lift server-side construct with Symfony.

### DIFF
--- a/docs/frameworks/symfony.md
+++ b/docs/frameworks/symfony.md
@@ -138,12 +138,12 @@ in `trusted_proxies`.
 
 ## Assets
 
-To deploy Symfony websites, assets needs to be served from AWS S3. The easiest solution to do this is to use the
+To deploy Symfony websites, assets need to be served from AWS S3. The easiest solution to do this is to use the
 [Server-side website construct of the Lift plugin](https://github.com/getlift/lift/blob/master/docs/server-side-website.md).
 
 This will deploy a Cloudfront distribution that will act as a proxy: it will serve
 static files directly from S3 and will forward everything else to Lambda. This is very close
-to how traditional web servers like Apache or Nginx work which means you application doesn't need to change!
+to how traditional web servers like Apache or Nginx work, which means your application doesn't need to change!
 For more details, see [the offical documentation](https://github.com/getlift/lift/blob/master/docs/server-side-website.md#how-it-works). 
 
 First install the plugin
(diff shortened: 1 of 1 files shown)


docs/frameworks/symfony.md
```diff
@@ -138,59 +138,48 @@ in `trusted_proxies`.
...
-```js
-// webpack.config.js
-
-if (Encore.isProduction()) {
-    // Note the '/build' at the end of the URL
-    Encore.setPublicPath('https://<bucket-name>.s3.amazonaws.com/build');
-    Encore.setManifestKeyPrefix('build/')
-}
+```yaml
+...
```
t-richard, line 158: Maybe we should add a more complete example like in Lift docs ?

mnapoli: Yes I think it could be worth it, because so far it could be confusing: "do I keep the lambda functions?" "where does this go in serverless.yml?", etc.

mnapoli:
That looks awesome! Thank you for getting this started.

Usually I'm not a fan of duplicating instructions between "Websites", "Laravel" and "Symfony", but that new construct makes it so simple that I'm fine with it :) Especially since the example can now contain the exact list of assets URL specifically for Symfony, that's awesome.

docs/frameworks/symfony.md
t-richard: I'm actually not sure how to tackle this. I don't want to put the whole serverless.yml either because that would mean having to maintain it in two places (here and the symfony/recipes-contrib repo). And that we tried to avoid in previous PRs like https://github.com/brefphp/bref/pull/923#discussion_r625660759

Maybe something like this would be a good compromise ? Really not sure, any suggestion appreciated :slightly_smiling_face: 

```yaml
service: symfony

provider:
    ...

plugins:
    - ./vendor/bref/bref
    - serverless-lift

constructs:
    website:
        type: server-side-website
        assets:
            '/bundles/*': public/bundles
            '/build/*': public/build
            '/favicon.ico': public/favicon.ico
            '/robots.txt': public/robots.txt
            # add here any file or directory that needs to be served from S3

functions:
   ...
```

mnapoli: Yes, this is exactly what I had in mind.

Just to show familiar stuff first, I would put `functions:` above `constructs:`

t-richard:
Forgot to mark this as ready when I updated code and docs last week. 

Let me know if there's anything I can do to help move this and/or the broader "Website assets" docs forward

mnapoli:
Oh cool, let's merge then, thanks a lot!

mnapoli:
If you have time to add something similar to the rest of the documentation I'm all for it!

In the "Website assets" docs we can remove the huge YAML template, we don't need to document that anymore.

t-richard:
@mnapoli What's your vision on this page ? Should we keep the part with s3 on a separate domain ?

mnapoli:
TBH I'm not 100% sure.

Maybe something like this (broad outline):

- Present "website + assets" architecture with CloudFront (https://bref.sh/docs/websites.html#same-domain)
  - "This architecture is more complex" -> this sentence can be removed
- Explain how to deploy that (with Lift)
- Alternative solutions (new section):
  - for those that don't want to use Lift, or don't want to use CloudFront, it is possible to set things up manually
  - link to the older version of the documentation in GitHub: https://github.com/brefphp/bref/blob/d1dd690d020cd03f134010db456bb61a6d0ffafb/docs/websites.md
- Feel free to remove this section too: https://bref.sh/docs/websites.html#more-examples because it will not be useful anymore

That way the documentation stays simple, focused on one approach. Yet we offer alternatives and the existing documentation (with full examples) is not 100% lost.

PR #1015, opened 2021-09-08
Write errors on stderr instead of echo

Nyholm:
To avoid issues like "output already sent" I want to make sure we never print to stdout. Not even when an error happens. 

Writing on strerr will still make the message show up on cloudwatch on both "function layer" and the FPM layer.

### DIFF
--- a/src/Runtime/LambdaRuntime.php
+++ b/src/Runtime/LambdaRuntime.php
@@ -217,7 +217,7 @@ final class LambdaRuntime
         // Log the exception in CloudWatch
         // We aim to use the same log format as what we can see when throwing an exception in the NodeJS runtime
         // See https://github.com/brefphp/bref/pull/579
-        echo $invocationId . "\tInvoke Error\t" . json_encode($errorFormatted) . PHP_EOL;
+        error_log($invocationId . "\tInvoke Error\t" . json_encode($errorFormatted) . PHP_EOL);
 
         // Send an "error" Lambda response
         $url = "http://{$this->apiUrl}/2018-06-01/runtime/invocation/$invocationId/error";
@@ -236,7 +236,7 @@ final class LambdaRuntime
     public function failInitialization(string $message, ?\Throwable $error = null): void
     {
         // Log the exception in CloudWatch
-        echo "$message\n";
+        error_log("$message\n");
         if ($error) {
             if ($error instanceof Exception) {
                 $errorMessage = get_class($error) . ': ' . $error->getMessage();


mnapoli:
> To avoid issues like "output already sent"

This should never happen though, that code never runs in PHP-FPM. It's the code that runs PHP-FPM.

Nyholm:
You are correct. I cannot happen in the PHP-FPM layer. (I just said that my PR works with the PHP-FPM layer)

This can happen when you run PSR-15 and using BREF_LOOP_MAX > 1. See this repo as an example: https://github.com/Nyholm/bref-issue-output-started

Here are some logs: 
Screenshot 2021-09-08 at 12 40 12

Sure this might look weird, but I did run into this issue when running `bref/runtime`. Creating session and cookies are valid as long as they are picked up by the runtime itself. 

I am trying to work around this issue first by make sure we never echo and secondly to make sure the runtime is aware when an error is happening: https://github.com/php-runtime/bref/blob/main/src/BrefRunner.php#L34

That is why I made the unofficial way in #1013 and the more official way in #1016

mnapoli:
That helps, thanks for sharing. I still feel like we are working from the details up to the original goal. It would work much better the other way around: if you explain the original goal, it will be much easier to understand _why_ we need to solve all these problems. Because these individual details are not actual problems for now.

Right now it still doesn't make sense:

- there is no "output already sent" in your screenshot: I don't see how stdout can mess up the "function" layer
- I would tell anyone using `session_start` (or any global feature not cleaned up) in the function layer to not change `BREF_LOOP_MAX`: why is this scenario different?

I really want to improve support for Symfony and the runtime component, but here we are working from details that don't make sense. It would be great to get the whole picture.

deleugpn:
Given that the function layer is practically a glorified `php-cli`, is it even possible to use `session_start()` there? What would the goal be?

Nyholm:
My goal is: I want the runtime be able to detect if there is an uncaught exception in the application. I want to be able to detect that because all requests after an uncaught exception failed for me in prod the other day. They failed because "output already sent". That is why I added #1013.

I added this PR to be extra sure that we don't write our errors to stdout, but to use strerr instead. 

I do know that an application shouldn't have uncaught errors and that it really would be better if that never happen, but that is nothing I can control for all users of runtime/bref. I thought I would prove that this scenario could happen by using non-runtime code and to use something that looks more normal (sessions) compared to using headers. But the results are the same. 

@deleugpn, yes it is possible. The purpose would be to start the session and get a session id. That id is later used in a cookie header in the AWS response. Just like in a normal application. 

There are 3 different ways to solve the problem: #1013, #1015 and #1016. #1013 is simplest and #1016 may require a bit more thinking since it is a public API. #1015 (or a version of it) is something that we should consider anyways since errors shouldn't really be on stdout.

> I would tell anyone using session_start (or any global feature not cleaned up) in the function layer to not change BREF_LOOP_MAX: why is this scenario different?

In a "normal" application, you don't need to clean that up. If there is an uncaught exception, it will be cleaned up automatically before the next request comes in. Since the [LambdaRuntime](https://github.com/brefphp/bref/blob/1.2.14/src/Runtime/LambdaRuntime.php#L106) is catching an not re-throwing exceptions, it is impossible for any bootstrap code (runtime or not) to clean things up (or kill) after the application had an uncaught exception.

GrahamCampbell:
What's the status here?

mnapoli:
Since we went with #1013/#1026 I think it's safe to close this one (let me know if this is a mistake).

PR #1016, opened 2021-09-08
Make LambdaRuntime throw errors

Nyholm:
As a follow-up on @deleugpn suggestion in https://github.com/brefphp/bref/pull/1013#issuecomment-913469456

This will define an API for LambdaRuntime. I did not make it non-internal yet. 

I make sure we only support BrefHandlers and we throw exceptions when something is wrong. If we like this approach I will continue working on this PR and update the tests. 

### TODO

- [ ] Update tests

### DIFF
--- a/runtime/layers/console/bootstrap
+++ b/runtime/layers/console/bootstrap
@@ -33,34 +33,38 @@ if (! is_file($handlerFile)) {
     $lambdaRuntime->failInitialization("Handler `$handlerFile` doesn't exist");
 }
 
-while (true) {
-    $lambdaRuntime->processNextEvent(function ($event, Context $context) use ($handlerFile): array {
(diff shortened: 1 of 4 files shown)


mnapoli:
Hey! I still don't understand the _why_: https://github.com/brefphp/bref/pull/1013#issuecomment-913480014 I'm not sure why we are doing all this.

mnapoli:
Same as #1015 I think we can now close this one.

Nyholm:
Im happy with the merge of #1013. I dont need this. 

@deleugpn expressed that this is a better solution than returning a boolean. I agree with you but this requires some more thinking and planning. We should maybe come back to this PR when we make `LambdaRuntime` non-internal.

PR #1020, opened 2021-09-16
Understand file include notation

abdala:
Serverless allows you to include external yml files using `${file(your-file/name.yml)}` notation. Like:

```
functions:
  - ${file(serverless/functions/web.yml)}
  - ${file(serverless/functions/artisan.yml)}
 ```

This PR will parse the file and add the file's content to the dedicated key if it matches the requested function name.

### DIFF
--- a/src/Console/Command/Local.php
+++ b/src/Console/Command/Local.php
@@ -98,6 +98,27 @@ class Local
 
         $serverlessConfig = Yaml::parseFile($file, Yaml::PARSE_CUSTOM_TAGS);
 
+        /**
+         * Verify and parse functions using `file()` notation:
+         *
+         * functions:
+         *   - ${file(serverless/functions/web.yml)}
+         *   - ${file(serverless/functions/artisan.yml)}
+         */
+        if (isset($serverlessConfig['functions'])) {
+            foreach ($serverlessConfig['functions'] as $key => $fileInclude) {
+                preg_match('/\$\{file\((.*?)\)\}/', $fileInclude, $matches);
+
+                if (is_int($key) && ! empty($matches[1])) {
+                    $parsedFunction = Yaml::parseFile($matches[1], Yaml::PARSE_CUSTOM_TAGS);
+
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by deleugpn — APPROVED** [review]

File: src/Console/Command/Local.php [hunk 1]
```diff
@@ -98,6 +98,20 @@ private function handlerFromServerlessYml(string $function, ?string $config): st
 
         $serverlessConfig = Yaml::parseFile($file, Yaml::PARSE_CUSTOM_TAGS);
 
+        if (isset($serverlessConfig['functions'])) {
```
mnapoli, line 101: Could you add a comment that explains _why_ this block of code exists? That will greatly simplify maintenance in the future.

deleugpn:
This will fix `bref local` when using `file()` on Serverless

mnapoli (commented):
Thank you for the contribution, just a minor thing and 👍  to merge for me

GrahamCampbell:
Usually I pref the syntax `- web: ${file(serverless/functions/web.yml)}` so I don't have to repeat the function name in the file, and moreover it means the direcyory and file structure perfectly matches up with the serverless file.

deleugpn:
uh, I see now that this change is very specific to `file` being parsed only when nested immediately below `functions` which would make `bref local` a bit too specific

mnapoli:
Yes, I agree this is very specific and limited fix.

In the future it might make sense to consider putting some of these Bref commands into the Serverless plugin. That way we get the whole behavior from the official CLI.

In the meantime I'm fine merging, as it's a very limited and local code addition (the risk is very limited).

deleugpn:
I think we need to revert this because of https://github.com/brefphp/bref/issues/1037

mnapoli:
👍  for a revert, unless @abdala you have an idea for a quick fix?

deleugpn:
I spoke with @abdala and he won't have time to look into this and suggested we revert it and once he gets a chance he'll look into whether it's possible to have a better support

PR #1026, opened 2021-09-19
Function runtime: force restarting the process for failed invocations

mnapoli:
This is a continuation of #1013.

In the function runtime, the PHP process will now always be restarted if the invocation failed. That change only impacts apps that use `BREF_LOOP_MAX` to keep the process alive.

The reason for this is that, in most cases, a failed invocation means the PHP application threw an exception.

When using `BREF_LOOP_MAX`, the PHP process is kept alive, and it is important for the application to cleanup global state.

If an exception was thrown, it is likely the cleanup wasn't done, or wasn't done properly. It sounds much safer to restart the process.

The downside is a slight performance cost: around 10ms to restart the process. But that provides a safer execution environment.

### DIFF
--- a/docs/environment/performances.md
+++ b/docs/environment/performances.md
@@ -90,6 +90,8 @@ In the example above, the PHP process will restart only every 100 invocations, r
 
 In that case, be careful with clearing in-memory data between every event.
 
+> Note: the PHP process will be restarted in case of a failed invocation (PHP exception thrown in the handler).
+
 ## Cold starts
 
 Code on AWS Lambda runs on-demand. When a new Lambda instance boots to handle a request, the initialization time is what we call a *cold start*. To learn more, [you can read this article](https://hackernoon.com/im-afraid-you-re-thinking-about-aws-lambda-cold-starts-all-wrong-7d907f278a4f).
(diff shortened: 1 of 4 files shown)


shouze (approved):

GrahamCampbell (approved):

t-richard (approved):

deleugpn:
As a rule of thumb, I don't like functions that have a boolean return as a way "silence" exceptions. More often than not it's more useful to move the try catch upwards so that not only you have the failure signal but also the ability to make different (non-binary) decisions based on information available on the exception.
On the other hand, since LambdaRuntime is an internal class, it's possible to make such a change at another point in time and disregard the BC break. This does look simpler than letting the exception bubble up specially in the limited use case.

deleugpn (approved):

GrahamCampbell:
This is a significant enough change, that I think this warrants bumping to 1.3.0, giving people a little more control over upgrading.

mnapoli:
👍  good point

Nyholm:
Thank you for merging. I've followed up by https://github.com/php-runtime/runtime/pull/82

PR #1028, opened 2021-09-19
Remove old workarounds for PHP 8.1

GrahamCampbell:


t-richard (approved):
Oh right, prophecy is compatible with 8.1 now 🚀

deleugpn (approved):

PR #1031, opened 2021-09-19
Collect which layer was used for invocation

deleugpn:
This Pull Request introduces a new StatsD metric to analyse Bref usage among different layers. One open question before merging is whether we want to collect PHP Version in use. My initial idea is that we can follow Composer installation stats to see php version usage trend, which makes the layer the most important informmation to extract from Bref.

Since LambdaRuntime is an internal and final class, we won't consider this a breaking change.

### DIFF
--- a/runtime/layers/console/bootstrap
+++ b/runtime/layers/console/bootstrap
@@ -26,7 +26,7 @@ if (getenv('BREF_DOWNLOAD_VENDOR')) {
     require $appRoot . '/vendor/autoload.php';
 }
 
-$lambdaRuntime = LambdaRuntime::fromEnvironmentVariable();
+$lambdaRuntime = LambdaRuntime::fromEnvironmentVariable('console');
 
 $handlerFile = $appRoot . '/' . getenv('_HANDLER');
 if (! is_file($handlerFile)) {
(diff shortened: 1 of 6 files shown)


GrahamCampbell:
👍 would be another good thing to include in the 1.3.0 release.

Nyholm (approved):
This is a good idea. Thank you!

t-richard:
Seems like a good idea but wouldn't it be simpler and more accurate to send the php version at the same time rather than relying on packagist and crossing data?

mnapoli:
Awesome, this will help get a better vision of which runtimes are used the most.

This is helpful for Bref's own development, but also when talking with AWS.

For PHP versions, it's a "nice to have" but much less important than layers, because we can retrieve that information from somewhere else (https://packagist.org/packages/bref/bref/php-stats). I don't want to block that improvement.

Thanks Marco!

chekalsky:
@mnapoli  this was non-BC change :-( But I see @Nyholm already fixing this

deleugpn:
@chekalsky `LambdaRuntime` is an internal class.

Nyholm:
I made a release on [runtime/bref](https://github.com/php-runtime/bref/releases/tag/0.2.1) just a few minutes agoi. 

@deleugpn is correct. The `LambdaRuntime` is internal. The `runtime/bref` package should either consider creating it's own copy of `LambdaRuntime` or we should consider taking steps to make it non-internal.

chekalsky:
@deleugpn sorry, didn't meant that. Yes, would be great to mitigate possibility in the future

PR #1032, opened 2021-09-19
Add docs about applications with Symfony Runtime

Nyholm:
I thought about making this part of the existing Symfony page, however, it didn't really fit. First, the Runtime component works very well with Symfony apps, Laravel app, PSR-15 apps and pure PHP applications. I also didnt want to remove docs how to use Bref with non-runtime symfony applications. 

That is why I added a new page. This page describe how to install and deploy a pure PHP application with the Runtime component. If you want to add assets you still need to read the Symfony/Laravel docs. 

This PR follows the recipe: https://github.com/symfony/recipes/pull/1001

### DIFF
--- a/docs/frameworks/symfony-runtime.md
+++ b/docs/frameworks/symfony-runtime.md
@@ -4,7 +4,7 @@ current_menu: symfony-runtime
 introduction: Learn how to deploy serverless applications on AWS Lambda using Bref and Symfony Runtime component.
 ---
 
-The Symfony Runtime runtime component was release with Symfony 5.3. All new Symfony
+The Symfony Runtime component was released with Symfony 5.3. All new Symfony
 applications are using that component as default, but the component can also be
 used with non-Symfony applications.
 
--- a/docs/frameworks/symfony-runtime.md
+++ b/docs/frameworks/symfony-runtime.md
@@ -1,0 +1,48 @@
+---
+title: Serverless applications with Symfony Runtime
+current_menu: symfony-runtime
+introduction: Learn how to deploy serverless applications on AWS Lambda using Bref and Symfony Runtime component.
+---
+
+The Symfony Runtime runtime component was release with Symfony 5.3. All new Symfony
+applications are using that component as default, but the component can also be
+used with non-Symfony applications.
+
+This guide helps you run any PHP applications on AWS Lambda using Bref and the Symfony
+Runtime component.
+
+## Setup
+
+First, **follow the [Installation guide](../installation.md)** to create an AWS
+account and install the necessary tools.
+
+Next, in an any project using the [Symfony Runtime](https://symfony.com/doc/current/components/runtime.html),
+install the Bref Runtime package.
+
+```
+composer require runtime/bref
+```
+
+If you are using [Symfony Flex](https://flex.symfony.com/), it will automatically run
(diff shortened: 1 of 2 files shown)


deleugpn:
The resistence I have with this is that Bref first philosophy is 

> simplify problems by removing choices

While I think Symfony Runtime seems a very interesting concept, it brings a bit of overhead coupled with something that's experimental.

Nyholm:
Thank you for your quick reply. 

Could you elaborate what you mean with "overhead"? Are you talking about loading ~5 extra files? When I did [some tests](https://developer.happyr.com/performance-test-bref-runtime-component) I found using the runtime component was faster that starting PHP-FPM. 

Yes, the component is marked as "experimental" for about 2 more months. It is really unlikely there will be any BC breaks. I would not worry about adding docs about it. 

I know that Bref is all about simplicity. **I dont make a suggestion now**, but using the Runtime component will remove the need for the console layer, FPM layer and FPM-debug layer. It will solve [time timeout issue](https://github.com/brefphp/bref/pull/895) and make sure all PHP frameworks/apps have the same setup/install. But this can be discussed in another thread. This PR is just to explain how to deploy modern/symfony applications with Bref.

deleugpn:
By overhead I just mean cognitive overhead. It's a shiny new thing that has the potential to completely change how everyone runs PHP. Serverless is another shiny new thing. It has it's own set of resistence barriers. Although I think Serverless is where the Symfony Runtime will "shine the most", I feel it would lead to just more cognitive stress for people adopting serverless right now.

deleugpn:
I'm not super fond of the fact the user is "forced" to have their own bootstrap file in order to correctly use `runtime/bref` (source https://github.com/php-runtime/bref#how-to-use). That seems a step backwards compared to running Bref with the existing layers.

Nyholm:
I agree with you. Im not super excited about that either. It is very similar to the current Bref console layer. 

I will try to make that bootstrap file as a part of Bref in the future. But I decided to start with some documentation.

t-richard:
> I agree with you. Im not super excited about that either. It is very similar to the current Bref console layer.

And what about the runtime layer that was added in [bref/extra-php-extensions](https://github.com/brefphp/extra-php-extensions/tree/master/layers/symfony-runtime) ?

docs/frameworks/symfony-runtime.md
```diff
@@ -0,0 +1,48 @@
+---
+title: Serverless applications with Symfony Runtime
+current_menu: symfony-runtime
+introduction: Learn how to deploy serverless applications on AWS Lambda using Bref and Symfony Runtime component.
+---
+
+The Symfony Runtime runtime component was release with Symfony 5.3. All new Symfony
```
shouze, line 7:
```suggestion
The Symfony Runtime component was released with Symfony 5.3. All new Symfony
```

Nyholm: Thank you

Nyholm:
> And what about the runtime layer that was added in bref/extra-php-extensions ?

That is a good idea. I opened #889 a while ago to discuss that. There are many things we can do to make the runtime component to work better with Bref. I am happy to do all of them, but we need to start somewhere. I thought it was a good idea to start writing docs.

mnapoli:
As mentioned in #889:

🎉  Symfony runtime is now implemented and supported in https://github.com/brefphp/symfony-bridge

I've released a beta version, there should be a stable in the coming weeks!

Issue #756, opened 2020-09-21
Value of $_SERVER['QUERY_STINRG] different when deployed


DaMitchell:
**Description**
When making a request to `/orders?billingAddress.surname=test` and accessing the value of `$_SERVER['QUERY_STRING']` the expected value should be:

```
billingAddress.surname=test
```

Instead the value is:

```
billingAddress_surname=test
```

**Cause**
When an instance of `HttpRequestEvent` is created the method `HttpRequestEvent::rebuildQueryString` is called which passed the value through `parse_str` causing the conversion. This value is set on the request passed the the FPM process by:
```php
$request->setCustomVar('QUERY_STRING', $event->getQueryString());
```
Which is here https://github.com/brefphp/bref/blob/master/src/Event/Http/FpmHandler.php#L182. I can also see it being used in the `Psr7Bridge` here https://github.com/brefphp/bref/blob/master/src/Event/Http/Psr7Bridge.php#L33

DaMitchell:
@mnapoli Any ideas on when this can be addressed? I am happy to put a PR in myself just a little unsure about what you would expect in it. Other than updating the `HttpRequestEvent::getQueryString()` is there anything else you would expect to be updated?

mnapoli:
A first step that could be very useful would be to add a failing test case. You can start by looking at these tests: https://github.com/brefphp/bref/blob/master/tests/Event/Http/CommonHttpTest.php#L76-L126 and maybe try to write a new one?

If that doesn't reproduce, then maybe you'll want to add it here: https://github.com/brefphp/bref/blob/master/tests/Handler/FpmHandlerTest.php 

If you see a solution feel free to implement it too.

vasoft:
I am migrating a large Api-platform 2.7(SF 6.2, PHP 8.1) to Bref & Serverless, and I am blocked on this problem with (.) becoming (_). Our project has a lot of nested filters in the request query string, and they are not working anymore. 
I suppose anyone migrating Api-platform would face this issue.

mnapoli:
FYI this PR might be very close to be merged: https://github.com/brefphp/bref/pull/1383

sladg:
Heya! is there any movement on this issue? Experiencing exactly same issue (param names including dot have underscore in them once Bref stack is deployed).

ApiPlatform makes use of dots pretty standart and having them converted makes it rather difficult to deal with.

@mnapoli any work left / blocker on the PR mentioned? :)

mnapoli:
[...]

PR #1042, opened 2021-09-27
Rewrite query string parsing to avoid use parse_str

alekitto:
I've encountered the same problems described in #756, so I've checked how PHP implemented the parse string filter and rewrote it in PHP to avoid using `parse_str` itself.

Spaces and dots are preserved in the resulting array keys.
The behavior of the parser on case of unclosed bracket is retained too.

### DIFF
--- a/src/Event/Http/HttpRequestEvent.php
+++ b/src/Event/Http/HttpRequestEvent.php
@@ -339,35 +339,35 @@ final class HttpRequestEvent implements LambdaEvent
             }
 
             $tokens = explode('[', $key);
-            $current = &$parameters;
-            foreach ($tokens as $idx => $token) {
-                if ($idx !== 0 && substr($token, -1) !== ']') {
-                    // Malformed string: unclosed parameter index
-                    return [];
+            $tokenCount = count($tokens);
+            for ($i = 1; $i < $tokenCount; ++$i) {
+                if (substr($tokens[$i], -1) === ']') {
+                    continue;
                 }
 
+                // Malformed string: unclosed parameter index
+                $tokens = [implode('[', $tokens)];
+                break;
+            }
+
+            $current = &$parameters;
+            foreach ($tokens as $idx => $token) {
+                $token = preg_replace('/]$/', '', $token);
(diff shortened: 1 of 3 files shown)


deleugpn:
This is quite a complex change and somewhat a maintenance overhead for bref. I'll need some time before I can review this thoroughly. 

It would be good to know how many people are having issues with this to help with prioritization.

Cafeine42:
Hi,

We are having this problem with Api Platform too.

curl -X 'GET' \
  'https://xxxxxxx/api/v4/group_classes?page=1&activity.id=6' \
  -H 'accept: application/ld+json'
  
is handled with this query string :
  
   "hydra:view": {
    "@id": "/api/v4/group_classes?activity_id=6",
    "@type": "hydra:PartialCollectionView"
  },

The usage of dot is a basic feature in Api Plaform : 
https://api-platform.com/docs/core/filters/#filtering-on-nested-properties

I am amazed that so few people have encountered the problem

Cafeine42:
For more info, Api Platform have an implementation to handle this too :
https://github.com/api-platform/core/blob/10827bb570c44cdee2c1bb7a307a8a18df9b4f00/src/Util/RequestParser.php#L50
Solution referenced in comments : https://stackoverflow.com/questions/68651/get-php-to-stop-replacing-characters-in-get-or-post-arrays/18209799#18209799

mnapoli:
@Cafeine42 nice, that sounds interesting, especially knowing it's already used by API Platform at scale. And it looks "simple" too.

A PR with that implementation (with correct `@author` attribution) could be easier to merge probably.

alekitto:
IMHO the "api platform" solution sounds hackish, is based on the same function that caused the original bug and is not really performant (~50% slower in my tests).

Could be "simpler" to see, but in an environment where every millisecond counts and it's billed, performance should be taken into consideration.

mnapoli:
Thanks for all the effort! This has been fixed in #1437

mnapoli:
FYI I'm considering changes for Bref v4: https://github.com/brefphp/bref/issues/2159#issuecomment-5439094957

PR #1048, opened 2021-09-30
Avoid legacy global endpoint for S3 bucket

buddhaCode:
The global S3 endpoint is called [legacy](https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html#VirtualHostingBackwardsCompatibility) and creates a redirection response to the region endpoint URL. Using the region endpoint directly, is cheaper and saves one request round trip.

### DIFF
--- a/docs/frameworks/laravel.md
+++ b/docs/frameworks/laravel.md
@@ -100,7 +100,7 @@ aws s3 sync public/ s3://<bucket-name>/ --delete --exclude index.php
 Then, the assets need to be included from S3. In the production `.env` file you can now set that variable:
 
 ```dotenv
-MIX_ASSET_URL=https://<bucket-name>.s3.amazonaws.com
+MIX_ASSET_URL=https://<bucket-name>.s3.<region>.amazonaws.com
 ```
 
 ### Assets in templates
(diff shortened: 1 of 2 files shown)


deleugpn:
Thanks!

mnapoli:
I'm afraid we may need to revert that, as regional endpoints are not consistent across regions:

image

(https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html#VirtualHostingBackwardsCompatibility)

I think that was the reason for using the global endpoints, but maybe I'm confused.

dytyniuk:
@mnapoli IMHO, you don't need to, because according to the same documentation:
Знімок екрана 2021-10-01 о 14 49 05

buddhaCode:
I guess the dash (-) notation for some regions is optional. But every region should support the `<bucket>.s3.<region>` style url.

mnapoli:
> I guess the dash (-) notation for some regions is optional. But every region should support the <bucket>.s3.<region> style URL.

I've been bitten by this in the past, but after digging it seems that was the case for the website endpoints: https://docs.aws.amazon.com/AmazonS3/latest/userguide/WebsiteEndpoints.html

image

So it seems that non-website endpoints may be different and hopefully always support "dot-notation" regional endpoints.

=> it seems this PR is fine 🎉

buddhaCode:
> I've been bitten by this in the past, but after digging it seems that was the case for the website endpoints: https://docs.aws.amazon.com/AmazonS3/latest/userguide/WebsiteEndpoints.html

I see. I never understood why there is a separate website endpoint. But I am quite new to the AWS world of magic ;)

> => it seems this PR is fine 🎉

Whoop Whoop :)

PR #1049, opened 2021-09-30
Fix trusted proxies configuration

RikudouSage:
As per the [documentation](https://symfony.com/doc/current/deployment/proxies.html#but-what-if-the-ip-of-my-reverse-proxy-changes-constantly).

### DIFF
--- a/docs/frameworks/symfony.md
+++ b/docs/frameworks/symfony.md
@@ -116,7 +116,7 @@ You should add the following lines to `config/packages/framework.yaml`
 
 framework:
   # trust the remote address because API Gateway has no fixed IP or CIDR range that we can target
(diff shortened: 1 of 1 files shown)


t-richard:
Did you have an issue with the config documented in Bref?

If I remember correctly `REMOTE_ADDR` is 127.0.0.1 in lambda so it would be redundant.

t-richard:
See https://github.com/brefphp/bref/blob/312c1a1a80a7fd709170472eb99586b777cf87de/src/Event/Http/FpmHandler.php#L184

RikudouSage:
Sorry, was just following the official Symfony documentation and thought it was just a mistake in Bref's documentation.

PR #1051, opened 2021-10-02
[RUNTIME] Docker images with ARM

t-richard:
Make bref layers compatible with ARM architecture for Lambda

### DIFF
--- a/runtime/base/base.Dockerfile
+++ b/runtime/base/base.Dockerfile
@@ -85,6 +85,7 @@ RUN mkdir -p ${BUILD_DIR}  \
 #   - xml2
 ENV VERSION_ZLIB=1.2.11
 ENV ZLIB_BUILD_DIR=${BUILD_DIR}/xml2
+ENV CHOST=arm
 
 RUN set -xe; \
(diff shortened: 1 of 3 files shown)


runtime/base/base.Dockerfile
```diff
@@ -85,6 +85,7 @@ RUN mkdir -p ${BUILD_DIR}  \
 #   - xml2
 ENV VERSION_ZLIB=1.2.11
 ENV ZLIB_BUILD_DIR=${BUILD_DIR}/xml2
+ENV CHOST=arm
```
t-richard, line 88: @deleugpn here's the important part for zlib

t-richard:
Don't take into account some changes, I just wanted to be able to run github action in my own fork and only for php 8.0 with ARM.

deleugpn:
I wonder if now it's the time to swap GitHub Actions with CodeBuild for the Image Builds. The biggest advantage to me would be using an atual ARM64 instead of build emulation

t-richard:
Yes, it's super slow on github actions. But it would be nice for contributors to know if a pipeline succeeded/failed when they make a PR that modifies a runtime.

PR #1054, opened 2021-10-04
Mention DynamoDB in storage description

RikudouSage:
DynamoDB is great for serverless apps because it doesn't incur costs when not in use. I also added link to my package (which was suggested to me by @mnapoli on Reddit about the Symfony bundle) but if you don't like it there I can remove it. I would still leave the part about DynamoDB because AFAIK it's the only cache solution that doesn't incur costs when not in use.

### DIFF
--- a/docs/environment/storage.md
+++ b/docs/environment/storage.md
@@ -42,7 +42,7 @@ The following solutions can be used instead:
 
 - pre-generate the cache in the project directory before deploying
 - store the cache into the `/tmp` directory
-- store the cache into a distributed cache service like Memcache or Redis
+- store the cache into a distributed cache service like Memcache, Redis or DynamoDB (an AWS service which fits the pay-per-request model nicely)
 
 ### Pre-generating the cache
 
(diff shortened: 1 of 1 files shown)


mnapoli (approved):
I think this is an excellent change, thank you!

DynamoDB should indeed be a default choice for caching: even though it's not as fast as the other, it's much easier to get started with (no instance to deal with).

RikudouSage:
If you want, I can add some example of deploying a DynamoDB table using the serverless file to the documentation.

mnapoli:
That would be great indeed!

t-richard:
Yes that would indeed be nice :slightly_smiling_face: I had to spend a bit of time when I used your Symfony Bundle.

While you're here, I have another question for you. Do you have any plan to use Async AWS instead of the default AWS SDK. 

When running in Lambda, size matters and the official SDK weighs 20MB (8% of Lambda max package size) while Async AWS would be less than 1MB. 

There are issues open for this https://github.com/async-aws/aws/issues/998 and https://github.com/symfony/symfony/issues/41498 which mentionned your package and this caveat of using the heavy AWS SDK.

Would love to hear your thoughts on this.

RikudouSage:
Oh, what were the trouble you had when deploying? Could you maybe open an issue so I can clarify anything that's hard to understand for newcomers? I've worked with CloudFormation so much that I probably miss some things that seem obvious to me but not necessarily to others.

I wasn't aware of the problems or the existence of Async AWS package, it definitely seems worth it.

t-richard:
> Oh, what were the trouble you had when deploying? [...] I've worked with CloudFormation so much that I probably miss some things that seem obvious to me but not necessarily to others.

Nothing specific, just getting the cloudformation template right, that's why it seems like a good idea to me.

i'm also very familiar with Cloudformation but not with DynamoDB at that moment so this was a bit of trial & error to get it to work.

here is the resource I came up with:

```yaml
        CacheTable:
            Type: 'AWS::DynamoDB::Table'
            DeletionPolicy: Retain
            Properties:
                AttributeDefinitions:
                    - AttributeName: id
                      AttributeType: S
                KeySchema:
                    - AttributeName: id
                      KeyType: HASH
                BillingMode: PAY_PER_REQUEST
                TableName: ${opt:stage, self:provider.stage}_cache
                TimeToLiveSpecification: # this will delete cache/session keys automatically based on the expiry timestamp
                    AttributeName: ttl
                    Enabled: true
```

Not sure it's optimal but it does work well.

> I wasn't aware of the problems or the existence of Async AWS package, it definitely seems worth it.

Lovely!

I started to implement it on your package some time ago but totally forgot about it and wasn't sure at that time if it was better to make the change there or in AsyncAWS because they already have a [DynamoDB session bridge there](https://async-aws.com/integration/sessions.html) or even in Symfony itself as an official Cache adapter ([Laravel has such native adpter](https://laravel.com/docs/8.x/cache#dynamodb)). 

If I remember correctly, the hard part was about exceptions but can't quite rember why, I'll see if i can find my changes and see what was wrong.

Let me know what you think.

PR #1055, opened 2021-10-05
Add example of deploying DynamoDB table

RikudouSage:
As per discussion in https://github.com/brefphp/bref/pull/1054

### DIFF
--- a/docs/environment/storage.md
+++ b/docs/environment/storage.md
@@ -76,3 +76,110 @@ Note that Redis and Memcache through ElastiCache run even when not used and thus
 pay-per-request mode where you don't pay when the store is not used. There is a package implementing the PSR cache interfaces using DynamoDB ([rikudou/psr6-dynamo-db](https://github.com/RikudouSage/DynamoDbCachePsr6)).
 
 This solution is ideal for cache data that can change during the life of the application (e.g. caching a website menu, an API response).
+
+#### Deploy DynamoDB table as part of serverless
+
+The example will be based on the minimal configuration which is as follows:
+
+```yaml
+service: app
+provider:
+    name: aws
+    runtime: provided.al2
+plugins:
+    - ./vendor/bref/bref
+functions:
+    app:
+        handler: index.php
+        layers:
+            - ${bref:layer.php-74-fpm}
+        events:
+            - httpApi: '*'
+```
+
+As any service, the DynamoDB table can be deployed using the `resources.Resources` subkey:
+
+```yaml
+service: app
+provider:
+    name: aws
+    runtime: provided.al2
+plugins:
+    - ./vendor/bref/bref
+functions:
+    app:
+        handler: index.php
+        layers:
+            - ${bref:layer.php-74-fpm}
+        events:
+            - httpApi: '*'
+resources:
+    Resources:
+        CacheTable:
+            Type: AWS::DynamoDB::Table
+            Properties:
+              AttributeDefinitions:
+                - AttributeName: id
+                  AttributeType: S
+              BillingMode: PAY_PER_REQUEST
+              TimeToLiveSpecification:
+                AttributeName: ttl
+                Enabled: true
+              KeySchema:
+                - AttributeName: id
+                  KeyType: HASH
+```
+
+Explanation of the properties:
+
+- [AttributeDefinitions](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-dynamodb-table.html#cfn-dynamodb-table-attributedef) contains the list of mandatory attributes and must contain a partition key (similar to primary key in databases), which is in this example simply called `id`. The [AttributeType](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-dynamodb-attributedef.html#cfn-dynamodb-attributedef-attributename-attributetype) `S` means it's a string. Note that no other properties except the keys need to be specified as DynamoDB is a NoSQL database without a fixed schema.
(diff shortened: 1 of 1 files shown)


docs/environment/storage.md
```diff
@@ -76,3 +76,110 @@ Note that Redis and Memcache through ElastiCache run even when not used and thus
...
+    runtime: provided.al2
+plugins:
+    - ./vendor/bref/bref
+functions:
+    app:
+        handler: index.php
+        layers:
+            - ${bref:layer.php-74-fpm}
+        events:
+            - httpApi: '*'
```
mnapoli, line 115: Thanks! I think we can simplify the documentation by skipping the "standard" config (that will also be less confusing to integrate in Laravel or Symfony apps where serverless.yml may be a bit different):

```suggestion
service: app
...
```

docs/environment/storage.md
```diff
@@ -76,3 +76,110 @@ Note that Redis and Memcache through ElastiCache run even when not used and thus
...
+```
```
mnapoli, line 98: I think the section above can be removed then.

docs/environment/storage.md
```diff
@@ -76,3 +76,110 @@ Note that Redis and Memcache through ElastiCache run even when not used and thus
...
+              AttributeDefinitions:
+                - AttributeName: id
+                  AttributeType: S
+              BillingMode: PAY_PER_REQUEST
+              TimeToLiveSpecification:
+                AttributeName: ttl
+                Enabled: true
+              KeySchema:
+                - AttributeName: id
+                  KeyType: HASH
```
mnapoli, line 184: We can get rid of this duplicated part

```suggestion
...
```

docs/environment/storage.md
```diff
@@ -76,3 +76,110 @@ Note that Redis and Memcache through ElastiCache run even when not used and thus
...
+- [BillingMode](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-dynamodb-table.html#cfn-dynamodb-table-billingmode) needs to be set to PAY_PER_REQUEST, otherwise you will incur costs even when no requests are arriving.
+- [TimeToLiveSpecification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-dynamodb-table.html#cfn-dynamodb-table-timetolivespecification) is optional and allows you to set a field that will act as a time to live attribute, it's very useful for caching purposes.
+- [KeySchema](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-dynamodb-table.html#cfn-dynamodb-table-keyschema) is a definition of the type of the primary/sort keys (not a datatype which was already defined in AttributeDefinitions). Possible values for [KeyType](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-dynamodb-keyschema.html#aws-properties-dynamodb-keyschema-keytype) are `HASH` (partition/primary key) and `RANGE` (sort/secondary key).
+
+You need to add a permission to access the DynamoDB from your function using `provider.iamRoleStatements`.
```
mnapoli, line 140: Instead of `provider.iamRoleStatements` it will be better to use the new format `provider.iam.role.statements` (see https://www.serverless.com/framework/docs/providers/aws/guide/iam/)

mnapoli:
Thank you, looks really good. I think we can simplify this, I added comments inline.

docs/environment/storage.md
```diff
@@ -76,3 +76,110 @@ Note that Redis and Memcache through ElastiCache run even when not used and thus
...
+              BillingMode: PAY_PER_REQUEST
+              TimeToLiveSpecification:
+                AttributeName: ttl
+                Enabled: true
+              KeySchema:
+                - AttributeName: id
+                  KeyType: HASH
+```
+
+Explanation of the properties:
```
t-richard, line 133: Not sure we need this much information here ? Maybe a curated list of  link to AWS docs would be better to keep this documentation simple ?

t-richard (commented):
One small comment

docs/environment/storage.md
mnapoli: Honestly the rendered version isn't that huge:

image

Ideally I'd like to shorten the sentences a bit, but I still think this is a net positive because that config is so cryptic (for newcomers) that it helps to have a bit of context. Even if it's a bit too much.

t-richard: Right, the links made it look like it was huge and it could indeed help newcomers :+1: 

At various places in bref docs, descriptions like this are inlined in the Cloudformation templates as comments which brings them closer to what it's describing. because this will likely be copied as-is in user projects, it will also be documented directly in their serverless.yml which can be handy for them.

And because we don't want whole paragraphs in YAML comments, it also helps keeping things small and simple. 

Then we can add links for further details. 

Something like this (written quickly, just to get an idea of what that could look like)

```yaml
        CacheTable:
            Type: 'AWS::DynamoDB::Table'
            Properties:
                AttributeDefinitions:
                    - AttributeName: id # adds a mandatory id field
                      AttributeType: S # the type of id will be a string
                KeySchema:
                    - AttributeName: id
                      KeyType: HASH # id will be the partition key (similar to primary keys in SQL)
                BillingMode: PAY_PER_REQUEST # billed for each request instead of paying for a fixed amount of operations
                TimeToLiveSpecification: # deletes cache keys automatically based on the ttl field which contains a timestamp
                    AttributeName: ttl
                    Enabled: true
```

mnapoli: That indeed looks interesting! @RikudouSage see what you can do with those suggestions?

RikudouSage: I'll look into it all in an hour or so

RikudouSage:
@mnapoli Updated, please let me know if it's ok.

t-richard (approved):
LGTM, thanks @RikudouSage 👍

mnapoli:
Thank you!

PR #1065, opened 2021-10-22
Add Bahrain, Hong Kong, and Osaka regions

sgrodzicki:
Things have changed since #681 and Osaka is now a standard AWS region:

https://aws.amazon.com/blogs/aws/aws-asia-pacific-osaka-region-now-open-to-all-with-three-azs-more-services/

Bahrain and Hong Kong need to be enabled manually on your account (opt-in required).

### DIFF
--- a/runtime/layers/regions.json
+++ b/runtime/layers/regions.json
@@ -10,11 +10,14 @@
     "us-east-2",
     "us-west-1",
     "us-west-2",
+    "ap-east-1",
     "ap-south-1",
     "ap-northeast-1",
     "ap-northeast-2",
+    "ap-northeast-3",
     "ap-southeast-1",
     "ap-southeast-2",
     "eu-south-1",
-    "af-south-1"
+    "af-south-1",
+    "me-south-1"
 ]


samuel881908:
Hello, I am from Hong Kong and my team are going to launch a service in Hong Kong. We would want to use bref to launch our code that built with Laravel. It would be great if ap-east-1 can be one of the support region. Thanks!

mnapoli:
Hey @samuel881908, thanks for chiming in! Let's merge!

samuel881908:
I would also want to ask if it is possible to provide new runtime layers of the new regions? I have tried to follow the instruction [AWS Lambda Custom Runtime for PHP: A Practical Example](https://aws.amazon.com/tw/blogs/apn/aws-lambda-custom-runtime-for-php-a-practical-example/) to create my own layer but it does not work with reason "Error: Couldn't find valid bootstrap(s): [/var/task/bootstrap /opt/bootstrap]". I have tried different methods to create the layers but still have the same error. Sorry to keep bothering.

t-richard:
> I would also want to ask if it is possible to provide new runtime layers of the new regions?

This is what this PR is about publishing the Bref layers in the mentioned regions (including Hong Kong).

But I think the change did not trigger a new build because the Bref bot didn't opened a PR after the merge to update the `layers.json` (cc @mnapoli @deleugpn)

Or the layer push failed because @sgrodzicki mentioned Hong-Kong and Bahrain have to be enabled to in the account so the Bref account might need to opt-in in those regions to publish layers.

samuel881908:
@t-richard Okay I got it! I will try to wait if there is any updates. Thanks a lot!

mnapoli:
> Or the layer push failed because @sgrodzicki mentioned Hong-Kong and Bahrain have to be enabled to in the account so the Bref account might need to opt-in in those regions to publish layers.

😅

t-richard:
@samuel881908 In case you didn't noticed, this has been released and is available as of Bref version 1.3.4

samuel881908:
It works great now! Thank you!

PR #1092, opened 2021-11-12
Report a custom user agent to AWS Lambda

mnapoli:
Teams at AWS are interested in estimating the usage of custom runtimes, including Bref, Rust (see https://github.com/awslabs/aws-lambda-rust-runtime/pull/363) and others.

While we already have some stats thanks to Bref ping (https://bref.sh/docs/runtimes/#bref-ping), which I'll detail below, that approach will allow AWS teams to directly get the info, and get that consistently across multiple runtimes.

The technique is to report a custom User-Agent to the Lambda Runtime API. It has been suggested the following pattern:

```
bref/<PHP version>-<Bref version>
```

However currently in the code we don't have the Bref version available in a handy way :/ 

Additionally, I think it's also important to report the layer actually used, because there's a huge difference between the FPM runtime and Function runtime ([more details](https://bref.sh/docs/runtimes/#bref-runtimes)).

Here is what this PR reports instead:

```
bref/<runtime name>/<php-major-minor-version>
```

`<runtime name>` could be `function`, `fpm` or `console` (again, see https://bref.sh/docs/runtimes/#bref-runtimes).

Here are some examples of values:

```
bref/function/8.0
bref/fpm/8.0
bref/function/7.4
bref/fpm/7.4
...
```

## Recap of current Bref stats

Total monthly invocations using Bref runtimes:

image

We have passed 4 billion invocations/month.

(the last month is not complete)

### DIFF
--- a/src/Runtime/LambdaRuntime.php
+++ b/src/Runtime/LambdaRuntime.php
@@ -129,6 +129,8 @@ final class LambdaRuntime
             $this->handler = curl_init("http://{$this->apiUrl}/2018-06-01/runtime/invocation/next");
             curl_setopt($this->handler, CURLOPT_FOLLOWLOCATION, true);
             curl_setopt($this->handler, CURLOPT_FAILONERROR, true);
+            // Set a custom user agent so that AWS can estimate Bref usage in custom runtimes
+            curl_setopt($this->handler, CURLOPT_USERAGENT, "bref/{$this->layer}/");
         }
 
         // Retrieve invocation ID
--- a/src/Runtime/LambdaRuntime.php
+++ b/src/Runtime/LambdaRuntime.php
@@ -130,7 +130,8 @@ final class LambdaRuntime
             curl_setopt($this->handler, CURLOPT_FOLLOWLOCATION, true);
             curl_setopt($this->handler, CURLOPT_FAILONERROR, true);
             // Set a custom user agent so that AWS can estimate Bref usage in custom runtimes
-            curl_setopt($this->handler, CURLOPT_USERAGENT, "bref/{$this->layer}/");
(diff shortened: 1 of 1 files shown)


t-richard:
Nice to see some interest from AWS again :slightly_smiling_face: 

The only thing I'm thinking about is that this change will only be available for people updating bref so the numbers might be off by a fair amount which might in turn biais numbers for a while and the bref ping could still be used to put numbers in perspective. 

According to packagist, there are still a fair amount of downloads on the `0.x` version of Bref. On last week, only about 20% of downloads were on Bref latest (1.3.4).

https://packagist.org/packages/bref/bref/stats#major/all

Maybe now is the moment to do it because of the approaching release of PHP 8.1 which might drive updates up for a while. 

But I'm :+1: if it can help.

PS: 4 billion, that's nice :fire:

mnapoli:
> so the numbers might be off by a fair amount

Yes, that's definitely true.

Even the bref ping stats are underrepresenting the reality: bref ping can be disabled, and not everyone has updated to the Bref version that reports it anyway. Bref 0.5 is 15% of users (on Composer/Packagist), these applications are not tracked at all for example.

But since we can't go back in time, now the second best time to do it I guess. It may also help AWS compare adoption of custom runtimes between each other? E.g. is Rust more used than PHP or not (where to put the attention).

nmoutschen:
Thanks a lot for opening this PR @mnapoli ! 🙌 

So if I understand correctly, this would report something like `bref/php-80-fpm/`, or `bref/php-74/`? If that's the case, that looks good to me!

mnapoli:
Not exactly, but I actually realized there's a simple way we can track the layer name + PHP version (I was too focused on the layer name to see that). Here is an example of what would be reported:

```
bref/function/8.0
bref/fpm/8.0
bref/function/7.4
bref/fpm/7.4
...
```

I updated the PR description.

PR #1105, opened 2021-11-29
Support Symfony 6.0

t-richard:
Symfony 6.0 was just released.

This PR brings support for the 6.x range of Symfony components

One major BC break for Symfony is the addition of types for arguments and return. 

Let's hope there's no impact :crossed_fingers:


mnapoli:
Thank you!

PR #1113, opened 2021-12-06
Support SQS partial batch failure

t-richard:
Allows to support the recent addition for partial batch failures with SQS.

The syntax looks like this

image

Todo:
- [x] Feature
- [x] Deploy & test
- [x] Automated Tests
- [x] Documentation


mnapoli:
That looks perfect, thank you for contributing that!

Regarding documentation, I guess it should go in this section: https://bref.sh/docs/function/handlers.html#sqs-events

I was thinking that eventually it could be interesting to have a dedicated `Async workers` page in the documentation about SQS, because right now it's a bit hidden in "Typed handlers" (and it's a very common use case for PHP apps). But I'm just sharing that to gauge the idea, this isn't something we need to do right now.

docs/function/handlers.md
```diff
@@ -137,6 +137,24 @@ class Handler extends SqsHandler
 return new Handler();
 ```
 
+### Partial Batch Response
+
+While handling a batch of records, you can mark it as partially successful to reprocess only the failed records.
```
mnapoli, line 142: That will only work if users have enabled partial batch responses on their apps though. With Lift [eventually](https://github.com/getlift/lift/pull/145) that will be automatic, but I think here we should mention that at least?

t-richard: Yeah, this is a first proposal made by @afu-dev (thanks by the way).

I agree we should add information for the config required to use this and maybe also mention that when using Lift this is enabled by default once the related PR is merged.

I'll try to get everything ready next week.

mnapoli: 👍  no problem, sounds great!

afu-dev: Good catch, I didn't know I had to declare a `functionResponseType` in the serverless file.

I'll send @t-richard another PR

mnapoli (approved):
LGTM, will probably merge and release tomorrow if nobody else has other comments.

Thanks!

mnapoli:
Tomorrow = a week later in open-source world of course 😄 

Thank you @t-richard!

PR #1114, opened 2021-12-07
Re-enable xdebug in the PHP 8.1 dev container following #980 and #1007

mnapoli:
This PR re-enables xdebug and blackfire for PHP 8.1.

These extensions were skipped in #980 and #1007 because they were not ready for PHP 8.1. They now are.

I've reverted the Dockerfile content to an older version: https://github.com/brefphp/bref/blob/126ccf9b90c6a394a35f8551ec69688678786f49/runtime/layers/fpm-dev/Dockerfile


mnapoli:
The Docker build fails because of rate limiting… thanks Docker.

danieloraca (approved):

PR #1117, opened 2021-12-09
Add standard Symfony commands

umulmrum:
This PR replaces Silly commands with default Symfony commands. As both Bref and Silly are maintained by @mnapoli, they might not be amused, so let me explain my thoughts:

Silly indeed removes some boilerplate code required by Symfony, turning command definitions significantly easier. So there is definitely benefit in using it, but I think it doesn't justify the related cost.

- Silly doesn't improve UX or DX directly, but is just an internal tool to simplify Bref development slightly.
- Silly hasn't evolved in more than 1,5 years. It isn't the maintainers fault if no one helps out, but still, the result is an outdated lib that needs quite some work to be done. My personal conclusion was that the required time could be invested better somewhere else. Maybe everyone else thinks alike, which would mean Bref is hindered by a lib that is farther down on everyone's list.
-  Removing the dependency from Bref would not only allow complete upgrades to Symfony 6, psr/container 2, and psr/log 3 instantly, but also lower the complexity/file sizes which isn't that bad in Lambda environments (not that much, but still - php-di/invoker as well as polyfills are also removed).

As a final point, I'd say that the new command classes in this PR make it easier to write tests for the commands (and I think separate classes are way easier to understand, but this might be a matter of opinion), but this could also be reached by simple refactoring.

All that said, I'm relatively new to Bref and might have overlooked significant things. Also, the changes are mostly untested yet and I'm sure that arguments/options aren't correct yet. I'm happy to complete the PR if you are :+1: in general. Thanks!


demo/console.php
```diff
@@ -1,27 +1,50 @@
...
-    $phpinfo = ob_get_clean();
-    $output->write($phpinfo);
-});
-$silly->command('error', function (OutputInterface $output) {
-    $output->writeln('There was an error!');
-    return 1;
-});
-$silly->command('sleep', function () {
-    sleep(120);
+$app->add(new class('hello') extends Command {
```
GromNaN, line 13: The `register` method should be used for this case.

```php
$app->register('hello')
    ->addArgument('name', InputArgument::OPTIONAL, '', 'World!')
    ->setCode(function (InputInterface $input, OutputInterface $output): int {
        $output->writeln('Hello, ' . $input->getArgument('name'));
        
        return Command::SUCCESS;
    });
```

GromNaN (approved):
I don't know if silly supports (proxy) autocompletion, but this is a feature that we can support by using the standard Symfony Console 5.4+.

mnapoli:
I do appreciate the effort of thinking out of the box to address the problem, instead of fixing all the dependencies upstream :D 

The PR looks clean, have you had a chance to test all (most?) commands? Is this easy to merge, or are manual tests required?

mnapoli:
Ohh tests on master fail, probably again because https://github.com/shivammathur/setup-php changed the way PHP-FPM installs again. I think we need to revert some parts of https://github.com/brefphp/bref/pull/991 I can't address this right now (I have a live stream tonight), maybe tomorrow if someone else doesn't get to this first.

composer.json
```diff
@@ -22,7 +22,7 @@
         "php": ">=7.3.0",
         "ext-curl": "*",
         "ext-json": "*",
-        "mnapoli/silly": "^1.7",
+        "symfony/console": "^3.0|^4.0|^5.0|^6.0",
```
GromNaN, line 25: Is it necessary to support old and unmaintained versions? `^4.4|^5.0|^6.0` should be enough.

src/Console/Command/Local.php
```diff
@@ -8,19 +8,41 @@
...
 
 /**
  * Local function invocation.
  */
-class Local
+class Local extends Command
 {
+    protected static $defaultName = 'local';
+
     public const SIGNATURE = 'local [function] [data] [--file=] [--handler=] [--config=]';
```
GromNaN, line 26: This const is not used.

demo/console.php
umulmrum: Way better, thanks!

composer.json
umulmrum: I just copied the requirement from Silly, and then added `^6.0`. I'd suggest to keep support for the old releases for now, and to think about it together with the other Symfony requirements in another issue/PR.

src/Console/Command/Local.php
umulmrum: Strictly speaking this is a BC break but should be ok anyway.

GromNaN: Removing `__invoke` from this class is a more risked breaking change. I don't know what's the policy here.

mnapoli: Feel free to remove, no BC break guarantee for console commands.

t-richard:
AFAIK defining the static property is only useful when you want to be able to lazy-load commands if using the Symfony **framework** (see https://symfony.com/doc/current/console/commands_as_services.html#lazy-loading)

I think it's OK to switch to the "old-school way" as Symfony 3 is not a huge obstacle here apparently.

mnapoli:
Let's give this a try, thank you @umulmrum!

PR #1118, opened 2021-12-09
Fix FPM failure in CI

t-richard:


t-richard:
The CI is all green, like the AWS status page during an outage :smirk:

PR #1125, opened 2021-12-16
Upgraded to openssl `1.1.1m`

GrahamCampbell:


deleugpn:
Funny how this looks like a downgrade 😄

deleugpn:
I'm thinking we can wait the PHP upgrade to release new layers

GrahamCampbell:
I agree. I understand this openssl release only contains performance fixes and not security fixes. :)

PR #1144, opened 2022-01-27
Restart FPM completely in case of timeouts

mnapoli:
Follow-up of #1133

This is because sending SIGUSR2 to FPM (the previously implemented solution) did not really stop with 100% certainty the PHP script that timed out.

Indeed, it merely interrupted the currently blocked call (e.g. a sleep, a DB call, etc.), flushed the logs and carried on. My guess is that this could have caused the PHP script to continue to run in some cases, possibly running into yet another timeout on a next line (e.g. another DB call).

This PR fixes the timeout test that wasn't really working (🤦) and restarts FPM completely in case of timeout. That is confirmed to completely stop the execution of the timed out script + flush the logs to stderr.

### DIFF
--- a/src/Event/Http/FpmHandler.php
+++ b/src/Event/Http/FpmHandler.php
@@ -93,7 +93,11 @@ final class FpmHandler extends HttpHandler
     public function stop(): void
     {
         if ($this->fpm && $this->fpm->isRunning()) {
-            $this->fpm->stop(2);
+            // Give it less than a second to stop (500ms should be plenty enough time)
+            // this is for the case where the script timed out: we reserve 1 second before the end
+            // of the Lambda timeout, so we must kill everything and restart FPM in 1 second.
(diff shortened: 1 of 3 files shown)


mnapoli:
This is what CloudWatch logs look like now when you have 1 good request followed by 1 timed-out request:

Screen 2022-01-27 16-44

shadowhand (approved):
Excellent! And good sleuthing to find that it attempts to resume after interrupt!

BenCoffeed:
Thank you

t-richard (approved):

GrahamCampbell (approved):
Noice.

shadowhand:
@mnapoli can we get a release with this? We're currently trying to upgrade to serverless v3 but stuck because this hasn't been released.

mnapoli:
@shadowhand you will have to wait until I do, maybe today.

shadowhand:
@mnapoli thank you! 💛

PR #1145, opened 2022-01-28
Fix anchor link in error message

j0k3r:
I fixed the anchor link in the error message when the response from the lambda is in binary mode when that mode isn't enabled.

I got the error when I enabled a gzipped response from Slim and forgot to enable that binary feature.


deleugpn (approved):
Thanks!

PR #1156, opened 2022-02-02
Add missing "type" in lift `constructs`

stloyd:
Otherwise erros popups:
```
Error:
The construct 'website' has no 'type' defined.
```

### DIFF
--- a/docs/websites.md
+++ b/docs/websites.md
@@ -90,7 +90,8 @@ Now add your domain name and certificate in your `serverless.yml` file
 ...
 constructs:
   website:
-    # ...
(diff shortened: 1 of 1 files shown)


mnapoli:
Thanks for the PR!

However it seems the `# ...` was intentional because the whole detailed configuration is a bit more complex than just `type`, and it is detailed just a few lines above the line that you modified:

image

So the section you changed is specifically about setting up a custom domain (setting up 2 extra lines in the config), that's why the full config (including `type`) is intentionally skipped, to avoid duplicating and adding extra noise.

Does that make sense?

PR #1173, opened 2022-03-08
Upgraded to libcurl 7.82.0

GrahamCampbell:


GrahamCampbell:
Yeh. Various bug fixes. It makes sense to always track the latest version, so that when a security issue does come around, we're not in a mess because we're miles behind and have to deal with broken builds when all we want is to get a security fix in.

Issue #965, opened 2021-07-06
SCRAM authentication requires libpq version 10 or above in


NeilJ247:
**Description:**

I am getting the following error when trying to establish a connection with Timeseries (postgres) database in Lambda:

`SCRAM authentication requires libpq version 10 or above in`

My application has to connect to 2 separate databases.  Their versions are:

Timeseries
PostgreSQL 10.15 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 4.8.3 20140911 (Red Hat 4.8.3-9), 64-bit

Non-timeseries
PostgreSQL 12.7 (Debian 12.7-1.pgdg100+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 8.3.0-6) 8.3.0, 64-bit

**How to reproduce:**

Runtime: provided.al2
Layers: bref:layer.php-74 and bref:layer.console
Extensions: pdo_pgsql
Libraries: Doctrine DBAL

I have a local docker image that I build which extends from php:7.4-fpm-alpine which runs the application fine:

```
FROM php:7.4-fpm-alpine

RUN apk update && \
    apk add -q --no-progress $PHPIZE_DEPS postgresql-dev postgresql-client git && \
    yes | pecl install xdebug pcov && \
    docker-php-ext-install pgsql pdo pdo_pgsql && \
    docker-php-ext-enable pcov xdebug
```

t-richard:
@NeilJ247 if you haven't found an answer to your problem, maybe you could give a try to the `pgsql` layer provided at https://github.com/brefphp/extra-php-extensions

Not sure it would fix the issue but it's worth a try.

NeilJ247:
@t-richard yeah I tried that but it's the same issue.

BordelD:
Same here

NeilJ247:
Corrected the original description.  I had the Postgres versions the wrong way around for the databases.

Our Timeseries DB is now upgraded to 12.9 and I have upgraded to PHP 8 but the issue still persists.

NeilJ247:
Which version of libpq is installed by default on bref?

Nyholm:
Bref is currently running 9.6.24 which was released in November 2021. https://github.com/postgres/postgres/releases/tag/REL9_6_24

I also got this issue when I tried to connect to a new Postgres 14 database.

Nyholm:
[...]

PR #1188, opened 2022-04-10
Upgrade Postgres to 14.2

HugoJF:
This seems to fix https://github.com/brefphp/bref/issues/965.

I have no clue what are the implications of this upgrade, but it seems to work fine for my use-case.

### DIFF
--- a/runtime/base/base.Dockerfile
+++ b/runtime/base/base.Dockerfile
@@ -83,7 +83,7 @@ RUN mkdir -p ${BUILD_DIR}  \
 #   - php
 # Used By:
 #   - xml2
-ENV VERSION_ZLIB=1.2.11
+ENV VERSION_ZLIB=1.2.12
 ENV ZLIB_BUILD_DIR=${BUILD_DIR}/xml2
 
 RUN set -xe; \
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
This needs rebasing. :)

mnapoli:
Before merging this it would be great if we knew what the impact was for upgrading the version by so much. I have no way of knowing if this is good or not.

tillkruss:
I've love to see support for v14 instances.

shadowhand:
Postgres is highly backwards compatible. As a Postgres user, I fully support this change.

t-richard:
> Postgres is highly backwards compatible. 

This is my feeling too. 

This has been a recurring issue for a number of month now and may be blocking more users to use Bref with Postgres than it might cause harm.

Maybe we could merge this and tag a beta release. This would give the community actual layers to test it properly and give feedback ?

WDYT @mnapoli @deleugpn ?

shadowhand:
@GrahamCampbell I agree with only making this change in PHP 8.x images.

shadowhand:
To be clear: this is only upgrading the Postgres **client** which has (almost) nothing to do with the **server** or **database** version. 

The database version is always created with the current server version, so if my server is version 11.0 then my database will be created as version 11.0 and the database will need to be upgraded (see `pg_upgrade`) if the server is upgraded. However, the **client** protocol has not changed since version 7.4, so any version of the client will be able to communicate with any server >= 7.4.

Upgrading the client version is **very low risk** and, if anything, should improve functionality for any Postgres connection.

allan-simon:
@mnapoli I've tested locally with version 9 to 14 of postgres on our application and I can confirm this PR works 

having it merged would allow us to upgrade in production our version of Postgresql , as the compatibility with bref is the only thing left

allan-simon:
@mnapoli  thanks for the merge,  is it possible to have a version tagged :angel:  ?

Nyholm:
Before we tag it, could someone with postgres < 12 confirm that this works? Or am I too afraid/careful?

allan-simon:
> I've tested locally with version 9 to 14 of postgres on our application and I can confirm this PR works

already done :)

mnapoli:
Thanks, I'll tag a release then.

Nyholm:
Thank you @allan-simon <3

PR #1193, opened 2022-04-19
Add Laravel event handler documentation

robmellett:
Added some documentation to help Laravel users understand to use Artisan commands with custom Lambda handlers.

### DIFF
--- a/docs/frameworks/laravel.md
+++ b/docs/frameworks/laravel.md
@@ -360,3 +360,48 @@ Instead, here is what you need to do:
    ```
 
 All these steps were replacements of running the `passport:install` command [from the Passport documentation](https://laravel.com/docs/passport#installation).
+
+## Lambda Handler Example
+
+If you need to handle custom Lambda events, you can can wire up a handler, and an artisan command in the following way.
+
+```yml
+# serverless.yml
+
+functions:
+  s3-events:
+    handler: serverless-handler.php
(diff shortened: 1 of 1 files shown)


docs/frameworks/laravel.md
```diff
@@ -360,3 +360,48 @@ Instead, here is what you need to do:
...
+return function ($event, Context $context) {
+    static $app;
+
+    if (null === $app) {
+        $app = require __DIR__.'/bootstrap/app.php';
+    }
+
+    return $app->make(Kernel::class)->call('generate:gif', [
+        'payload' => $event
+    ]);
```
mnapoli, line 403: Oh, the best practice is not to invoke an artisan command, but to return an actual Lambda handler (which is a PHP class).

Here what you document is a function Lambda handler. What's in the handler is calling an artisan command, but it could do anything. I'm not sure if I'm clear enough, sorry 😕 

https://github.com/brefphp/laravel-bridge/blob/master/worker.php#L18-L21 is an example of that: it doesn't return a function handler, it returns a class handler.

robmellett: Yeap that makes sense. 

Give me a few days to test it, and I'll update the PR.

Thanks for the review!

mnapoli:
Closing this PR for now to clean up the PR queue. Let me know if we need to reopen this later!

PR #1210, opened 2022-05-20
close #1187: we log the AWS request id when timeout-ing

### LINKED ISSUE #1187: Log the  lambda request id in the timeout exception
So that it's easier to map this log to other log in tools like kibana /grafana 

Implementation:  make the  TimeOut exception take a second parameter (optional  so that it works in the local docker environment ?) which is the lambda request id coming from the context 

If you think it is a good idea, I can provide the PR

allan-simon:
so that it's easier when using tools like graphana/kibana to find all the
logs related to the same request that ultimately failed


mnapoli:
Looks great thank you!

PR #1215, opened 2022-06-06
Add Laravel Octane Handler

datpmwork:
This pull request is to support bootstraping the server by Laravel Octane

### Why?
- Since I started to use Laravel Octane for my project, I don't know how to setup with current bref

### How does it work?
- Inspired by current FpmHandler
- Instead of using FastCgi Client, OctaneHandler will use Guzzle to forward the HttpEvent to Laravel Octane server

#### New Layers
- [x] `php-80-roadrunner`
- [x] `php-80-roadrunner-dev`
- [x] `php-80-swoole`
- [x] `php-80-swoole-dev`
- [x] `php-81-roadrunner`
- [x] `php-81-roadrunner-dev`
- [x] `php-81-swoole`
- [x] `php-81-swoole-dev`
- [x] `bref/octane-dev-gateway`


t-richard:
Hi!

Thanks for opening this PR.

It seems like this should belong in the Laravel bridge and/or the bref/extra-extensions and not in core.

Apart from that I'm really not sure running Swoole or Roadrunner would provide a real benefit in a Lambda environment given the execution model.

There has been some similar work in the Symfony Bridge that doesn't require php extensions, AFAIK Laravel Vapor integration of Octane doesn't require those either and I think @GrahamCampbell did something similar with Bref on his projects.

Let's see what others think 😃

mnapoli:
Laravel Octane support has been released in Bref v2!

https://bref.sh/docs/news/02-bref-2.0.html

PR #1216, opened 2022-06-09
Initial support for Kafka-Events

rejinka:
Like for SNS or SQS, lambda has triggers for Kafka. Having first level support in bref would be helpful for kafka-user.


src/Event/Kafka/KafkaRecord.php
```diff
@@ -0,0 +1,64 @@
...
+        return base64_decode($this->record['value']);
+    }
+
+    public function getHeaders(): array
+    {
+        return array_map(
+            function (array $chars): string {
+                return implode('', array_map(
+                    function (int $char): string {
+                        return chr($char);
```
mnapoli, line 55: Why is this behavior needed?

src/Event/Kafka/KafkaRecord.php
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    public function getTimestamp(): int
+    {
+        return $this->record['timestamp'];
+    }
+
+    public function getValue(): mixed
+    {
+        return base64_decode($this->record['value']);
```
mnapoli, line 46: Is the value always encoded with base64?

mnapoli (commented):
Thanks for contributing!

So far that looks good, I think there are extra empty lines here and there to remove (i.e. fix code formatting), but the CI should help validate all that.

Also could you update the documentation as well in the PR? (https://bref.sh/docs/function/handlers.html)

src/Event/Kafka/KafkaRecord.php
rejinka: I haven't found any documentation stating something else. Also unlike api-gateway events, there isn't a property indicating that the payload is base64 encoded or not. I assume the encoding based on the (one and only) example in the documentation: https://docs.aws.amazon.com/lambda/latest/dg/with-kafka.html

src/Event/Kafka/KafkaRecord.php
mnapoli: Oh interesting, thanks!

Could you add that explanation in a comment in the code? That way it's there forever :p (especially useful in 2 years when we revisit that code and are surprised by it)

src/Event/Kafka/KafkaRecord.php
mnapoli: Oh the docs actually say:

> Each array item contains details of the Kafka topic and Kafka partition identifier, together with a timestamp and a base64-encoded message.

So we're good here, thanks for the link!

mnapoli:
FYI you might want to rebase on (or merge) master, there were fixes for the PHPStan build.

src/Event/Kafka/KafkaRecord.php
rejinka: I added a small explanation and a link to the javadocs. i think that should be enough.

rejinka:
All your feedback is applied. I hope the documentation is sufficient ;-)

docs/function/handlers.md
```diff
@@ -430,3 +430,31 @@ return new Handler();
 ```
 
 [Full reference of Kinesis in `serverless.yml`](https://www.serverless.com/framework/docs/providers/aws/events/streams/).
+
+## KafkaEvent events
```
mnapoli, line 434:
```suggestion
## Kafka events
```

src/Event/Kafka/KafkaEventHandler.php
```diff
@@ -0,0 +1,17 @@
+<?php declare(strict_types=1);
+
+namespace Bref\Event\Kafka;
+
+use Bref\Context\Context;
+use Bref\Event\Handler;
+
+abstract class KafkaEventHandler implements Handler
```
mnapoli, line 8: It should be `KafkaHandler` to match the name of other classes, right?

mnapoli (commented):
Thanks, I added 2 more suggestions after seeing the docs and comparing to the previous sections.

Once these are good, I think we're good to merge 👍

mnapoli (approved):
Looking great, thank you!

I'll wait for the CI build to finish and merge.

src/Event/Kafka/KafkaRecord.php
```diff
@@ -0,0 +1,72 @@
...
+    public function getValue()
+    {
+        return base64_decode($this->record['value']);
+    }
+
+    /**
+     * A header in Kafka is an objects, with a single property. The name of the property is the name of the header. It's value is a
+     * byte-array, representing a string. We'll normalize it to a hashmap with key and value being strings.
+     *
+     * @see https://kafka.apache.org/25/javadoc/org/apache/kafka/common/header/Headers.html
```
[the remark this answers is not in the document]
rejinka: I added the newline, but then the code-style mentioned a wrong group-ordering. So, here we are.

mnapoli:
Thanks!

PR #1219, opened 2022-06-15
Update Laravel documentation to clarify headers to whitelist if using Lift

selfsimilar:
I recently struggled to get Livewire and Telescope for Laravel working. I was encountering [419 errors all over the place](https://github.com/livewire/livewire/discussions/5031). It turns out Cloudfront was not forwarding some necessary HTTP Headers as configured by default via the severless-lift plugin. This update to documentation makes more clear the necessity of explicitly whitelisting certain headers, though it's not exhaustive. I did a quick search in the core Laravel framework for calls to header (`grep header\(`), and at the very least I left out `X-Socket-ID` from [BroadcastManager.php](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Broadcasting/BroadcastManager.php#L129) since there's already a lot of headers listed and the Lift documentation says there's a max of 10 headers you can whitelist. However I can't find any AWS documentation to support that assertion and I am already listing more than ten, seemingly successfully.

### DIFF
--- a/docs/frameworks/laravel.md
+++ b/docs/frameworks/laravel.md
@@ -164,6 +164,26 @@ constructs:
       '/favicon.ico': public/favicon.ico
       '/robots.txt': public/robots.txt
       # add here any file or directory that needs to be served from S3
+    # Laravel uses some headers that are not in CloudFront's default whitelist.
+    # To add any, we need to list all accepted headers to pass through.
+    # https://github.com/getlift/lift/blob/master/docs/server-side-website.md#forwarded-headers
+    forwardedHeaders:
+      - Accept
+      - Accept-Language
+      - Authorization
+      - Content-Type
+      - Origin
+      - Referer
+      - User-Agent
+      - X-Forwarded-Host
+      - X-Requested-With
+      # Laravel Framework Headers
+      - X-Csrf-Token
+      - X-Xsrf-Token
+      # Laravel Telescope
+      - Location
+      # Other Headers, e.g. Livewire
+      - X-Livewire
 ```
 
 Before deploying, compile your assets using Laravel Mix.
--- a/docs/frameworks/laravel.md
+++ b/docs/frameworks/laravel.md
@@ -170,7 +170,6 @@ constructs:
     forwardedHeaders:
       - Accept
       - Accept-Language
-      - Authorization
       - Content-Type
       - Origin
       - Referer
@@ -179,9 +178,6 @@ constructs:
       - X-Requested-With
       # Laravel Framework Headers
       - X-Csrf-Token
-      - X-Xsrf-Token
-      # Laravel Telescope
-      - Location
       # Other Headers, e.g. Livewire
       - X-Livewire
 ```


selfsimilar:
I  know this is closed, but the ten header limit is still in effect ;-/ – it will accept eleven headers if one of them is the `authorization` header as it strips that header before counting.

 I think ultimately putting a CloudFront distribution in front of an API Gateway may not be a recommended configuration for Laravel because of this limitation. I'm currently [messing with the serverless-lift plugin](https://github.com/getlift/lift/issues/229) to see if there's some way to make this work with the `server-side-website` function, but it may just be that having a separate domains for app and assets will be the recommended configuration for anything more than a very simple Laravel deployment.

t-richard:
Indeed, the limit didn't changed and is still 10 by default (apparently it's not a hard limit but never got AWS to increase it :sweat:)

In the current setup, I agree 10 headers is too few but we can't really get around it because of a strange limitation of API Gateway (forwarding the Host header breaks everything).

Having separate assets and web URL is a solution which has it's own limitations (no caching, not globally distributed, requires code/config changes for popuplar frameworks, etc) but can be a very effective way to do it if you need many headers.

Another solution may be to use Lambda URLs behind Cloudfront, I haven't tested yet but maybe this weird Host limitation is gone. It mostly depends on which trade off you're willing to make.

t-richard:
> Another solution may be to use Lambda URLs behind Cloudfront, I haven't tested yet but maybe this weird Host limitation is gone. 

It's not gone... Spent part of my day trying this and it still messes up when sending Host so the situation is the same.

It's like they're willingly making it hard to integrate AWS services together...

mnapoli:
Oh that sucks…

IMO the main problem with not using CloudFront is that we don't get the HTTP -> HTTPS redirect with API Gateway only. We could have CloudFront only for serving the assets (i.e. in front of S3 only) and solve the caching/CDN issue. Sure, assets would be on a separate domain but it's fairly easy to configure that in most frameworks.

eduarguz:
I just reached the limit of 10 forwarded headers :( 

```
      - Accept
      - Content-Type
      - Origin
      - User-Agent
      - X-Forwarded-Host
      - X-Requested-With
      - X-Csrf-Token # From Laravel
      - X-Xsrf-Token # From Laravel
      - X-Inertia # From Inertia
      - X-Inertia-Version # From Inertia
```

Currently using Laravel and Inertiajs there are 4 headers we need to forward, I had to drop `Accept-Language` but im not sure which one is less important
I could say that there will be other headers later like `Authorization` or whatever need we have...

Other option is to use Apigateway CustomDomains right? got that working but lost http to https redirection.

Something in between? hope anyone has ideas. :) 

Awesome work btw, thanks for your work

mnapoli:
@eduarguz ahh that sucks!

OK one solution, in that case, would be:

- example.com -> API Gateway [**REST** API](https://www.serverless.com/framework/docs/providers/aws/events/apigateway) with custom domain: REST APIs are slightly more expensive than HTTP APIs, but with custom domains setup with _edge_ endpoint (not _regional) then you get the HTTP -> HTTPS redirect
- assets.example.com -> S3 assets (possibly with CloudFront in front)

That means 2 separate domains, but at least your app will work correctly.

eduarguz:
Ok I am testing with different options and there is one that magically seems to work 💫 🧙

1. Manually updated the Cloudfront origin request policy to `Headers - All viewer headers`
2. Deployed and the app is not working. (I think thats the problem discussed [here #229](https://github.com/getlift/lift/issues/229))
3. Then added the same domain as an apigateway custom domain using `serverless-domain-manager` plugin
4. Deployed again
5. It works 🎊

**But.. why?** is it ok to have that setup?, is there a way to automate the `All viewer headers` policy setting?

Hope anyone can explain this magic 🧙

Reference `serverless.yml`

```yml
...
plugins:
  - ./vendor/bref/bref
  - serverless-lift
  - serverless-domain-manager
custom:
  customDomain:
    domainName: domain.example.com
    stage: dev
    basePath: ''
    certificateName: 'example.com'
    createRoute53Record: false
    createRoute53IPv6Record: false
    endpointType: 'regional' #not sure
    securityPolicy: tls_1_2
    apiType: http
    autoDomain: true
constructs:
  # Lift construct for app assets
  website:
    type: server-side-website
    domain: domain.example.com
    certificate: arn:aws:acm:us-east-1:AAAA
    assets:
      '/build/*': public/build
      '/favicon.ico': public/favicon.ico
      '/robots.txt': public/robots.txt
```

mnapoli:
Hahaha yes that is a trick that in theory works, but you seem to have validated that it works in practice.

In theory it shouldn't work to forward "All headers" because CloudFront would forward `Host: example.com`, which breaks API Gateway (API Gateway uses the domain name to know which API Gateway to call).

But in your case, since you setup the same custom domain (`example.com`) on your API Gateway, it works!

I'm not sure we want to make this the documented approach though, because it involves a lot of steps (setting up the custom domain twice basically). Also depending on how you set it up, you may have to create HTTPS certificates in multiple regions.

PR #1232, opened 2022-07-02
Add PHP 8.2 Runtimes

dhrrgn:
PHP 8.2 releases have started being rolled out. This adds the necessary runtimes to allow devs to start testing their applications.

### DIFF
--- a/docs/runtimes/README.md
+++ b/docs/runtimes/README.md
@@ -71,17 +71,19 @@ functions:
 
 The `${...}` notation is the [syntax to use variables](https://serverless.com/framework/docs/providers/aws/guide/variables/) in `serverless.yml`. Bref provides a serverless plugin ("`./vendor/bref/bref`") that provides those variables:
 
+- `${bref:layer.php-82}` **experimental**
 - `${bref:layer.php-81}`
 - `${bref:layer.php-80}`
 - `${bref:layer.php-74}`
 - `${bref:layer.php-73}`
+- `${bref:layer.php-82-fpm}` **experimental**
 - `${bref:layer.php-81-fpm}`
 - `${bref:layer.php-80-fpm}`
 - `${bref:layer.php-74-fpm}`
 - `${bref:layer.php-73-fpm}`
 - `${bref:layer.console}`
 
-Bref currently provides runtimes for PHP 7.3, 7.4, 8.0 and 8.1.
+Bref currently provides runtimes for PHP 7.3, 7.4, 8.0 and 8.1. It also provides **experimental** runtimes for PHP 8.2.
 
 > `php-80` means PHP 8.0.\*. It is not possible to require a specific "patch" version.
 
--- a/runtime/base/php-82.Dockerfile
+++ b/runtime/base/php-82.Dockerfile
@@ -1,0 +1,139 @@
+# The container we build here contains everything needed to compile PHP + PHP.
+#
+# It can be used as a base to compile extra extensions.
+
+# PHP Build
+# https://github.com/php/php-src/releases
+# Needs:
+#   - zlib
+#   - libxml2
+#   - openssl
+#   - readline
+#   - sodium
+
+FROM bref/tmp/step-1/build-environment as build-environment
+
+
+###############################################################################
+# Oniguruma
+# This library is not packaged in PHP since PHP 7.4.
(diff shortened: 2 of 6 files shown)


t-richard:
Thanks for opening this, I approved the CI to see how it goes!

dhrrgn:
Thanks @t-richard. Looks like it failed installing xdebug for the dev images, which makes sense.

Looks like when the 8.1 runtimes were added the dev images were built with `--target=without_extensions`. Is that still the best option (until xdebug is ready for 8.2)?

t-richard:
Yes, we'll have to do that again 👍

dhrrgn:
@t-richard I implemented the skipping of extensions differently than was done previously.

Can you approve the CI and se if it passes now. 

I am open to changing how we skip the extension install, but want to make sure everything else works, then we can address any changes wanted. 

Thanks!

runtime/layers/fpm-dev/Dockerfile
```diff
@@ -1,11 +1,12 @@
 ARG PHP_VERSION
+ARG SKIP_EXTENSIONS=0
+
 FROM bref/build-php-$PHP_VERSION as build_extensions
 
-RUN pecl install xdebug
-RUN cp $(php -r "echo ini_get('extension_dir');")/xdebug.so /tmp
+COPY extensions.sh /tmp/extensions.sh
+RUN chmod +x /tmp/extensions.sh
+RUN /tmp/extensions.sh && rm /tmp/extensions.sh
```
dhrrgn, line 8: Moving this into a separate script makes more sense than adding another stage IMO.

Using a separate bash script is also more maintainable and flexible in this scenario IMO.

mnapoli:
Thank you!

I'd rather avoid having a new script file (the build is already very complex), could we just skip building `php-82-fpm-dev` for now instead?

dhrrgn:
@mnapoli Works for me. I have reverted that change and commented out all references to the `php-82-fpm-dev` build.

docs/runtimes/README.md
```diff
@@ -71,17 +71,19 @@ functions:
 
 The `${...}` notation is the [syntax to use variables](https://serverless.com/framework/docs/providers/aws/guide/variables/) in `serverless.yml`. Bref provides a serverless plugin ("`./vendor/bref/bref`") that provides those variables:
 
+- `${bref:layer.php-82}`
 - `${bref:layer.php-81}`
 - `${bref:layer.php-80}`
 - `${bref:layer.php-74}`
 - `${bref:layer.php-73}`
+- `${bref:layer.php-82-fpm}`
```
GrahamCampbell, line 79: We should probably not include this, or add a note saying coming soon.

GrahamCampbell (changes requested):

docs/runtimes/README.md
dhrrgn: Ah, shoot. Missed that one. I'll update after the CI run. 

How about making it:

```markdown
- `${bref:layer.php-82-fpm}` **experimental**
```

(and the php-82 layer above as well)

t-richard (approved):

runtime/base/php-82.Dockerfile
```diff
@@ -0,0 +1,139 @@
...
+    curl -Ls https://github.com/kkos/oniguruma/releases/download/v${VERSION_ONIG}/onig-${VERSION_ONIG}.tar.gz \
+    | tar xzC ${ONIG_BUILD_DIR} --strip-components=1
+WORKDIR  ${ONIG_BUILD_DIR}/
+RUN set -xe; \
+    ./configure --prefix=${INSTALL_DIR}; \
+    make -j $(nproc); \
+    make install
+
+
+ENV VERSION_PHP=8.2.0alpha2
```
t-richard, line 39: Alpha 3 has been released yesterday, could you update it here ?

dhrrgn:
@t-richard Done. I also updated the download URL, as it changed. 

Also, I added the `**experimental**` note in the docs. 

Let me know if there's anything else you need me to do. Thanks!

mnapoli:
Let's gooooo!

Thank you @dhrrgn

GrahamCampbell:
The download URL typically changes for every release or every other release, for pre-releases. We probably should make the download URL part of the env variable, rather than just the version.

mnapoli:
It's not a big deal to change the URL in the Dockerfile. I'd rather avoid complicating the build any further.

Issue #1248, opened 2022-07-14
Conflict between separateVendor and serverless-iam-roles-per-function plugin


NicolasMugnier:
**Description:**

The custom parameter `separateVendor` allow to move the `vendor` directory into an S3 bucket and the bref runtime will copy the content of the directory in `/tmp/vendor` as describe in the [documentation](https://bref.sh/docs/environment/php.html#separate-vendor-folder).

It works well as soon as I have installed the [IAM roles per function serverless plugin](https://www.npmjs.com/package/serverless-iam-roles-per-function). As well described in bref [documentation](https://bref.sh/docs/environment/serverless-yml.html#permissions) this serverless plugin allow to define specific roles at the lambda level, which is a good practice. 

But when using `separateVendor` custom parameter and `serverless-iam-roles-per-function` plugin, I've got the following error when trying to invoke the lambda after deploy : 

```php
PHP Fatal error: Uncaught Exception: Unable to unzip vendor archive. in /opt/bref/breftoolbox.php:69
Stack trace:
#0 /opt/bref/bootstrap.php(15): Bref\ToolBox\BrefToolBox::downloadAndConfigureVendor()
#1 {main}
thrown in /opt/bref/breftoolbox.php on line 69
````
**How to reproduce:**

In order to reproduce this error, in the `serverless.yml` file add the following configuration (I've put the whole file content so you have all the context) : 

```yaml
service: app

provider:
    name: aws
    region: eu-west-1
    runtime: provided.al2

custom:
    bref:
        separateVendor: true # true in order to enable the feature

plugins:
    - serverless-iam-roles-per-function # when disable this plugin the `separateVendor` feature works well
    - ./vendor/bref/bref

functions:
    addBook:
        handler: ./src/Events/create/index.php
        description: 'Add a new book'
        layers:
            - ${bref:layer.php-81}
        iamRoleStatements:
            - Effect: 'Allow'
              Action:
                - dynamodb:PutItem
                - dynamodb:GetItem
              Resource: 'arn:aws:dynamodb:${aws:region}:${aws:accountId}:table/book'

    getBook:
        handler: ./src/Events/read/index.php
        description: 'Get a specific book'
        layers:
            - ${bref:layer.php-81}
        iamRoleStatements:
            - Effect: 'Allow'
              Action:
                - dynamodb:GetItem
              Resource: 'arn:aws:dynamodb:${aws:region}:${aws:accountId}:table/book'

    deleteBook:
[...]

PR #1249, opened 2022-07-14
Update serverless per function permissions documentation

NicolasMugnier:
WHY : clarify the documentation about **per function** permissions usage with `separateVendor: true` parameter.

related to this issue : https://github.com/brefphp/bref/issues/1248


mnapoli:
Thanks!

PR #1259, opened 2022-08-10
Add header explanation for Laravel Nova

CodeTechNL:
I've been searching why my Laravel Nova wasn't working. Some extra explanation for the docs should help others out

### DIFF
--- a/docs/frameworks/laravel.md
+++ b/docs/frameworks/laravel.md
@@ -178,11 +178,12 @@ constructs:
       - X-Requested-With
       # Laravel Framework Headers
       - X-Csrf-Token
-      # Other Headers, e.g. Livewire
-      - X-Livewire
+      # Other Headers, e.g. Livewire, Laravel Nova
+      # - X-Livewire
+      # - X-Inertia
 ```
 
-> If you use Laravel Nova, don't forget to add the header `X-Inertia`
+> If you use Laravel Nova, don't forget to uncomment the header `X-Inertia`
 
 Before deploying, compile your assets using Laravel Mix.
 
--- a/docs/frameworks/laravel.md
+++ b/docs/frameworks/laravel.md
@@ -183,8 +183,6 @@ constructs:
       # - X-Inertia
 ```
 
-> If you use Laravel Nova, don't forget to uncomment the header `X-Inertia`
-
 > Note: the limit of forwardedHeaders for AWS is set to 10 
 
 Before deploying, compile your assets using Laravel Mix.
--- a/docs/frameworks/laravel.md
+++ b/docs/frameworks/laravel.md
@@ -182,6 +182,8 @@ constructs:
       - X-Livewire
 ```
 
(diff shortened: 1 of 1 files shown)


afu-dev:
In the file you modified, we already have a list of "other headers": https://github.com/brefphp/bref/blob/master/docs/frameworks/laravel.md?plain=1#L181-L182

Shouldn't we add the `X-Inertia` in the same list?

CodeTechNL:
@afu-dev  I’ve added it as optional. Not everyone uses Nova or Inertia. There is a limit of 9 (if I’m right) on the forwarded headers.

afu-dev:
Likewise, not everyone uses Livewire. I propose to have all optional headers in the same place instead of one in the `forwardedHeaders` list and one underneath the source code.

Indeed, the `forwardedHeaders` list is limited (to 10 headers, ref: https://github.com/getlift/lift/blob/master/docs/server-side-website.md?plain=1#L365), and adding everything to it will break the code. Maybe a note like you did with "if you use this, add this header, if you use that, add that header"?

(Again, just a proposition to be extra clear with everyone reading the doc 😉)

CodeTechNL:
I agree on it @afu-dev, I’ll update it later today and change livewire too. Also I’ll add a notice about the limit of 10 forwardedHeaders

mnapoli:
Thanks for the PR, and good points here!

How about adding it to the list, but commented (with a comment above saying "uncomment if you use Nova"). We could even do the same thing for the Livewire header (commented by default), that way we avoid hitting the limit of 10 headers too easily?

CodeTechNL:
@mnapoli ill see what’s the best to keep it clear and brief and update my PR

CodeTechNL:
@mnapoli little bit late, kinda busy. Updated as requested

docs/frameworks/laravel.md
```diff
@@ -178,10 +178,15 @@ constructs:
       - X-Requested-With
       # Laravel Framework Headers
       - X-Csrf-Token
-      # Other Headers, e.g. Livewire
-      - X-Livewire
+      # Other Headers, e.g. Livewire, Laravel Nova
```
afu-dev, line 181:
```suggestion
      # Other Headers (e.g. Livewire, Laravel Nova), uncomment based on your needs
```

docs/frameworks/laravel.md
```diff
@@ -178,10 +178,15 @@ constructs:
...
+      # - X-Livewire
+      # - X-Inertia
 ```
 
+> If you use Laravel Nova, don't forget to uncomment the header `X-Inertia`
+
```
afu-dev, line 187:
```suggestion
```

afu-dev (changes requested):
The note is specific for Laravel Nova, but the code now contains headers for Livewire AND Nova. I would make the comment in the YAML more detailed and remove the note.

mnapoli:
Thank you both for contributing!

PR #1263, opened 2022-08-31
Change from 'error' to 'warning' if `serverless` not installed

### LINKED ISSUE #1262: RFC: Switch to `->warning()` when `serverless` executable is not found
I've been putting together a sample application for a workshop on refactoring a stateful PHP application to run on top of Lambda with Bref, starting with an alpine PHP-FPM container and a separate Serverless container.

I'm starting with the bog-standard PHP-FPM alpine image rather than the Bref image because I want to show a step-by-step refactoring to using Serverless from first-principles _without_ a lot of the abstractions and improvements which does the hard stuff for you.

But... when I try to run `vendor/bin/bref init` I run into the blocking exception

```
'The `serverless` command is not installed.'
```

And then the script quits with `return 1;`

As far as I can see this is just a sanity check to make sure the project would be deployable with the current setup.

However, I know that serverless _is_ installed, it's just not available within the PHP container I'm running in my local env. I've had to add the following additional concerns into my dockerfile just to be able to init the project (which feels like a coupling it shouldn't need to be aware of):

```
RUN apt update
RUN apt install nodejs npm -y

RUN npm i -g serverless
```

I propose to change the `$io->error(...)` to `$io->warning(...)` and allow the script to continue, so that Docker setups which have moved the SLS executable into another container can proceed without hacking the existing Dockerfile.

Currently we have this:

```
$exeFinder = new ExecutableFinder;
if (! $exeFinder->find('serverless')) {
    $io->error(
        'The `serverless` command is not installed.' . PHP_EOL .
        'Please follow the instructions at https://bref.sh/docs/installation.html'
    );

    return 1;
}
```

I propose to switch to:

```
$exeFinder = new ExecutableFinder;
    if (! $exeFinder->find('serverless')) {
        $io->warning(
        'The `serverless` command is not installed locally.' . PHP_EOL .
        'You will not be able to deploy your application unless it is installed' . PHP_EOL .
        'Please follow the instructions at https://bref.sh/docs/installation.html to install'  . PHP_EOL .
        'If you have the `serverless` command available elsewhere (eg in a Docker container) you can ignore this warning'
    );
}
```

WDYT?

nealio82:
If `serverless` isn't installed in the local env, we should warn the user instead of blocking them from proceeding, as they may already have serverless available via another container (if using Docker, for example).

Closes #1262

### DIFF
--- a/src/Console/Command/Init.php
+++ b/src/Console/Command/Init.php
@@ -22,12 +22,12 @@ final class Init extends Command
         $io = new SymfonyStyle($input, $output);
         $exeFinder = new ExecutableFinder;
         if (! $exeFinder->find('serverless')) {
-            $io->error(
+            $io->warning(
                 'The `serverless` command is not installed.' . PHP_EOL .
-                'Please follow the instructions at https://bref.sh/docs/installation.html'
+                'You will not be able to deploy your application unless it is installed' . PHP_EOL .
+                'Please follow the instructions at https://bref.sh/docs/installation.html' . PHP_EOL .
+                'If you have the `serverless` command available elsewhere (eg in a Docker container) you can ignore this warning'
             );
-
-            return 1;
         }
 
         if (file_exists('serverless.yml') || file_exists('index.php')) {


t-richard:
Thanks @nealio82 !

PR #1272, opened 2022-09-12
Make existing files just a warning instead of a crash

Martincic:
## Make existing files warnings (index.php & serverless.yml)

So I've just started out with serverless and I've noticed a strange 'feature', that whether I own `serverless.yml` or `index.php`, I cannot initialize bref. I'm creating both workshops and converting old projects into serverless applications. The problem here is, that most often old applications, as well as workshops, have predefined `index.php`. I cannot simply dive into serverless without having a use-case (which most often contains **just** `index.php`).

I would find it much less annoying to tell my audience: 
```
Now, initialize bref!
```
much rather then
```
Now save the contents of index.php file
Delete index.php file
Initialize bref
Overwrite the contents of the new index.php with the old contents you copied.
```

Once I added this for `index.php`, I was looking at that `serverless.yml` restriction and made them both warnings rather than critical errors.

### Explanation
If this init command can only work with fresh projects, I would much rather go off creating my own files, rather than modifying the project to use the command to create files. Possibly if this is not something of your interest, you could separate the init command into init-entry and init-serverless or give out options within init itself. I just need something which isn't a critical error.

Best,
Tomas

### DIFF
--- a/src/Console/Command/Init.php
+++ b/src/Console/Command/Init.php
@@ -5,6 +5,7 @@ namespace Bref\Console\Command;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ConfirmationQuestion;
 use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\Process\ExecutableFinder;
@@ -20,6 +21,11 @@ final class Init extends Command
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
         $io = new SymfonyStyle($input, $output);
+        $helper = $this->getHelper('question');
+
+        /**
+         * @param $filesToGitAdd<string>
+         */
         $filesToGitAdd = [];
         $exeFinder = new ExecutableFinder;
         if (! $exeFinder->find('serverless')) {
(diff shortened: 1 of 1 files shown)


afu-dev:
I'm not keen on this modification.

`./bref init` should be used to initialize a new repo and you should end up in a working state after that.

If you already have an `index.php` in your repo and if `./bref init` only creates a `serverless.yml` file, running `sls deploy`, you will mostly have an error when requesting the endpoint.

In that case, the migration process should look more like this:

```
- composer require bref/bref
- create a `serverless.yml` (from a template?)
- if present, change `index.php` to make it work
```

Maybe the Bref CLI could have a new command to create a `serverless.yml` file from a template?

Martincic:
As mentioned, I'm up for this, I just need something to separate both files. 

From your response, I think the best course of action would be to add optional flags to `./bref init` such as `-s` and `-i` which would create either `serverless.yml` or `index.php` respectively from the templates. The default behavior would be to create both, with flags we could define which one we wanted to create.

Possibly adding a subquestion to the `-s` flag would ask where we want to point the handler (where's the existing `index.php` located)?

afu-dev:
In that case, maybe a `--no-php-file` and `--no-yml-file` options can work to avoid generating a file that you don't need. This can be more specific: `--no-php-index` or `--no-serverless-config`.

Let's wait for a feedback from the maintainers 😉

Martincic:
@mnapoli I like the idea, more elegant then flags, I'll change it up this week. Would it make sense for both serverless and index files, or just index?

> First-time contributors need a maintainer to approve running workflows

Also, could you approve the workflow so I can work on this?

mnapoli:
> Would it make sense for both serverless and index files, or just index?

I think both.

Martincic:
Hey, I've redone this PR so that it asks questions instead of warnings. I refactored the function a bit so that it uses a static function because of code repetition. The only caveat is that there's replacing of `PHP_VERSION` within both `serverless.yml` and `index.php` to reduce code repetition, but I've figured it's no problem since `index.php` never contains `PHP_VERSION`, and if it would, we would know the exact one.

afu-dev (commented):
The complexity of the code is maybe a bit too high now.

Instead of a static function in the middle of the method, maybe a new private method should be created?

mnapoli:
Thanks for the effort! I would tend to agree with @afu-dev, a private method would be more "common" and boring. Boring code is useful on open-source projects where a different contributor (with maybe a different experience) might want to work on the same code in a few years 😅

Martincic:
There, I've refactored it into a separate function, and I've also joined the part where it adds files to git into a new function as well. This should fix the code complexity issue.

mnapoli:
Ohh yeah, that looks even better than what we have on master! Thanks, I'll merge once the tests pass.

src/Console/Command/Init.php
```diff
@@ -6,20 +6,23 @@
...
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
-use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\Process\ExecutableFinder;
 use Symfony\Component\Process\Process;
 
 final class Init extends Command
 {
+    private SymfonyStyle $io;
+    private string $rootPath;
```
mnapoli, line 15: Unfortunately this isn't compatible with PHP 7.3

Martincic: This should do it!
```
/**
     * Symfony console input/output handler
     *
     * @var SymfonyStyle
     */
    private $io;

    /**
     * Absolute path pointing into project's
     * template directory
     *
     * @var string
     */
    private $rootPath;
```

src/Console/Command/Init.php
```diff
@@ -44,42 +53,49 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+    private function createFile(string $file): void
+    {
+        $overwrite = true;
 
-        $template = str_replace('PHP_VERSION', PHP_MAJOR_VERSION . PHP_MINOR_VERSION, $template);
+        $this->io->writeln("Creating $file");
 
-        file_put_contents('serverless.yml', $template);
+        if (file_exists($file)) {
+            $overwrite = $this->io->confirm("An $file file already exists, do you want to overwrite it?", true);
```
t-richard, line 83: `An` should be followed by a vowel and that's not the case

May be we shoudl phrase it like this ?

```suggestion
            $overwrite = $this->io->confirm("A file named $file already exists, do you want to overwrite it?", true);
```

Martincic:
Sure thing, English is not my primary, sorry 'bout that.

src/Console/Command/Init.php
```diff
@@ -44,42 +53,49 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
-         * We check if this is a git repository to automatically add files to git.
-         */
-        if ((new Process(['git', 'rev-parse', '--is-inside-work-tree']))->run() === 0) {
-            foreach ($filesToGitAdd as $file) {
-                (new Process(['git', 'add', $file]))->run();
+            /*
+             * We check if this is a git repository to automatically add file to git.
+             */
+            if ((new Process(['git', 'rev-parse', '--is-inside-work-tree']))->run() === 0) {
+                    (new Process(['git', 'add', $file]))->run();
```
afu-dev, line 95: There is an extra indentation here

src/Console/Command/Init.php
```diff
@@ -44,42 +53,49 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
-            $io->success([
-                'Project initialized and ready to test or deploy.',
-                'The files created were automatically added to git.',
-            ]);
-        } else {
-            $io->success('Project initialized and ready to test or deploy.');
-        }
 
-        return 0;
+            $this->io->success("$file successfully created and added to git automatically.");
```
afu-dev, line 98: The message could be conditioned with the previous if-block to show "added to git..." only when the user is in a git repo.

afu-dev (commented):
I don't approve or decline, only two suggestions. Anyway, the code is better than before, so that's a win 👍

t-richard (approved):
No need to apologize, we all do our best. From the discussion we've had here, your English is good actually :+1: 

LGTM

src/Console/Command/Init.php
t-richard: Do you mean something like this?

```php
$successMessage = "$file successfully created"

if ((new Process(['git', 'rev-parse', '--is-inside-work-tree']))->run() === 0) {
    (new Process(['git', 'add', $file]))->run();

    $successMessage .= ' and added to git automatically.'
}

$this->io->success($successMessage);
```

Martincic:
> No need to apologize, we all do our best. From the discussion we've had here, your English is good actually +1

Grammarly does wonders! 

There I've refactored the output in the following manner:
```
$message = "$file successfully created";
            if ((new Process(['git', 'rev-parse', '--is-inside-work-tree']))->run() === 0) {
                (new Process(['git', 'add', $file]))->run();
                $message .= ' and added to git automatically';
            }

            $this->io->success("$message.");
```

src/Console/Command/Init.php
Martincic: Resolved in [latest commit](https://github.com/brefphp/bref/pull/1272/commits/ff43ac262ea4429e238bd6f3280109304301e760)

src/Console/Command/Init.php
Martincic: Resolved in [latest commit](https://github.com/brefphp/bref/pull/1272/commits/ff43ac262ea4429e238bd6f3280109304301e760)

afu-dev (approved):
🚀

mnapoli:
Thank you all!

PR #1287, opened 2022-10-11
Revamp init command

afu-dev:
First draft for the revamp of the `init` command. I'm looking for feedbacks on what's done for now.

Preview of the new command output:

image

image

<details>
<summary>Final content of the `serverless.yml` file for function</summary>

```yml
service: bref-tmp-test-project

provider:
  # general settings
  name: aws
  stage: dev
  region: eu-west-3
  profile: default
  deploymentMethod: direct
  # global function settings (apply to each declared function)
  runtime: provided.al2
  memorySize: 1024
  timeout: 6
  logRetentionInDays: forever
  layers:
    - ${bref:layer.php-81}

plugins:
  - ./vendor/bref/bref

functions:
  hello:
    description: 'Deployed with Bref dev-master for PHP 8.1'
    handler: index.php

# Exclude files from deployment
package:
  patterns:
    - '!tests/**'
```
</details>

<details>
<summary>Final content of the `serverless.yml` file for webapp</summary>

```yml
service: my-new-php-project

provider:
  # general settings
  name: aws
  stage: dev
  region: eu-west-1
  profile: default
  deploymentMethod: direct
  # global function settings (apply to each declared function)
  runtime: provided.al2
  memorySize: 1024
  timeout: 28 # in seconds (API Gateway has a timeout of 29 seconds)
  logRetentionInDays: forever
  layers:
    - ${bref:layer.php-81-fpm}

plugins:
  - ./vendor/bref/bref

functions:
  api:
    handler: index.php
    description: 'Deployed with Bref dev-master for PHP 8.1'
    events:
      - httpApi: '*'

# Exclude files from deployment
package:
  patterns:
    - '!node_modules/**'
    - '!tests/**'
```
</details>

List of changes:

- command asks interactively:
  - the name of the project (default to the name of the folder)
  - the region to deploy to (default: us-east-1)
  - the runtime to use (default: function)
  - if files are present:
    - if index.php should be overwritten (default: yes)
    - if serverless.yml should be overwritten (default: yes)
- if the option `--no-interaction`/`-n` is used, the command will ask for the required data (via `--name=NAME`, `--runtime=RUNTIME`, etc.)
- serverless.yml file (for webapp AND functions):
  - `service` field contains the name of the project
  - `provider`:
    - added the region from the last step
    - added the stage and profile with the default values used by serverless (stage = dev and profile = default)
    - added `deploymentMethod` with the `direct` value (faster deployment, new default in sls v4)
    - added memorySize, timeout, and logRetentionInDays, with the default values used by AWS.
    - switched layers to the provider section
- index.php (for webapp):
  - removed Google Fonts (not GDPR compliant without asking for authorization from the user. I know we don't host the final website, but it's like a "trap" that triggers if the user deploys the demo page and checks it)

I would love to add a bunch of comments in the serverless.yml file, like here: https://www.serverless.com/framework/docs/providers/aws/guide/serverless.yml. But I'm afraid it will be too much for newcomers and redundant for veterans.

Remaining TODOs:
- [ ] Documentation (section "first steps")
- [ ] Fix CS 😅 
- [ ] tbd


src/Console/Command/Init.php
```diff
@@ -73,31 +154,40 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+    }
+
+    /**
+     * @throws Exception
+     */
+    public function validateName(string $name): string
+    {
+        $regex = "/^[a-zA-Z][0-9a-zA-Z-]+$/";
+        if (!preg_match($regex, $name)) {
+            throw new Exception("Project name must match pattern \"$regex\"");
```
mnapoli, line 188: Could we get a more descriptive error message for users?

template/function/serverless.yml
```diff
@@ -1,21 +1,29 @@
...
 provider:
-    name: aws
-    region: us-east-1
-    runtime: provided.al2
+  # general settings
+  name: aws
+  stage: dev
+  region: TPL_REGION
+  profile: default
+  deploymentMethod: direct
```
mnapoli, line 9:
```suggestion
  # Use the newer and faster deployment method
  deploymentMethod: direct
```

As a user I hate when there's boilerplate I don't understand. If we can't remove the boilerplate, we can explain it.

template/function/serverless.yml
```diff
@@ -1,21 +1,29 @@
...
+  # global function settings (apply to each declared function)
```
mnapoli, line 10: I think this comment is more confusing than anything, and useless for 90% of PHP devs?

template/function/serverless.yml
```diff
@@ -1,21 +1,29 @@
...
+  runtime: provided.al2
+  memorySize: 1024
+  timeout: 6
```
mnapoli, line 13: Both of these options are already the default value, I think if we remove these it will be less overwhelming for new users.

template/function/serverless.yml
```diff
@@ -1,21 +1,29 @@
...
+  logRetentionInDays: forever
+  layers:
+    - ${bref:layer.php-TPL_PHP_VERSION}
```
mnapoli, line 16: I'm afraid of putting the layer on all functions, if a user wants to use FPM later it will mess things up.

template/function/serverless.yml
```diff
@@ -1,21 +1,29 @@
...
+  name: aws
+  stage: dev
+  region: TPL_REGION
+  profile: default
+  deploymentMethod: direct
+  # global function settings (apply to each declared function)
+  runtime: provided.al2
+  memorySize: 1024
+  timeout: 6
+  logRetentionInDays: forever
```
mnapoli, line 14: Can we explain why we need this line? I'm not even sure we need this, why do we want retention forever?

template/function/serverless.yml
```diff
@@ -1,21 +1,29 @@
...
+  - ./vendor/bref/bref
 
 functions:
-    hello:
-        handler: index.php
-        description: ''
-        layers:
-            - ${bref:layer.php-PHP_VERSION}
+  hello:
+    description: 'Deployed with Bref TPL_BREF_VERSION for PHP TPL_PHP_VERSION_PRETTY'
```
mnapoli, line 23: I don't think this is a good description of what the function does. Only the user knows what to put in here no?

template/function/serverless.yml
```diff
@@ -1,21 +1,29 @@
-service: app
+service: TPL_SERVICE_NAME
 
 provider:
-    name: aws
-    region: us-east-1
-    runtime: provided.al2
+  # general settings
+  name: aws
```
mnapoli, line 5: 4 spaces indentation plz (easier to read, less likely to make a mistake, complies with most PHP practices)

template/function/serverless.yml
```diff
@@ -1,21 +1,29 @@
-service: app
+service: TPL_SERVICE_NAME
 
 provider:
-    name: aws
-    region: us-east-1
-    runtime: provided.al2
+  # general settings
```
mnapoli, line 4: Not sure this comment is useful

template/webapp/serverless.yml
```diff
@@ -0,0 +1,32 @@
...
+  # general settings
+  name: aws
+  stage: dev
+  region: TPL_REGION
+  profile: default
+  deploymentMethod: direct
+  # global function settings (apply to each declared function)
+  runtime: provided.al2
+  memorySize: 1024
+  timeout: 28 # in seconds (API Gateway has a timeout of 29 seconds)
```
mnapoli, line 13: We need to move these back into the function. It's common to start with FPM and add "function" functions, that will not work with this template.

mnapoli (commented):
Awesome, that looks really good!

I have started added inline comments, but more generally my feedback is "this changes too many things at once" 😅

The `serverless.yml` templates were intentionally as small as possible to avoid overwhelming new users and start as simple as possible. I think we should keep these same goals in mind, and add only what is truly valuable (and there's room for improvement here, so I still think there are ideas in this PR we can keep).

It seems like `function` is now the default, but currently AFAIR it's web-app: it's intentional, because most new users start with a web app.

Also I'm not sure about the renaming from `http` to `webapp`, but that sounds reasonable (considering all the renamings done in the docs). I'll keep thinking about that. Maybe `web-app` might look less weird, not sure what's bothering me here.

> added the stage and profile with the default values used by serverless (stage = dev and profile = default)

Why do we need this? I don't think I understand what it's doing actually.

I commented inline for other points because that's easier to discuss there with separate threads.

> removed Google Fonts

Could we explore that in a separate PR?

> I would love to add a bunch of comments in the serverless.yml file

That sounds good for anything where the comment is truly useful the exact moment someone starts using Bref and sees `serverless.yml` for the first time. I.e. to answer the kind of questions where they might actually questions deploying or not because they don't understand an option.

Anything else might be actually making things worse by adding noise and making it look more complex that necessary.

In other words, this is a tough line to figure out 😄

mnapoli:
Hey hey @afu-dev! Just a heads up, this wasn't really planned 2 weeks ago but since then I've been looking at getting rid of symfony/console: #1303

I'd still like to keep some of the work here, but I wonder how doable that is without symfony/console (because it's our biggest dependency) 🤔 

It's not really in scope here, but just sharing my thoughts as I think this impacts this PR a lot.

afu-dev:
Following a discussion with @mnapoli, I'm closing this PR for 1.x. Here are the details:

This PR was a draft with a few ideas I would like to develop for Bref. First, I will need to split the PR into multiple smaller ones to avoid changing a big chunk of the codebase in one go and to make more "atomic changes". Secondly, Bref v2 will change how the CLI works. Accordingly, I will redo the PR when v2 is out.

PR #1290, opened 2022-10-13
Remove init set for the function layer

### LINKED ISSUE #1288: Some php.ini values are overridden
Hi,

I use a `php/conf.d/php.ini` file to override the bref default php configuration and I am not able to update some of them.

In my case, I tried to update the `display_errors` value (I also tried to update the `error_reporting` for the example).
This is my `php/conf.d/php.ini` file:
```
display_errors = 0
error_reporting = E_USER_DEPRECATED
```
But at the first line of my `index.php`, I tried to get these two values and I get this:
```
Directive => Local Value => Master Value
display_errors => STDOUT => Off
error_reporting => 32767 (E_ALL) => 16384 (E_USER_DEPRECATED)
```
The master value is well updated, but the local value has been overriden.
If I add the following line, everything is fine again but I would like to avoid doing this:
`ini_set('display_errors', 0)`

I only get this issue when using the php-function layer and not with the php-fpm layer.

I'm guessing that the value is probably updated by bref just when the lambda start, maybe there https://github.com/brefphp/bref/blob/master/runtime/layers/function/bootstrap.php#L6 ?

If yes, Is there a reason why bref update these values outside of the initial php.ini file https://github.com/brefphp/bref/blob/master/runtime/layers/function/php.ini ?

Thank you in advance for your responses :)

josephLaurent:
For more informations: https://github.com/brefphp/bref/issues/1288

Fixes #1288

### DIFF
--- a/runtime/layers/function/bootstrap.php
+++ b/runtime/layers/function/bootstrap.php
@@ -3,9 +3,6 @@
 use Bref\Bref;
 use Bref\Runtime\LambdaRuntime;
 
-ini_set('display_errors', '1');
-error_reporting(E_ALL);
-
 $appRoot = getenv('LAMBDA_TASK_ROOT');
 
 if (getenv('BREF_DOWNLOAD_VENDOR')) {


afu-dev (approved):

deleugpn:
Since this will change the layer from `E_ALL` to `E_ALL & ~E_DEPRECATED & ~E_STRICT`, maybe this should be documented as a potential slight breaking change or perhaps `php.ini` should be adjusted to `E_ALL`?

mnapoli:
Good point. PHP recommends `E_ALL & ~E_DEPRECATED & ~E_STRICT` for production (https://github.com/php/php-src/blob/d758f5867281dc6c423288ef77696e748547e71f/php.ini-production#L477-L486), and this is what we intended initially through php.ini. Makes sense to not log deprecated warnings by default for everyone 🤔 

Since the end result is "less" logs (deprecated logs might be removed for those who may have some), the change shouldn't cause, for example, the triggering of log alarms or similar. We will always have the option to revert if that causes unexpected issues.

I will document that in the release.

I'll keep that open for a few days if anyone wants to add to the discussion.

afu-dev:
I think Bref should give you the possibility to deploy a "production-ready" project. Switching to `E_ALL & ~E_DEPRECATED & ~E_STRICT` is the recommended/right value to achieve that.

Regarding the BC: can this be considered a BC? In other words, can this change breaks the project of others? _I don't think so_. Upgrading from one PHP version to another while having some deprecated errors will break applications, but it is the responsibility of the developer, not the package.

Anyway, I think we should complete the documentation here: https://bref.sh/docs/environment/logs.html#php-errors-and-warnings

Currently:
> By default, all PHP errors, warnings and notices emitted by PHP will be forwarded into CloudWatch.

E_DEPRECATED errors are considered "Run-time notices", so not "all notices" are forwarded into CloudWatch.

Ref: https://www.php.net/manual/en/errorfunc.constants.php

deleugpn:
I'm not 100% sure, but my fear is that if you have a custom error handler that converts warnings and deprecations into exceptions, your code flows one way. When the developer runs `composer update`, he/she will blindly update Bref without knowing that deprecations and warnings are being silenced which may not be converted into exception anymore which will completely change the flow of the code.

To reiterate, I don't know if this setting has the power to cause this, but that's my fear. And custom error handler is not uncommon, Laravel comes with one by default on every Laravel project.

mnapoli:
Since we are working on 2.0 right now, I think we shouldn't sweat this too much. Let's target 2.0 for this change.

Here is the code for 2.0: https://github.com/brefphp/aws-lambda-layers/blob/main/layers/function/bootstrap.php The `ini_set` call is gone, could you confirm that this is good and we can close this PR?

PR #1295, opened 2022-10-14
Upgrade to postgres 15.0

GrahamCampbell:


mnapoli:
Will this have any impact on users with lower versions? Any breaking change that could apply here?

GrahamCampbell:
As far as I'm aware it's backwards compatible, just like the upgrade to 14 was.

mnapoli:
Thanks!

PR #1297, opened 2022-10-17
Opt-in: throw exceptions on lambda invoke errors

negoziator:
**Motivation**

When an exception is thrown by an invokation, the exception is getting swallowed.
This basically means, that it's impossible to do exception handling.

In my opinion, a package should throw exceptions and let the user determine how to handle them.

I understand the reasoning for why to log errors in a CW way, but i think that having the opportunity to opt-in on exceptions instead, is a great add to the package.

**Example**

Here's a simple example of an application using Bref that does the following:

1. Get an event from a lambda-invoke (using Bref's EventBridgeEvent class)
2. Parse the event
3. Store parsed data to DynamoDB

In this case the AWS SDK throws an exception, but rather than the application has a chance to "catch" the exception and gracefully handle it. It silently get's catched, and the error is echo'ed out ( fine with the echo, since it does it in the CW style. )

As you also can see on the screenshot. This is echoed out to stdout, which means it's really hard to destinguish logs on their level. (INFO, WARNING. ERROR etc.) - this can of course be done, by parsing all CW-logs and looking for the word "Invoke error" and then transform the log into an error-log instead, but that seems really cumbersome.

Screenshot:
image

### DIFF
--- a/src/Runtime/LambdaRuntime.php
+++ b/src/Runtime/LambdaRuntime.php
@@ -94,6 +94,10 @@ final class LambdaRuntime
         } catch (\Throwable $e) {
             $this->signalFailure($context->getAwsRequestId(), $e);
 
+            if (in_array(getenv('BREF_INVOKE_LAMBDA_ERROR_THROW_EXCEPTIONS'), ['true', '1'], true)) {
+                throw $e;
+            }
+
             return false;
         }
 
--- a/tests/Runtime/LambdaRuntimeTest.php
+++ b/tests/Runtime/LambdaRuntimeTest.php
@@ -339,6 +339,42 @@ ERROR;
         $this->assertErrorInLogs('Exception', 'The lambda handler must be a callable or implement handler interfaces');
     }
 
+    public function test_throw_exception_if_env_is_set_to_true()
+    {
+        putenv('BREF_INVOKE_LAMBDA_ERROR_THROW_EXCEPTIONS=true');
+
+        // Will throw exception
+        $this->expectException(\Throwable::class);
+        $this->expectExceptionMessage('The lambda handler must be a callable or implement handler interfaces');
+
+        $eventData = json_decode(file_get_contents(__DIR__ . '/../Event/Http/Fixture/ag-v1-simple.json'), true);
+        $this->givenAnEvent($eventData);
+
+        $this->runtime->processNextEvent(null);
+
+        // Will log exception
+        $this->assertInvocationErrorResult('Exception', 'The lambda handler must be a callable or implement handler interfaces');
+        $this->assertErrorInLogs('Exception', 'The lambda handler must be a callable or implement handler interfaces');
+    }
+
+    public function test_throw_exception_if_env_is_set_to_one()
+    {
+        putenv('BREF_INVOKE_LAMBDA_ERROR_THROW_EXCEPTIONS=1');
+
+        // Will throw exception
+        $this->expectException(\Throwable::class);
+        $this->expectExceptionMessage('The lambda handler must be a callable or implement handler interfaces');
+
+        $eventData = json_decode(file_get_contents(__DIR__ . '/../Event/Http/Fixture/ag-v1-simple.json'), true);
+        $this->givenAnEvent($eventData);
+
+        $this->runtime->processNextEvent(null);
+
+        // Will log exception
+        $this->assertInvocationErrorResult('Exception', 'The lambda handler must be a callable or implement handler interfaces');
+        $this->assertErrorInLogs('Exception', 'The lambda handler must be a callable or implement handler interfaces');
+    }
+
     /**
      * @param mixed $event
      */


src/Runtime/LambdaRuntime.php
```diff
@@ -94,6 +94,10 @@ public function processNextEvent($handler): bool
         } catch (\Throwable $e) {
             $this->signalFailure($context->getAwsRequestId(), $e);
 
+            if (in_array(getenv('BREF_INVOKE_LAMBDA_ERROR_THROW_EXCEPTIONS'), ['true', '1'], true)) {
```
GrahamCampbell, line 97: `getenv` is not really suitable for this. `$_SERVER` is always the best place for env variable reading and writing... yes PHP did a super bad job of naming things.

GrahamCampbell (changes requested):

src/Runtime/LambdaRuntime.php
negoziator: I have no argument againts that - can you elaborate on that?

Also, if i grep in the package for `getenv` i'ts all over the place.

mnapoli: We indeed use getenv all over the place, it's single-process single-thread everywhere, any issue we are risking?

mnapoli:
Hi @negoziator, thanks for opening this.

The LambdaRuntime class is very low-level and [internal](https://github.com/negoziator/bref/blob/dea4a15891a432cde73046db989908493400607d/src/Runtime/LambdaRuntime.php#L29), Bref users are not supposed to use it.

Could you explain how you use it? It seems you may be doing something quite sophisticated out of Bref's bounds. Or maybe this is just a mis-use of Bref that could be solved differently?

Also it is not clear what you would expect to happen with your change. If the exception is not caught, then the whole runtime process fails. It's like "exit 1" in a webserver, there is no-one that will catch the exception except the AWS Lambda service itself, and it will consider the exception as a runtime failure and consider the Lambda function to be broken.

> In my opinion, a package should throw exceptions and let the user determine how to handle them.

Just to be sure, are you using Bref? The part you modified is not really a package (in the traditional sense), that's why I'm very curious to understand the use case here.

negoziator:
Let me elaborate.

So we use bref by having a SAM template that uses the bref-layer like this:

image

Then the eventbridge handler handle the event like this (Note the commen with the exception):

image

As far as i see it - this is pretty straight forward, but i'm curious if we are doing something wrong?

mnapoli:
OK that looks good honestly! What is the problem you are facing? Or said differently, what would you expect to happen and what happens instead?

negoziator:
@mnapoli I would expect the exception that got thown - would still be thrown, so that my handling of the event could catch that exception and gracefully do something. Instead of the Lambda function just silently finishes the execution and logs the error out to stdout.

mnapoli:
To be transparent I think you are misunderstanding how AWS Lambda works.

> so that my handling of the event could catch that exception

Where would you catch it?

negoziator:
I might misunderstand it, but the way i would do it is like this:

So i'm running Laravel, which has an exception handler that cathes all exceptions.

This makes it possible to render the exception freely as you like.
What i'm doing is this:

First set the logging to be **stdout**

image

Then in the ExceptionHandler

I register a handler that catches all exceptions and transform them into json.

**Handler.php**
image

Implementation:

image


In this way - no "exceptions" is ever thrown - but returned as json ( that's just how i did it - but one could render it another way )

However - this is not possible to do, because the upper exception that is throw was catched by the LambdaRuntime.

This addition does not change anything unless one explicitly set the environment variable - which means you will have to know how to handle the exception that is getting thrown.

Hope it makes sense 😄

mnapoli:
Ah I see now. Here is how things work with the FPM layer: Lambda calls Bref, which calls Laravel's index.php, which calls your controller. If there's an exception in your controller, Laravel catches it and returns a nice error response.

In the function layer, Lambda calls Bref, which calls your handler class. Laravel is booted to retrieve the handler class, but Laravel does not wrap the handler class. That means Laravel is never aware of any exception happening, Laravel does not handler the call.

One thing you could do is catch exceptions yourself in your handler class. Another more convoluted thing would be to wrap the container integration, but that's very advanced.

Now that I think about it, that could be a feature of the Laravel bridge 🤔  I'm going to open an issue in https://github.com/brefphp/laravel-bridge

mnapoli:
OK actually thinking about it again: we need the exception to bubble up to AWS Lambda, else the invocation will not be marked as a failure. If we have Laravel handle the exception, it will not bubble it up to Lambda, so that would be bad.

Again, I insist but I would love to understand _why_ you want that exception to be handled by Laravel.

> In this way - no "exceptions" is ever thrown - but returned as json

This is an EventBridge event, there is no response to return (not even an HTTP response). That's why I'm still not seeing the use case.

negoziator:
Hmm - i think i understand now 😄 

The reason why i want to do it - is only to format the exception nicely and forward it (in this case to Datadog)

When the error is echoed out like it is - all errors is not formatted correctly for Datadog, which makes it really hard to see errors in the application.

So i could maybe just catch the exception myself - then format it - send it to Datadog and then throw the exception again, so the runtime also will get the error ?

negoziator:
@deleugpn @mnapoli Thanks alot for you taking your time on this 👍🏼 

I will do it as you proposed.

Closing the PR again.

Again, thank you very much for your extensive help 🎉

negoziator:
Just a follow up:

This works perfectly, and is super super simpel.

```php
$app = require __DIR__ . '/../../bootstrap/app.php';
$kernel = $app->make(Kernel::class);
$kernel->bootstrap();

try {
    $app->make(EventBridgeResolver::class)->resolveAndHandle($event);
} catch (\Throwable $e) {
    $app->make(ExceptionHandler::class)->report($e);
    throw $e;
}
```

PR #1312, opened 2022-11-01
Revert "Upgrade to curl 7.86.0"

GrahamCampbell:
Soooo, it turns out this is causing segfaults. Not sure why at this point. 😭


mnapoli:
Oh interesting, thanks for jumping in quickly on this. I'll merge and tag a new release 👍

PR #1325, opened 2022-11-14
Change content type check and add basic auth support

georgeboot:
I've discovered that its also possible for browsers (or in my case, `fetch` on a Cloudflare worker) to append a suffix to the urlencoded form header: `application/x-www-form-urlencoded;charset=UTF-8` is the value that was sent in my case.

According to [the standard ](https://url.spec.whatwg.org/#application/x-www-form-urlencoded)the values should always be utf-8 and hence it does not need to be specified, but here we are.

I've updated the code to check if the header starts with the value, instead of checking for equality.

### Basic auth support
While in there, I've also implemented support for basic auth, as that relies on PHP setting $_SERVER vars.

### $_SERVER vars
While in there, I've also added support for some widely used $_SERVER vars. Admitted, modern PHP apps should not depend on them, but to make the entry barrier to Bref as low as possible, I figured the most used ones should be included by default. Examples include HTTP headers but also:
- CONTENT_LENGTH
- CONTENT_TYPE
- SERVER_NAME
- SERVER_PORT
- HTTP_HOST

Big thanks to @tillkruss who has done most of the work here!


src/Event/Http/Psr7Bridge.php
```diff
@@ -157,4 +163,31 @@ private static function parseKeyAndInsertValueInArray(array &$array, string $key
 
         $pointer = $value;
     }
+    
+    /**
+     * Parse the username and password from the `Authorization` header.
+     * Only "Basic" is supported.
```
mnapoli, line 169:
```suggestion
     * Only "Basic" is supported for now.
```

Just to clarify that this isn't intentional and there is no reason to not change that later.

src/Event/Http/Psr7Bridge.php
```diff
@@ -157,4 +163,31 @@ private static function parseKeyAndInsertValueInArray(array &$array, string $key
...
+     *
+     * @param  array  $headers
+     * @return string[]|null[]
+     */
+    protected static function parseBasicAuthorization(array $headers)
```
mnapoli, line 174:
```suggestion
    protected static function parseBasicAuthorization(array $headers): array
```

src/Event/Http/Psr7Bridge.php
```diff
@@ -157,4 +163,31 @@ private static function parseKeyAndInsertValueInArray(array &$array, string $key
...
+    {
+        $authorization = trim($headers['authorization'][0] ?? '');
```
mnapoli, line 176: Just to confirm, at this point all headers are forced to be in lowercase? (not risk of getting `Authorization` instead?)

src/Event/Http/Psr7Bridge.php
```diff
@@ -157,4 +163,31 @@ private static function parseKeyAndInsertValueInArray(array &$array, string $key
...
+        $auth = base64_decode(trim(explode(' ', $authorization)[1]));
+
+        if (! $auth || ! strpos($auth, ':')) {
+            return [null, null];
+        }
+
+        return [
+            strstr($auth, ':', true),
+            substr(strstr($auth, ':'), 1),
+        ];
```
mnapoli, line 191: Nitpick, but could this be simplified to an `explode` call with a limit of `1`?

src/Event/Http/Psr7Bridge.php
```diff
@@ -157,4 +163,31 @@ private static function parseKeyAndInsertValueInArray(array &$array, string $key
 
         $pointer = $value;
     }
+    
+    /**
+     * Parse the username and password from the `Authorization` header.
+     * Only "Basic" is supported.
+     *
+     * @param  array  $headers
```
mnapoli, line 171: This line can be removed as it is redundant with the code.

mnapoli (commented):
Looks great, thank you!

Would you be able to add tests?

- 1 testing the basic auth variables
- 1 testing the content type

src/Event/Http/Psr7Bridge.php
georgeboot: Jup: https://github.com/brefphp/bref/blob/12afd8a49ceb97c2a073b1764ba01cf3efb25e87/src/Event/Http/HttpRequestEvent.php#L298

src/Event/Http/Psr7Bridge.php
```diff
@@ -23,22 +25,32 @@ final class Psr7Bridge
...
+            'PHP_AUTH_USER' => $user,
+            'PHP_AUTH_PW' => $password,
+        ], fn ($value) => ! is_null($value));
 
-        $headers = $event->getHeaders();
-        if (isset($headers['Host'])) {
-            $server['HTTP_HOST'] = $headers['Host'];
+        foreach ($headers as $name => $values) {
+            $server['HTTP_' . strtoupper(str_replace('-', '_', $name))] = $values[0];
         }
```
georgeboot, line 54: Some less modern php apps expect headers to be available via `HTTP_*` server vars.

src/Event/Http/Psr7Bridge.php
```diff
@@ -23,22 +23,33 @@ final class Psr7Bridge
...
-            'SERVER_PROTOCOL' => $event->getProtocolVersion(),
+        $server = array_filter([
+            'CONTENT_LENGTH' => $headers['content-length'][0] ?? null,
+            'CONTENT_TYPE' => $event->getContentType(),
+            'DOCUMENT_ROOT' => getcwd(),
+            'QUERY_STRING' => $event->getQueryString(),
             'REQUEST_METHOD' => $event->getMethod(),
+            'SERVER_NAME' => $event->getServerName(),
+            'SERVER_PORT' => $event->getServerPort(),
+            'SERVER_PROTOCOL' => $event->getProtocol(),
```
georgeboot, line 39: This was previously set to the protocol version, but that is incorrect.

georgeboot:
@mnapoli added some more fields, but all sourced directly from implemented methods of the HttpRequestEvent.

src/Event/Http/Psr7Bridge.php
```diff
@@ -23,22 +23,33 @@ final class Psr7Bridge
...
+        [$user, $password] = self::parseBasicAuthorization($headers);
 
-        $server = [
-            'SERVER_PROTOCOL' => $event->getProtocolVersion(),
+        $server = array_filter([
+            'CONTENT_LENGTH' => $headers['content-length'][0] ?? null,
+            'CONTENT_TYPE' => $event->getContentType(),
+            'DOCUMENT_ROOT' => getcwd(),
+            'QUERY_STRING' => $event->getQueryString(),
             'REQUEST_METHOD' => $event->getMethod(),
```
tillkruss, line 36: I'd say it wouldn't hurt to also add:

```php
'SERVER_ADDR' => '127.0.0.1',
'SCRIPT_FILENAME' => $_SERVER['_HANDLER'] ?? $_SERVER['SCRIPT_FILENAME'],
'GATEWAY_INTERFACE' => 'FastCGI/1.0',
```

src/Event/Http/Psr7Bridge.php
```diff
@@ -87,7 +98,7 @@ private static function parseBodyAndUploadedFiles(HttpRequestEvent $event): arra
         $parsedBody = null;
         $contentType = $event->getContentType();
         if ($contentType !== null && $event->getMethod() === 'POST') {
-            if ($contentType === 'application/x-www-form-urlencoded') {
+            if (substr($contentType, 0, 33) === 'application/x-www-form-urlencoded') {
```
tillkruss, line 101: Would a `strpos()` not be a bit safer?

src/Event/Http/Psr7Bridge.php
georgeboot: My way of thinking was that these imply FastCGI and/or a local reverse proxy, but that doesn't have to be the case since this is just a PSR-7 bridge.

src/Event/Http/Psr7Bridge.php
georgeboot:
```suggestion
            if (strpos($contentType, 'application/x-www-form-urlencoded') === 0) {
```

tillkruss:
> Is there a specific scenario where these variables are useful to you?

It's really backwards compatibility with 3rd party packages that may access their weirdest variables, because most develop on PHP-FPM.

mnapoli:
Got it, but these work with PSR-7? Or could it be something related to Octane maybe? I'm having trouble seeing where a legacy codebase/package would run with the `function` layer, because that sounds really advanced.

georgeboot:
Yes I tend to agree with you. Octane indeed uses the PSR-7 bridge. Don't think there will be a lot of legacy apps running Octane out there.

Laravel for example, reads the basic auth user from these two variables. And there are plenty Laravel apps using basic auth out there.

And I bet there are some use cases for the other ones as well.

Shall I review all of them (including existing ones) one by one and check if they are used anywhere in Laravel or Symfony?

mnapoli:
> Laravel for example, reads the basic auth user from these two variables. And there are plenty Laravel apps using basic auth out there.

OK that's enough to convince me then 😄  This isn't the best in theory, but this is best in practice and practice trumps theory.

I'm all good for this approach then 👍

src/Event/Http/Psr7Bridge.php
mnapoli: Agreed, specifically these might be overkill and even misleading?

src/Event/Http/Psr7Bridge.php
mnapoli: Beautiful thanks!

tests/Event/Http/Psr7BridgeTest.php
```diff
@@ -31,6 +31,24 @@ public function test I can create a response from a PSR7 response()
         ], $response->toApiGatewayFormat());
     }
 
+    public function test request with basic auth contains a user and password()
```
mnapoli, line 34: Could you add these tests instead to the `CommonHttpTest` class? That way they are applied across function and fpm layers.

Which brings me to another request (sorry!): could you ensure the FPM layer has the same behavior (basic auth)? Because I guess the same problem exists with that layer as well, right?

src/Event/Http/Psr7Bridge.php
```diff
@@ -23,22 +23,33 @@ final class Psr7Bridge
...
             'REQUEST_TIME' => time(),
             'REQUEST_TIME_FLOAT' => microtime(true),
-            'QUERY_STRING' => $event->getQueryString(),
-            'DOCUMENT_ROOT' => getcwd(),
             'REQUEST_URI' => $event->getUri(),
-            'REMOTE_ADDR' => $event->getSourceIp(),
-        ];
+            'PHP_AUTH_USER' => $user,
+            'PHP_AUTH_PW' => $password,
+        ], fn ($value) => ! is_null($value));
```
mnapoli, line 49: I _think_ `array_filter` works without a callback too, and it might do what you want (remove empty values). That would solve the php-cs error as well.

georgeboot: done

tests/Event/Http/Psr7BridgeTest.php
mnapoli: As sorry I meant that the FPM layer should behave the same way as well (`PHP_AUTH_USER` and `PHP_AUTH_PW` defined).

I can try to send a pull request today to add that if you want (PR on your branch).

georgeboot: Sorry, misunderstood you.

Pushed a fix now. If that's now what you meant, please PR :)

mnapoli: 👍  I'm working on a PR to adjust some details in the tests, but just saw the last commit, that should definitely help thanks!

mnapoli: 🤔  actually while adding tests for this behavior, the tests pass without your last commit.

In retrospect, that makes sense now: the FPM process is automatically processing the HTTP Auth header and adding the variables manually. So we don't have to add code to support that. I'll remove those lines in my PR.

mnapoli: OK we should be good, the PR is here: https://github.com/georgeboot/bref/pull/1

mnapoli (approved):

mnapoli:
All good for me, it seems there are two failures in tests that might be a matter of a `?? ''` or something similar.

georgeboot:
> All good for me, it seems there are two failures in tests that might be a matter of a `?? ''` or something similar.

Fixed now

mnapoli:
Tests are failing on PHP 7.3 because of `str_starts_with`

georgeboot:
Ugh missed that. Fixed now.

You probably need allow Actions run once again

mnapoli:
🙌  thank you @georgeboot!

PR #1326, opened 2022-11-16
Remove `separateVendor` feature for Bref v2

### LINKED ISSUE #1318: Get rid of `separateVendor: true` for Bref v2
I am considering removing the `separateVendor: true` feature for Bref 2.0 (https://bref.sh/docs/environment/php.html#separate-vendor-folder).

Motivation:

- this feature was contributed a while ago but it's never been maintained, it adds a lot of technical debt and bugs (e.g. last one is #1311)
- I'm not sure it is used a lot
- I'd actually recommend deploying using Docker images instead, as this provides flat cold starts (better performances)

Any opinions here? Anyone using it?

mnapoli:
Fixes #1318

### DIFF
--- a/docs/environment/php.md
+++ b/docs/environment/php.md
@@ -177,17 +177,3 @@ provider:
 ```
 
 The path must start with `/var/task`, which is the directory where projects are installed on AWS Lambda.
-
-## Separate vendor folder
-
-Bref has the ability to shrink the Lambda deployment archive by separating the vendor folder from the applications code and injecting it later on into the Lambda function. This allows Bref to circumvent Lambda file size limitations.
-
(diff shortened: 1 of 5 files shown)


GrahamCampbell:
Maybe we could have better DX around when the vendor folder is too large, or indeed anything is too large. Maybe someone accidentality puts 40MB of test stubs in the lambda, for example.

mnapoli:
Indeed, that's a good idea. Improving error messages needs to be done in the serverless codebase, it's slightly out of scope today for me.

But I have pushed improvements to the storage and Docker documentation. Hopefully these will make a bit more sense now.

PR #1331, opened 2022-11-23
Support Europe/Spain region

### LINKED ISSUE #1332: Support Europe/Spain region
Recently a new AWS region was launched in Europe/Spain
I've opened a PR to add support for "eu-south-2" and be able to push layers to that region also

kevincerro:
Closes #1332 

Recently a new AWS region was launched in Europe/Spain
This PR adds support for "eu-south-2" and pushes layers to that region also

### DIFF
--- a/runtime/layers/regions.json
+++ b/runtime/layers/regions.json
@@ -18,6 +18,7 @@
     "ap-southeast-1",
     "ap-southeast-2",
     "eu-south-1",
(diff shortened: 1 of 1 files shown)


mnapoli:
Thank you! Are you able to send a similar PR to https://github.com/brefphp/aws-lambda-layers (Bref v2 layers) as well?

PR #1343, opened 2022-12-08
Add PHP 8.2 to the CI

mnapoli:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
         timeout-minutes: 15
         strategy:
             matrix:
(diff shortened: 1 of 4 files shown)


shouze:
@mnapoli you will also need #1344.
Also, what about making a PR that removes php 7.x support? :angel:

mnapoli:
Oh I'm just noticing your PR after working on fixing PHP 7.3 and adding 8.2 dev image. But it seems I missed some part, your PR will be useful 😄

For now I'll keep PHP 7.x in Bref 1, it's dropped in Bref 2. There are still 25% of users on 7.4 so I think it's good if we can keep having Bref 1 supporting 7.4.

Let me merge both PRs.

Issue #1043, opened 2021-09-29
ARM64/Graviton support


jasonmccallister:
AWS launched ARM64 support on Lambda (https://aws.amazon.com/blogs/aws/aws-lambda-functions-powered-by-aws-graviton2-processor-run-your-functions-on-arm-and-get-up-to-34-better-price-performance/).

Would be useful to get Bref functions and container images setup with ARM64 variants... Happy to assist where I can.

mnapoli:
AFAIK PHP isn't running best on ARM (e.g. JIT not supported). I may be wrong, but it might be worth digging into that?

deleugpn:
I think the first step is to wait Cloudformation support? https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/926

jasonmccallister:
@deleugpn for Lambda layers yes, but I think there should not be anything holding up pushing the ARM images to Docker Hub for container lambda functions?

dytyniuk:
@deleugpn people claim in https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/926#issuecomment-931785524 that CloudFormation API already supports the `Architecture` property.

@mnapoli maybe, this helps to melt your concern about running PHP on ARM -- https://aws.amazon.com/tr/blogs/compute/improving-performance-of-php-for-arm64-and-impact-on-amazon-ec2-m6g-instances/

mnapoli:
@dytyniuk definitely, thank you. I was wrong, ARM can be interesting for PHP.

Also relevant: https://twitter.com/sebastienhouze/status/1443742924534132843

> And that ARM JIT support is coming at least with 8.1

Also FYI: CloudFormation support is in (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html#cfn-lambda-function-architectures) and we're planning to add support in Serverless Framework early next week.

willvincent:
So if the roadblocks to this were cloudformation and the serverless framework supporting arm - both of which now do.. is there any remaining issue or is this good to go now? And if so.. how so? 😁

I'm about to dig into a new project and I think I'd like to use the graviton lambdas if possible.

dytyniuk:
Hey @deleugpn 
Maybe, this repository can help -- https://repo.drpixel.fr/enterprise/7/drpixel-php74/aarch64/ ?

deleugpn:
[...]

PR #1360, opened 2022-12-15
Support PHP 8.1 for ARM layers

mnapoli:
Follow-up of https://github.com/brefphp/aws-lambda-layers/pull/40

Ref #1043

### DIFF
--- a/docs/runtimes/README.md
+++ b/docs/runtimes/README.md
@@ -106,19 +106,21 @@ functions:
         handler: public/index.php
 +        architecture: arm64
         layers:
--            - ${bref:layer.php-80-fpm}
-+            - ${bref:layer.arm-php-80-fpm}
+-            - ${bref:layer.php-81-fpm}
++            - ${bref:layer.arm-php-81-fpm}
         // ...
 ```
 
 Bref ARM layers are the same as the x86 layers, but the `arm-` prefix in their name:
 
+- `${bref:layer.arm-php-81}`
 - `${bref:layer.arm-php-80}`
+- `${bref:layer.arm-php-81-fpm}`
 - `${bref:layer.arm-php-80-fpm}`
 - `${bref:layer.console}` (this is the same layer for both x86 and ARM)
 
 > **Warning**
-> ARM runtimes are built from Amazon Linux Extra. **Only PHP 8.0 is available** for now.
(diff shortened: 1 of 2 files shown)


aran112000:
The PHP 8.1 ARM FPM layers appear to be missing cURL so don't actually run correctly at the moment:

```
Fatal error: Uncaught Error: Call to undefined function Bref\Runtime\curl_init() in /opt/bref/php-fpm-runtime/vendor/bref/bref/src/Runtime/LambdaRuntime.php:111
Stack trace:
#0 /opt/bref/php-fpm-runtime/vendor/bref/bref/src/Runtime/LambdaRuntime.php(80): Bref\Runtime\LambdaRuntime->waitNextInvocation()
#1 /opt/bref/php-fpm-runtime/vendor/bref/php-fpm-runtime/src/bootstrap.php(41): Bref\Runtime\LambdaRuntime->processNextEvent(Object(Bref\FpmRuntime\FpmHandler))
#2 {main}
thrown in /opt/bref/php-fpm-runtime/vendor/bref/bref/src/Runtime/LambdaRuntime.php on line 111
```

mnapoli:
Oh interesting, it seems our tests are not catching it: https://github.com/brefphp/aws-lambda-layers/blob/main/tests/test_2_extensions.php#L8

I'll investigate that, thanks @aran112000!

mnapoli:
@aran112000 the new beta was kind of broken actually, I pushed a new beta to fix this (tested ARM PHP 8.1): https://github.com/brefphp/bref/releases/tag/2.0.0-beta3

Thank you!

aran112000:
You're a star - thanks so much for diagnosing and getting a new release pushed out so quickly!! Just rolled out an internal service on it as a test and it's working perfectly (PHP 8.1 on Graviton) 🚀

mnapoli:
> Just rolled out an internal service on it as a test and it's working perfectly (PHP 8.1 on Graviton) 🚀

That is awesome feedback, thank you for sharing!

aran112000:
The performance of these new ARM layers on Bref 2 is insane! We've upgraded another workload this afternoon and here are the performance results looking at the total Lambda duration (Orange: AVG, Blue: MIN):

image

If you can't guess, the changes rolled out 13:35. 

What's more crazy, is we reduced the allocated memory from 512 to 360 as well! 🚀

mnapoli:
@aran112000 wooow that's awesome!

PR #1379, opened 2023-01-17
Chore: Update command documentation to not use PHP in docker-compose run

ryanolee:
# What
Removes `php` from the `docker-compose run console` documentation

# Why
The example given suggests to run 
` docker-compose run console php bin/console`

Given the entrypoint is already `php` in the docker-compose.yml example given above.
The `php` in the `Then commands can be run via:` section causes a `Could not open input file: php` error.


mnapoli:
Thank you!

PR #1383, opened 2023-01-27
Make sure we support query strings with period in them

Nyholm:
This PR will pick up the work from @DaMitchell in #769. 

I used all existing tests and added a few new. I took inspiration from [this method](https://github.com/crwlrsoft/url/blob/v2.0.2/src/Helpers.php#L308-L330) from https://github.com/crwlrsoft/url by @otsch. 
I decided not to use that package as a dependency for two reasons: 
1. We only needed one small method. 
2. It wasn't 100% perfect

Im pinging Otsch so they can see what I did and maybe implement those changes in `crwlrsoft/url`. (Sorry for not making a PR to your package too). I did add your copyright text and credit in the method description. 

The main thing is that I wanted to support query strings with both "a_b" and "a.b".

### DIFF
--- a/src/Event/Http/HttpRequestEvent.php
+++ b/src/Event/Http/HttpRequestEvent.php
@@ -171,8 +171,7 @@ final class HttpRequestEvent implements LambdaEvent
 
     public function getQueryParameters(): array
     {
-        parse_str($this->queryString, $query);
-        return $query;
+        return $this->queryStringToArray($this->queryString);
     }
 
     public function getRequestContext(): array
@@ -225,7 +224,7 @@ final class HttpRequestEvent implements LambdaEvent
             $queryString = $this->event['rawQueryString'] ?? '';
             // We re-parse the query string to make sure it is URL-encoded
             // Why? To match the format we get when using PHP outside of Lambda (we get the query string URL-encoded)
-            parse_str($queryString, $queryParameters);
+            $queryParameters = $this->queryStringToArray($queryString);
             return http_build_query($queryParameters);
         }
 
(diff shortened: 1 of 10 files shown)


otsch:
Hey, thx for the ping @Nyholm! 🙏🏻 I'll have a look at it and of course take your improvements over to my package 👍🏻
If you want, you can also remove the license text, the hint that it's inspired by the method in my package is totally fine 👍🏻 but maybe you want to give my url package repo a star? 😅

Btw. I also once opened an issue in the PHP src repo about this: https://github.com/php/php-src/issues/8639

Nyholm:
I just now found #1042. I added tests from it and make sure we dont change the current behaviour. 

> Hey, thx for the ping @Nyholm! 🙏🏻 I'll have a look at it and of course take your improvements over to my package 👍🏻
If you want, you can also remove the license text, the hint that it's inspired by the method in my package is totally fine 👍🏻 but maybe you want to give my url package repo a star? 😅

Thank you. Sure, of course!

> Btw. I also once opened an issue in the PHP src repo about this: https://github.com/php/php-src/issues/8639

Yes, I saw your comments there. I 100% agree with you. I event feel compelled to make an RFC with a `parse_query_string()` method that works as you describe.

otsch:
> Yes, I saw your comments there. I 100% agree with you. I event feel compelled to make an RFC with a `parse_query_string()` method that works as you describe.

That would be great. Tell me if I can support you with that. I must admit that I didn't really have the confidence to create an RFC 🙈

otsch:
@Nyholm as I already extracted the logic for query strings to a separate package a while ago, and the url package also already used it in its `Url::queryString()` method, I now also replaced the usage of the simpler helper for the `Url::queryArray()` method by using the `Query` class from the separate package. (https://github.com/crwlrsoft/url/pull/40/files#diff-951296e7f45f3adb0f3e0975964c423ed4dcd74743499d1a3159ede66024d4cdL147) 

The query-string package does it a little differently: https://github.com/crwlrsoft/query-string/blob/main/src/Query.php#L657 
I just added all the test cases from this pull request to the tests of this package: https://github.com/crwlrsoft/query-string/pull/3/files#diff-bd359578b78f3980ae6d0e8b8f6ca9591ea8eded2e9ef474f55c61cf9f8cef59R71 
So, if you maybe want to use this package instead now, at least the examples for dots and spaces in keys should work.

mnapoli:
Thanks for tackling this, that's awesome ❤️

And thanks @otsch for the updates, https://github.com/crwlrsoft/query-string looks awesome, well set-up, well tested, and a single-class with no dependencies -> it would make sense for us to switch to using your package now, wdyt @Nyholm? (that's one less complexity to have in here)

otsch:
> it would make sense for us to switch to using your package now, wdyt @Nyholm? 

In case you decide to do so, we can also have a look and add all query related test cases from this repo to my package. And knowing it's used in this project, I'd then definitely be very careful with potentially breaking changes! 👍🏻

PR #1389, opened 2023-01-31
Enable faster deployments by default

mnapoli:


tillkruss:
Huh. What does this do?

GrahamCampbell:
Doesn't use change sets.

PR #1394, opened 2023-02-01
Allow setting PHP as a runtime in serverless.yml

mnapoli:
TODO:

- [x] Gather feedback from the community
- [x] Update docs
- [x] Update templates

---

This new feature allows setting PHP as a runtime when deploying with `serverless.yml`:

```yml
service: myapp

provider:
    name: aws

functions:
    function:
        handler: function.php
        runtime: php-81

    api:
        handler: api.php
        runtime: php-81-fpm
        url: true


plugins:
    - ./vendor/bref/bref
```

This replaces the `layers` variables (**which still work!**). One could migrate to this new simpler syntax like this for example:

```diff
service: myapp

provider:
    name: aws
-    runtime: provided.al2

functions:
    function:
        handler: function.php
-        layers
-            - ${bref:layer.php-81}
+        runtime: php-81

    api:
        handler: api.php
-        layers
-            - ${bref:layer.php-81-fpm}
+        runtime: php-81-fpm
        url: true

plugins:
    - ./vendor/bref/bref
```

This [removes several wtfs for beginners](https://www.osnews.com/story/19266/wtfsm/):

- wtf is "provided.al2"
- wtf are "layers"
- wtf is that `${bref:layer.php-81-fpm}` syntax
- wtf do I need 2 layers for console

Regarding the last one, indeed things get simpler for `console`:

```diff
functions:
    console:
        handler: bin/console
-        layers
-            - ${bref:layer.php-81}
-            - ${bref:layer.console}
+        runtime: php-81-console
```

And also it makes switching to ARM much simpler too because the runtime stays the same, the layer is automatically switched to the ARM layer by the Bref plugin (it detects `architecture: arm64`):

```yml
functions:
    console:
        handler: bin/console
        architecture: arm64
        runtime: php-81 # no need to add the `arm-` prefix
```

### DIFF
--- a/docs/environment/php.md
+++ b/docs/environment/php.md
@@ -119,10 +119,7 @@ extension=pdo_pgsql
 
 ### Extra extensions
 
-Due to space limitations in AWS Lambda, Bref cannot provide every possible extension.
-There are a list of additional PHP extensions that can be installed as a separate
-layer. They are less common extensions or extensions that for some reasons should
-not be in the normal Bref layer.
+Due to space limitations in AWS Lambda, Bref runtimes cannot include every possible PHP extensions. These additional PHP extensions can be included as separate AWS Lambda layers.
 
 All extra PHP extensions are found in [brefphp/extra-php-extensions](https://github.com/brefphp/extra-php-extensions).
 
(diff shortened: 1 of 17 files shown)


afu-dev:
Feedback: less things to write = less error possible. Also it feels nicer to read/understand.

Cons: runtime and architecture may be flagged as "error" or "unknown" when using an IDE. Example with PhpStorm

image

georgeboot:
Only thing I could really see is that newcomers might also want additional layers (isn't intl going to be a separate layer) and now you are figuring out why there are 2 ways to declare stuff.

But overall and assuming that this os properly documented, I'm all for!

mnapoli:
@afu-dev agreed, this is annoying to me as well.

@georgeboot regarding extensions, I was thinking about something like this too:

```yaml
functions:
    hello:
        handler: ...
        runtime: php-81
        extensions:
            - redis
            - amqp
```

Would that be better?

(and yes we would need a clear error message in case users hit the layer limit)

georgeboot:
> Would that be better?

Yup amazing!

mnapoli:
Oh yes definitely!

mnapoli:
Note: will look into extensions separately.

PR #1395, opened 2023-02-02
Enable faster deployments by default

mnapoli:
I set this manually in every project:

```yaml
provider:
  name: aws
  deploymentMethod: direct
```

Why not have Bref set it by default?

This PR enable the following feature by default: https://www.serverless.com/framework/docs/providers/aws/guide/deploying#deployment-method (unless the user set it explicitly in `serverless.yml`)

That makes full deployments 2 times faster or more.

See https://github.com/serverless/serverless/issues/10815 for some numbers.

---

TL/DR about the underlying issue: Serverless Framework v3 deploys with CloudFormation *changesets* enabled to show a nice progress count while the deployment progresses.

But changesets, unfortunately, make the deployment MUCH longer for no apparent reason. That's unfortunate. We introduced an option in `serverless.yml` to opt-out of these slower deployments, with the goal of making this the default for Serverless Framework v4.

The idea here is to enable this option by default with Bref. The only downside is that we lose the interactive progress, specifically the number of resources deployed in real time 🤷

### DIFF
--- a/index.js
+++ b/index.js
@@ -21,6 +21,12 @@ class ServerlessPlugin {
             throw new serverless.classes.Error('Bref requires Serverless Framework v3, but an older v2 version is running.\nPlease upgrade to Serverless Framework v3.');
         }
 
+        // Automatically enable faster deployments (unless a value is already set)
+        // https://www.serverless.com/framework/docs/providers/aws/guide/deploying#deployment-method
+        if (! serverless.service.provider.deploymentMethod) {
+            serverless.service.provider.deploymentMethod = 'direct';
+        }
+
         const filename = path.resolve(__dirname, 'layers.json');
         const layers = JSON.parse(fs.readFileSync(filename).toString());
 


GrahamCampbell:
Hmmmm, I don't feel like a plugin should be messing with that, and I prefer using change sets to deploy.

mnapoli:
@GrahamCampbell I'm interested, why do you use changesets with serverless framework specifically?

GrahamCampbell:
I don't particularly care either way. They were just the default in v3, and I didn't change it. I would object to us overwriting it though. If we want people to get the "best" option by default, serverless framework v4 should be released I guess, or maybe just all the bref docs explicitly sets the deployment method, so copy-paste examples are already using it.

mnapoli:
I worked on adding that flag in SF v3, not making it the default was a mistake to be honest (we had long debates).

SF v4 will use it, but is not coming anytime soon unfortunately. And there are no know downsides to skipping changesets (which was the default in v2). Even AWS CDK, SST, and other tools are going in that direction because of the performance issue with changesets.

It really feels like a cheap way to make deployments 2 times faster (or often more). If I had the choice, I'd set that by default that in SF directly.

I'll be bold and move forward with this, even though it does feel like overstepping boundaries between software.

Issue #1396, opened 2023-02-05
Make handlers more easily testable by removing `final` keyword


mnapoli:
Reported by @marcguyer on Slack.

### Problem

Typed events (like `SqsEvent`) and `Context` are all `final`.

That was so that we could add any method without fearing breaking BC.

However, it makes mocking in tests very hard. You have to do create the payload manually, for example:

```php
$handler->handleSqs(
		$sqsEvent = new SqsEvent(['Records' => [[
			'eventSource' => 'aws:sqs',
			'body' => '{"eventRequestId":999,"startingState":"event_planned","targetState":"event_canceled"}',
		]]]),
		new Context('', intval(microtime(true) * 1000) + 6000, '', '')
);
```

Removing `final` would allow using `$this->createMock()`.

### Proposed solution

Remove `final` on these classes, but **document** in phpdoc that these classes should not be extended with new methods (no BC promise). They are "final" in spirit, but not in practice to allow mocking.

mnapoli:
Alternative approach: adding helper methods to quickly create instances without having to create a fake Lambda payload by hand.

Here's an example with just the `Context` class: https://github.com/brefphp/bref/pull/1398

mnapoli:
Note: the approach above is great for `Context` I think, but is much more work to apply (and apply consistently) across all typed events.

It turns out `@final` is enforced by Phpstan and Psalm (https://phpstan.org/r/80a438f2-775f-4c79-b0e0-624bd17dd4f2), so relying on this annotation and removing `final` could work!

Slamdunk:
The classes you mention here should never be mocked. Mocks serve a controlled and manipulated behaviour. Those classes are, as far as I can tell, value objects therefore with a behavior that do not rely on external nor internal states.

Fake helpers are the correct way to go.

`@final` keyword enforced by SA works, but hides a technical debt under the carpet: fake helpers should be easy to create and maintain; if they aren't, I'd say there's a smell somewhere in the code architecture.

mnapoli:
@Slamdunk I'm not really interested in debating generic theoretical ideas, because in my experience these are never useful to solve a specific problem.

I'm more interested in solving the problem here specifically. And the problem is, these typed handlers are hard to write tests for.

[...]

PR #1398, opened 2023-02-05
Facilitate testing handlers by adding `Context::fake()`

mnapoli:
This is an alternative approach to solve #1396, but just fo the `Context` class. It implies adding `::fake()` to facilitate creating instances.


mnapoli:
I won't apply this to all events because it would be too much work (to create relevant data).

But for the Context class it makes a lot of sense (instead of resorting to mocks).

PR #1399, opened 2023-02-05
Introduce SqsEventFactory

bambamboole:
@mnapoli , You asked on [Twitter](https://twitter.com/matthieunapoli/status/1622258374837420032), if there is an elegant way, to mock the Event classes, but keep the final.
data structures normally do not have to be mocked, since they are just data structures. If the construction of them is too complex, it should get encapsulated in a factory.

I prepared and example, which also showcases, that the test responsibility moved too the FactoryTest and leaves the EventTest small.

In Tests, you can easily have a helper, for these Events. We at work have for example a trait for tests which is called like `ProvidesFakeEvents` and this has then methods like `createRandomValidEvent()`.
Here a code example:
```php
<?php declare(strict_types=1);

namespace Bref\Test\Traits;

use Bref\Event\Sqs\SqsEvent;
use Bref\Event\Sqs\SqsEventFactory;

trait ProvidesFakeEvents
{
    protected function getValidSqsEvent(): SqsEvent
    {
        $rawEvent = json_decode(file_get_contents(dirname(__DIR__) . '/Event/Sqs/sqs.json'), true);

        return (new SqsEventFactory())->createFromPayload($rawEvent);
    }
}

```
It is absolutely fine to use such a factory for static data in you test helpers, since it is also covered by unit tests.

Happy to hear you opinion on this.

### DIFF
--- a/src/Event/Sqs/SqsEventFactory.php
+++ b/src/Event/Sqs/SqsEventFactory.php
@@ -1,0 +1,33 @@
+<?php declare(strict_types=1);
+
+namespace Bref\Event\Sqs;
+
+use Bref\Event\InvalidLambdaEvent;
+use InvalidArgumentException;
+
+final class SqsEventFactory implements SqsEventFactoryInterface
+{
+    public function createFromPayload(array $payload): SqsEvent
+    {
+        if (! isset($payload['Records'])) {
+            throw new InvalidLambdaEvent('SQS', $payload);
+        }
+
+        return new SqsEvent(array_map([$this, 'createSqsRecord'], $payload['Records']), $payload);
+    }
+
+    private function createSqsRecord(array $payload): SqsRecord
+    {
+        if (! isset($payload['eventSource']) || $payload['eventSource'] !== 'aws:sqs') {
+            throw new InvalidArgumentException;
+        }
+        return new SqsRecord(
+            $payload['messageId'],
+            $payload['body'],
+            $payload['messageAttributes'],
+            (int) $payload['attributes']['ApproximateReceiveCount'],
+            $payload['receiptHandle'],
+            $payload,
+        );
+    }
+}
--- a/src/Event/Sqs/SqsEventFactoryInterface.php
+++ b/src/Event/Sqs/SqsEventFactoryInterface.php
@@ -1,0 +1,8 @@
+<?php declare(strict_types=1);
+
+namespace Bref\Event\Sqs;
+
+interface SqsEventFactoryInterface
+{
+    public function createFromPayload(array $payload): SqsEvent;
+}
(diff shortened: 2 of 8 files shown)


src/Event/Sqs/SqsHandler.php
```diff
@@ -10,6 +10,16 @@
  */
 abstract class SqsHandler implements Handler
 {
+    private SqsEventFactoryInterface $eventFactory;
+
+    public function __construct(?SqsEventFactoryInterface $eventFactory = null)
```
mnapoli, line 15: We can't have a constructor, this is a base class for userland code.

src/Event/Sqs/SqsEventFactoryInterface.php
```diff
@@ -0,0 +1,8 @@
+<?php declare(strict_types=1);
+
+namespace Bref\Event\Sqs;
+
+interface SqsEventFactoryInterface
```
mnapoli, line 5: I don't see the use for the interface?

src/Event/Sqs/SqsEventFactory.php
```diff
@@ -0,0 +1,33 @@
...
+
+final class SqsEventFactory implements SqsEventFactoryInterface
+{
+    public function createFromPayload(array $payload): SqsEvent
+    {
+        if (! isset($payload['Records'])) {
+            throw new InvalidLambdaEvent('SQS', $payload);
+        }
+
+        return new SqsEvent(array_map([$this, 'createSqsRecord'], $payload['Records']), $payload);
```
mnapoli, line 16: With this "early parsing" approach, all data of the event will be parsed even if not used. In the context of such a low-level library, I'd rather avoid that and favor lazy loading.

This isn't a deal breaker though, just something I'd avoid.

mnapoli (commented):
> data structures normally do not have to be mocked, since they are just data structures. If the construction of them is too complex, it should get encapsulated in a factory.

I invite you to read https://github.com/brefphp/bref/issues/1396#issuecomment-1418172170 I don't think we should be discussing concrete problems with black and white theoretical solutions.

To clarify as well, I'm not looking at mocking anything. My goal is to make writing tests simpler for end users.

> In Tests, you can easily have a helper, for these Events.

If you mean "end users", then I don't think it's a good approach because it requires all users to create a class. We want to simplify things for end users, not add work for them.

The example you provide of `ProvidesFakeEvents` shows that the new "factory" classes don't add any value too. The `ProvidesFakeEvents` is interesting in itself, but factory classes add way too much complexity here.

It's actually easy to see: try to apply that PR to all classes in `src/Events` and you'll see what I mean. It will create dozens of new classes for no actual benefit.

Anyway, judging by the examples we are discussing an abstract solution but not the real problem reported by a Bref user. I'll be honest, I don't think this will lead to anywhere useful as this is the application of a generic programming pattern. I'll be closing this PR so that we all save some time.

PR #1402, opened 2023-02-07
Exclude .git directory from deployment

jericopulvera:
Due to some repository having a big .git file, this change adds the exclusion of the .git directory to the list of patterns in the deployment process. This also ensures that sensitive version control information is not deployed along with the rest of the code, reducing the size of the deployment package.

### DIFF
--- a/template/http/serverless.yml
+++ b/template/http/serverless.yml
@@ -24,3 +24,4 @@ package:
     patterns:
         - '!node_modules/**'
         - '!tests/**'
+        - '!.git/**'


t-richard:
Did you have a case where it was uploaded ? It is excluded by default by the serverless framework so this should not be needed.

And if it happened, maybe it's an issue inside serverless framework.

See https://www.serverless.com/framework/docs/providers/aws/guide/packaging#patterns

jericopulvera:
I apologize for the confusion. I didn't realize that the serverless framework excluded this by default. Thank you for pointing that out and for the reference to the packaging patterns. I will make sure to review it and keep this in mind for future.

t-richard:
No problem, and thank you for opening the pull request in the first place :slightly_smiling_face:

PR #1411, opened 2023-02-13
Add hook system to make the Bref runtime more extensible

mnapoli:
This new "hook system" allows 3rd party packages or user code to register hooks on:

- startup (i.e. cold start), before anything else runs
- before Lambda invocations

So far, I have seen that being (legitimately) useful twice for low-level integrations:

- Laravel bridge, to run `php artisan make:config` (or run other caching commands) on cold starts -> that could also be useful for Symfony
- for Bref Live (currently at the prototype stage) that lets users edit code live on Lambda (not published yet)

My initial approach was to try and add environment variables to register shell commands to execute, but that is unnecessarily expensive and harder to set up (because environment variables must be set in `serverless.yml`). With this new system, packages can register hooks automatically, just like with `Bref::setContainer()`.

The targets for this new hook system are not end users (at least from what I can see). This is targeted at packages providing framework integrations or enhancing Bref on a low level.

Adding new extension points is something I don't want to do lightly, as it's the best way to cement/add crust to a system, leading to it being very hard to refactor and maintain down the road. However here it seems like a very good solution for framework integrations, and it is very little invasive. So the balance seems ok.

I'll keep this PR open at first to do some implementation tests.

Note: there is no public documentation for this, this is intentional.

### DIFF
--- a/src/Bref.php
+++ b/src/Bref.php
@@ -11,6 +11,10 @@ class Bref
 {
     private static ?Closure $containerProvider = null;
     private static ?ContainerInterface $container = null;
+    private static array $hooks = [
+        'beforeStartup' => [],
+        'beforeInvoke' => [],
+    ];
 
     /**
      * Configure the container that provides Lambda handlers.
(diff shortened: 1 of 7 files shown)


georgeboot:
This seems to depend on the composer autoloader to load php files and register hooks in that way.

Is it use case where one wants to run things before PHP or at least at a lower level? My initial thought was to add something to `php/bootstrap.d/` where you can place bash scripts or something. Not sure if there is even a shell in Lambda though.

If SnapStart or similar ever comes to php, you probably need lower level to pre-cache things, right?

mnapoli:
Oh interesting idea. But what would be the specific use case for this?

Putting things in `php/bootstrap.d` can only be done by end users though, so it's not something we can use for the Laravel bridge.

SnapStart would be compatible (AFAIK) with the current solution, as the whole initialization phase is cached.

But I need to note here as well that after using this feature more widely today, I have found a limitation. It isn't good as it is, and it has implications on other parts in Bref. I've been looking at solutions, so this isn't the end of it yet :p

mnapoli:
@georgeboot oh I forgot to mention: I added a similar system in Bref v1 a while ago (via the `EXPERIMENTAL_AWS_LAMBDA_EXEC_WRAPPER` environment variable), and I used it in place of these new hooks. With this technique, the extension indeed starts before PHP.

I never heard of anyone else using this, and in my case running even before PHP starts was not that useful.

georgeboot:
Good point that the bridge etc can't use it.

Regarding SnapStart: if they ever comes available I will want to run additional tasks during boot, like pre-compile all views, warm other caches etc etc. Need some place to put that all.

Out of curiosity, what was the limitation you found?

mnapoli:
> Regarding SnapStart: if they ever comes available I will want to run additional tasks during boot, like pre-compile all views, warm other caches etc etc. Need some place to put that all.

👍 AFAICT this would work with this PR right?

> Out of curiosity, what was the limitation you found?

I feel quite stupid tbh, I worked all day on this and missed one critical thing which is that with Bref v2, the autoloader for the FPM runtime is separated from the userland autoloader (intentionally). So… that just doesn't work.

I have a more detailed recap and info, as all this is starting to make me rethink that autoloader/package separation. I'll be working on this tomorrow so I'll post an update.

mnapoli:
@georgeboot here's a follow-up with more details:

https://github.com/brefphp/bref/pull/1413

PR #1419, opened 2023-02-19
Update usage examples to a current version of PHP

nealio82:
The usage examples showed using the PHP 7.4 PHP-FPM layer, but 7.4 has been unsupported since Nov 2022


mnapoli:
Thanks!

PR #1425, opened 2023-02-21
Lighter Bref v2 by moving secrets loading to an optional dependency

### LINKED ISSUE #1409: Make bref 2 even lighter by making `async-aws/ssm` optional
Move dep to suggest and adjust code to to do a class_exists check and crash out with good DX if someone tries to resolve an env variable via ssm without installing the dep (https://github.com/brefphp/bref/blob/2.0.0-beta11/src/Secrets/Secrets.php#L90).

mnapoli:
Fix #1409

Instead of making `async-aws/ssm` an optional dependency (which is confusing for users as they must install a low-level package that is not directly related to the feature they use), I created a `bref/secrets-loader` package: https://github.com/brefphp/secrets-loader

This new package is an optional dependency for users of the `bref-ssm:xxx` feature.

As a result, we gain a lot in the final Composer package size:

- before:
Screen-000985

- after:
Screen-000984

When a user uses the `bref-ssm:` feature without the optional dependency installed, they get:

- a CLI warning when deploying
- an error in Lambda

I added the CLI warning to let users catch this as soon as possible (before deploying). Here is what it looks like:

Screen-000987


mnapoli:
I recorded a video while working on this, in case anyone's interested: https://www.youtube.com/watch?v=mdPVzNJbKAw

composer.json
```diff
@@ -29,13 +29,13 @@
...
         "nyholm/psr7": "^1.2",
-        "psr/container": "^1.0|^2.0",
-        "async-aws/ssm": "^1.3"
+        "psr/container": "^1.0|^2.0"
     },
     "require-dev": {
         "aws/aws-sdk-php": "^3.172",
         "async-aws/core": "^1.0",
         "async-aws/lambda": "^1.0",
+        "bref/secrets-loader": "dev-main",
```
mnapoli, line 38: TODO

PR #1437, opened 2023-03-03
Better support for querystrings (qs)

### LINKED ISSUE #756: Value of $_SERVER['QUERY_STINRG] different when deployed

sladg:
PR based on #1383. Credits to @Nyholm.

This pull requests removed reliance on build-in `parse_str` function in favour of @otsch's https://github.com/crwlrsoft/query-string.

**Allows to use dots and dashes in query parameters (won't covert them to underscores as `parse_str` does).**

Fixes #756.

Makes #1042 and #1383 redundant.

---

Extra reading on dealing with `parse_str`:
- https://github.com/php/php-src/issues/8639
- https://gist.github.com/OldStarchy/1131111bc921b9f09156e64613bd2a32
- https://github.com/crwlrsoft/url/issues/2
- https://github.com/symfony/symfony/issues/9009


sladg:
@mnapoli I picked up on @Nyholm's work and switched to 3rd party solution. Let me know if this is up to the standards or some work is necessary on my end. Thanks!

composer.json
```diff
@@ -56,7 +57,7 @@
         "static-analysis": [
             "vendor/bin/phpstan analyse"
         ],
-        "tests": [
+        "unit-tests": [
```
Nyholm, line 60: Hm.. Generally this is not appreciated by the maintainers.

Is there any reason you decided to change this?

src/Event/Http/HttpRequestEvent.php
```diff
@@ -346,4 +342,17 @@ public function isFormatV2(): bool
...
+     * When keys within a URL query string contain dots, PHP's parse_str() method
+     * converts them to underscores. This method works around this issue so the
+     * requested query array returns the proper keys with dots.
+     *
+     * @return array<string, string>
+     */
+    private function queryStringToArray(string $query): array
+    {
+        $queryString = Query::fromString($query);
+        return $queryString->toArray();
```
Nyholm, line 356: We could maybe make this easier
```suggestion
        return Query::fromString($query)->toArray();
```

Nyholm (approved):
Lovely. 

Thank you. I got stuck with other thing and it slipped my mind. 
Im happy with the PR, I did two small comments.

composer.json
mnapoli: 👍 it would be best to avoid changing lines that are not related to the pull request

sladg: My bad, left it by accident. Anyways, reference `@unit-tests` does not correlate to `"tests"` so it's not started with rest of scripts.
Reverting change.

Nyholm (approved):
Thank you

sladg:
Issue opened to address failing pipeline: https://github.com/crwlrsoft/query-string/issues/4

otsch:
@sladg Can I maybe help bringing it to Bref v2 as discussed in https://github.com/crwlrsoft/query-string/issues/4?

sladg:
@mnapoli rebased, we should be good to go :)

composer.json
```diff
@@ -29,7 +29,8 @@
         "riverline/multipart-parser": "^2.0.6",
         "psr/http-server-handler": "^1.0",
         "nyholm/psr7": "^1.4.1",
-        "psr/container": "^1.0|^2.0"
+        "psr/container": "^1.0|^2.0",
+        "crwlr/query-string": "^1.0"
```
GrahamCampbell, line 33: Do we want to alphabetically sort the package requirements?

GrahamCampbell (changes requested):

composer.json
GrahamCampbell: I think we should also go with ^1.0.1` here.

otsch: For the latter: Yes, I think the changes from 1.0.1 are pretty important, as I actually copied over the test cases from https://github.com/brefphp/bref/pull/1383 and fixed/changed a few things for that. 
But I also don't know if there's any scenario where composer would now install 1.0.0, as there are no dependencies that could conflict and nothing was changed in composer.json for 1.0.1 🤔 but maybe better be safe than sorry 👍🏻

Btw. just to mention it: maybe changes from 1.0.1 could be considered breaking. I also thought about tagging 2.0.0 but went for 1.0.1 then. But I'll definitely be very careful with changes and stick to semver very strictly when bref now uses it! 🫡

GrahamCampbell: Yes, best to insist on at least 1.0.1, if the code at 1.0.0 does not work.

composer.json
```diff
@@ -4,7 +4,7 @@
     "type": "project",
     "description": "Bref is a framework to write and deploy serverless PHP applications on AWS Lambda.",
     "homepage": "https://bref.sh",
-    "keywords": ["bref", "serverless", "aws", "lambda", "faas"],
+    "keywords": [ "bref", "serverless", "aws", "lambda", "faas" ],
```
GrahamCampbell, line 7: unrelated change shouldn't be included.

GrahamCampbell (changes requested):

mnapoli:
All good, thank you for this beautiful work to everyone involved!

aran112000:
Seems this PR introduced a bug for a rare edge case; if you have duplicate query string keys, only the final value will be passed through, these aren't treated as an array as expected.

Just submitted a PR to resolve this here:
https://github.com/crwlrsoft/query-string/pull/5

otsch:
> Seems this PR introduced a bug for a rare edge case;

Correct me if I'm wrong, but I think actually it didn't introduce that behavior? I think it already behaved that way in previous (1.x) versions of bref.

I tried this, with both (v1.7 and v2.0): 
```php
include 'vendor/autoload.php';

use Bref\Event\Http\HttpRequestEvent;

$requestEvent = new HttpRequestEvent([
    'rawQueryString' => 'foo=bar&foo=baz',
    'httpMethod' => 'GET',
    'version' => '2.0',
]);

var_dump($requestEvent->getQueryParameters());

$requestEvent = new HttpRequestEvent([
    'requestContext' => ['elb' => true, 'http' => ['method' => 'GET']],
    'queryStringParameters' => [
        'foo' => 'bar',
        'foo' => 'baz',
    ],
]);

var_dump($requestEvent->getQueryParameters());
```

And in both cases the output is:
```bash
array(1) {
  ["foo"]=>
  string(3) "baz"
}
array(1) {
  ["foo"]=>
  string(3) "baz"
}
```

And looking at the bref source code, I think changing this behavior, would also require a change in bref, because:
* Only when it's about an HTTP 2.0 request, the query params are actually initialized from the actual query string (`foo=bar&foo=baz`) https://github.com/brefphp/bref/blob/master/src/Event/Http/HttpRequestEvent.php#L219 
* In other situations, it is actually initialized from an array of query parameters like `['foo' => 'bar', 'foo' => 'baz']` https://github.com/brefphp/bref/blob/master/src/Event/Http/HttpRequestEvent.php#L230 

So, in the second situation the incoming array would already have to be `['foo' => ['bar', 'baz']]`, because an array with the same key twice isn't possible.

Further, I think it'd best to keep the current behavior and just use a query string like `foo[]=bar&foo[]=baz` in those situations. Generally I think there is no strict rule what `foo=bar&foo=baz` should result in, server implementations can solve it as they want. But I think in most implementations (e.g. laravel) it results in `['foo' => 'baz']`.
Is this OK for you @mnapoli @aran112000 ?

aran112000:
> Correct me if I'm wrong, but I think actually it didn't introduce that behavior? I think it already behaved that way in previous (1.x) versions of bref.

Seems you're right 👍

> And looking at the bref source code, I think changing this behavior, would also require a change in bref, because:

I don't believe that to be the case:
 1. Correct, this does now work with API Gateway V2 with the changes in the PR (this is also the default for Bref these days I believe @mnapoli ?)
 2. In other scenarios (ELB or API GW < 2), we have access to more in the Event payload (`multiValueQueryStringParameters`) which sorts this out from what I've seen in Bref and the tests coverage already there. I did start a PR for Bref earlier thinking that wasn't the case and closed it off after realising it already worked on older versions.

> Further, I think it'd best to keep the current behavior and just use a query string like foo[]=bar&foo[]=baz in those situations. Generally I think there is no strict rule what foo=bar&foo=baz should result in, server implementations can solve it as they want. But I think in most implementations (e.g. laravel) it results in ['foo' => 'baz'].

Completely agree the 'best' structure uses array notation[] on the parameters,  but it's not uncommon to have query strings repeated. Even Google use it for Google Fonts for example when specifying font families - example:
https://fonts.googleapis.com/css2?family=Baloo&family=Roboto

otsch:
> 1. Correct, this does now work with API Gateway V2 with the changes in the PR

Yes, with API Gateway V2 I guess it would work with the changes from the PR, but I don't think it's a good idea if it works differently in different environments. 

> 2. In other scenarios (ELB or API GW < 2), we have access to more in the Event payload (`multiValueQueryStringParameters`) which sorts this out from what I've seen in Bref and the tests coverage already there. I did start a PR for Bref earlier thinking that wasn't the case and closed it off after realising it already worked on older versions.

OK, I just read a bit further on https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html#multi-value-headers and in the comments in bref's source. So, when multi-value is enabled, it reads like the `multiValueQueryStringParameters` already is an array like `['foo' => ['bar', 'baz']]`. The crwlr/query-string is totally fine with that and wouldn't throw away any element, but actually bref converts it 
* first to an array like this: `['foo=bar', 'foo=baz']` in this foreach loop https://github.com/brefphp/bref/blob/master/src/Event/Http/HttpRequestEvent.php#L264 
* and then concatenates this array, back to a query string like `foo=bar&foo=baz`, which is then used to create a crwlr `Query` object instance from a string and use it to convert it to an array again 😅 (here https://github.com/brefphp/bref/blob/master/src/Event/Http/HttpRequestEvent.php#L272)

So, maybe the solution in this case is to just create the crwlr `Query` instance directly from the `multiValueQueryStringParameters` array, because it actually does the same back and forth conversion that bref does, inside the Query object. I'll try this and maybe send a PR.

otsch:
Addition: I'll consider accepting your PR to the crwlr/query-string package @aran112000 👍🏻 have to take a closer look.

aran112000:
> So, maybe the solution in this case is to just create the crwlr Query instance directly from the  array, because it actually does the same back and forth conversion that bref does, inside the Query object. I'll try this and maybe send a PR.

I tried this route, API GW 2 doesn't provide `multiValueQueryStringParameters`, which I why I went this route

otsch:
> I tried this route, API GW 2 doesn't provide `multiValueQueryStringParameters`, which I why I went this route

Yes, that's why I then added, that I consider accepting your PR as well 👍🏻 I'm currently reviewing it...would you mind if I do some minor changes instead of requesting changes?

aran112000:
Gotcha, sure thing. If I can help at all, just shout

otsch:
@mnapoli we changed the behavior in crwlr/query-string => https://github.com/crwlrsoft/query-string/pull/5 
In bref I know see that there's one test case failing with the new version of the package. It's `FpmHandlerLoadBalancerTest::test alb query strings encoded with multi value enabled()`. Is this bad?

otsch:
@mnapoli OK, the tests are succeeding again with crwlr/query-string 1.0.3

otsch:
@mnapoli Yes 👍🏻 and as 1.0.2 only was the latest version for about an hour, not a lot (if any) should have gotten that version.

mnapoli:
FYI I'm planning some changes regarding this in Bref v4: https://github.com/brefphp/bref/issues/2159#issuecomment-5439094957

PR #1442, opened 2023-03-09
Fix bref console bootstrap if using old versions of symfony/process

kevincerro:
If you're using "symfony/process" older than ^4.2 method "fromShellCommandline" doesn't exists.

Captura de pantalla 2023-03-09 a las 13 26 52

I checked that most of the php libraries, like "knp-snappy" or Symfony itself checks if method exists.
Captura de pantalla 2023-03-09 a las 13 28 40
Captura de pantalla 2023-03-09 a las 13 30 37


mnapoli:
Thanks!

Will merge in master (Bref v1), in v2 Symfony 4.2+ is required anyway.

PR #1449, opened 2023-03-14
Upgrade guzzle in dev-deps

GrahamCampbell:
Guzzle 6 is EOL (well, at my discretion - the security policy says support for 2 years after release of v7, which means it has been EOL since June 2022, though I have not yet officially marked it as EOL in the supported versions table).


mnapoli:
Thanks!

PR #1501, opened 2023-04-06
Laravel docs: clarify persistent connections apply to MySQL too

mnapoli:
It wasn't clear whether persistent connections applied to PostgreSQL only or not.

cc @tillkruss, @georgeboot


tillkruss:
I think we should also add MySQL support: https://github.com/laravel/vapor-core/blob/e6a7806d73f8c1cd5bf5ba56aed0838bfa284a35/src/Runtime/Octane/Octane.php#L68-L89

PR #1504, opened 2023-04-12
Fix unit test for multi-valued query string parameters

aran112000:
`multiValueQueryStringParameters` don't always come through as `value[]`, AWS supports "?petType=dog&petType=fish" as an example which would come through as:

```JSON
{
    "multiValueQueryStringParameters": { "petType": [ "dog", "fish" ] }
}
```


mnapoli:
Thanks @aran112000 for looking into all this.

Any reason you closed this one?

PR #1509, opened 2023-04-13
Add test cases for query arrays without brackets

otsch:
Support for getting array structure from duplicate query string keys without array syntax (brackets) was added in crwlr/query-string (v1.0.2) as requested in the discussion of this PR
https://github.com/brefphp/bref/pull/1437.

Example:
foo=bar&foo=baz
becomes
['foo' => ['bar', 'baz']]

Add test case to make sure this actually works (and keeps working) in bref.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
         "php": ">=8.0",
         "ext-curl": "*",
         "ext-json": "*",
-        "crwlr/query-string": "^1.0.1",
+        "crwlr/query-string": "^1.0.3",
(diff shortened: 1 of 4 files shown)


GrahamCampbell (changes requested):
We need to bump the minimum version to `1.0.2`.

otsch:
@GrahamCampbell Sorry, of course. I updated it to min 1.0.3 as it fixes another bug introduced in 1.0.2.

mnapoli:
Thank you!

PR #1516, opened 2023-04-21
Add note for deprecated usage of !php/const:

ramondelafuente:
When linting the yaml files with Symfony yaml 6.2 and higher you will get a deprecation warning when using the original notation. This proposed change makes users aware of this upcoming change.

> Since symfony/yaml 6.2: YAML syntax for key "!php/const:PDO::MYSQL_ATTR_SSL_CA" is deprecated and replaced by "!php/const PDO::MYSQL_ATTR_SSL_CA"

### DIFF
--- a/docs/environment/database-planetscale.md
+++ b/docs/environment/database-planetscale.md
@@ -135,6 +135,8 @@ doctrine:
             # Connect to the database via SSL
             !php/const:PDO::MYSQL_ATTR_SSL_CA: /opt/bref/ssl/cert.pem
 
+            # Note: When using symfony/yaml 6.2 and up it should be: (there is no colon after "const")
+            !php/const PDO::MYSQL_ATTR_SSL_CA: /opt/bref/ssl/cert.pem
 # ...
 ```
 
--- a/docs/environment/database-planetscale.md
+++ b/docs/environment/database-planetscale.md
@@ -133,10 +133,8 @@ doctrine:
         url: '%env(resolve:DATABASE_URL)%'
         options:
             # Connect to the database via SSL
-            !php/const:PDO::MYSQL_ATTR_SSL_CA: /opt/bref/ssl/cert.pem
-
-            # Note: When using symfony/yaml 6.2 and up it should be: (there is no colon after "const")
             !php/const PDO::MYSQL_ATTR_SSL_CA: /opt/bref/ssl/cert.pem
+
 # ...
 ```
 


mnapoli:
Thanks! Can we replace the link entirely? (do we need to keep the old version?)

ramondelafuente:
Yeah, it looks like the syntax without a colon has been supported since 3.4 and default since 4.0 (but the changelog is a little vague, see below). So indeed perhaps simply replacing with the new syntax is best.

```
6.2
 * Deprecate the `!php/const:` tag in key which will be replaced by the `!php/const` tag (without the colon) since 3.4

3.4.0
 * Deprecated the `!php/const:` tag which will be replaced by the
   `!php/const` tag (without the colon) in 4.0.

3.2.0
 * Added support for parsing PHP constants:
   Yaml::parse('!php/const:PHP_INT_MAX', Yaml::PARSE_CONSTANT);
```

docs/environment/database-planetscale.md
```diff
@@ -133,7 +133,7 @@ doctrine:
         url: '%env(resolve:DATABASE_URL)%'
         options:
             # Connect to the database via SSL
-            !php/const:PDO::MYSQL_ATTR_SSL_CA: /opt/bref/ssl/cert.pem
+            !php/const PDO::MYSQL_ATTR_SSL_CA: /opt/bref/ssl/cert.pem
```
mnapoli, line 136: Wait don't we need quotes to escape that? Like this:

```suggestion
            '!php/const PDO::MYSQL_ATTR_SSL_CA': /opt/bref/ssl/cert.pem
```

Have you tested this and it's working correctly for you?

ramondelafuente: Screenshot 2023-04-24 at 10 46 21

Works unquoted indeed.

mnapoli: OK let's do that then, thanks!

ramondelafuente: I think it's safe enough. I found the official docs also use the unquoted notation here:

https://symfony.com/doc/current/configuration.html#configuration-parameters

```
# config/services.yaml
parameters:
    # PHP constants as parameter values
    app.some_constant: !php/const GLOBAL_CONSTANT
    app.another_constant: !php/const App\Entity\BlogPost::MAX_ITEMS
```

PR #1521, opened 2023-04-24
Fix compatibility with Node 12

mnapoli:
See https://phpc.social/@thickey@mas.to/110254565376105520


GrahamCampbell:
Node 12 is already EOL, and node 14 will be EOL in just a few days, too.

mnapoli:
Yes but it is still supported by the `serverless` CLI. The patch is fairly simple so why not 🤷

PR #1548, opened 2023-05-24
New format serverless links and phrasing

crisman:
New format serverless links and phrasing
    
serverless.com wants a www. prefix on URLs now and also moved from
calling them case stodies to user stories with a new local part.
Also update some odd phrasing in the text.


mnapoli:
Thank you!

PR #1594, opened 2023-07-14
Fix #1593 `WebsocketEvent` `getRegion()` not returning correct region for custom domains

### LINKED ISSUE #1593: `WebsocketEvent::getRegion` is incompatible with custom domain names.
**Description:**

`Bref\Event\ApiGateway\WebsocketEvent` has a method `getRegion()` that should return the current AWS region e.g. `eu-west-1` 

This is implemented by parsing it out of the `domainName` from the event `requestContext`. 

This makes sense only in the case that default AWS provided API gateway domains are used. When you set up your own domain for the websocket API you end up getting a segment of that domain returned as your region. 

For example if you have a domain configured of `websocket.example.org` the AWS region returned by this method would be `org` which isn't correct.

Hope that makes sense, it's a fairly straightforward one but happy to provide more context if needed. I'm not sure if it's a good solution but the current region is available in Lambda via an environment variable. 

The below resolves it for me:

image

Shkeats:
Fixes https://github.com/brefphp/bref/issues/1593 using the `AWS_REGION` env variable set in the lambda runtime.

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -15,4 +15,7 @@
       <directory>./tests/Sam</directory>
     </testsuite>
   </testsuites>
+  <php>
+    <env name="AWS_REGION" value="eu-west-1" force="true" />
+  </php>
 </phpunit>
--- a/src/Event/ApiGateway/WebsocketEvent.php
+++ b/src/Event/ApiGateway/WebsocketEvent.php
@@ -105,7 +105,6 @@ class WebsocketEvent implements LambdaEvent
 
     public function getRegion(): string
     {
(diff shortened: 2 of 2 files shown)


mnapoli:
It seems the tests are failing, they might need some update.

Shkeats:
@mnapoli updated with the env set for the test suite. I'm not sure if it's fine to assume that it's always set and never returning false in the real world, I think we do that elsewhere in the project for lambda environment variables like `LAMBDA_TASK_ROOT` though!

PR #1615, opened 2023-08-14
Make sure Bref layers are always added

### LINKED ISSUE #1605: Error: Couldn't find valid bootstrap(s): [/var/task/bootstrap /opt/bootstrap] using bref/bref 2.1.0
**Description:**

i have the error `Error: Couldn't find valid bootstrap(s): [/var/task/bootstrap /opt/bootstrap]`  on my lambda after `sls deploy --stage=#env# -- function=#function_name#`

**How to reproduce:**

1. deploy you stack with multiple function `sls deploy --stage=#env#`
2. deploy only one fonction with `sls deploy --stage=#env# -- function=#function_name#`
3. run the function and see the error

t-richard:
Trying to fix #1605


index.js
```diff
@@ -197,22 +197,16 @@ class ServerlessPlugin {
...
         const isArmGlobally = config.provider.architecture === 'arm64';
-
-        // Check provider config
-        if (this.runtimes.includes(config.provider.runtime || '')) {
-            config.provider.layers = includeBrefLayers(
-                config.provider.runtime,
-                config.provider.layers || [], // make sure it's an array
-                isArmGlobally,
-            );
-            config.provider.runtime = 'provided.al2';
```
t-richard, line 208: By deleting this, we lose the global `provided.al2` which might be considered a BC break

PR #1627, opened 2023-08-30
Enable file path to data on bref-local with CDK

theoboldalex:
This change provides command line access to the `--path` flag where an application is built on AWS CDK rather than Serverless framework. This enables the passing of a path to a JSON file when running `vendor/bin/bref--local` instead of passing data as a string.

If you agree with the premise of this change, I will also update the relevant documentation and include in the PR.

### DIFF
--- a/src/bref-local
+++ b/src/bref-local
@@ -18,10 +18,14 @@ if (file_exists(__DIR__ . '/../vendor/autoload.php')) {
 
 $handler = $argv[1] ?? '';
 $data = $argv[2] ?? null;
-$opts = getopt("f::");
+$opts = getopt('', ['path:']);
 
-if (isset($opts['f']) && file_exists($opts['f'])) {
-    $data = file_get_contents($opts['f']);
+if (isset($opts['path'])) {
+    if (! file_exists($opts['path'])) {
+        throw new Exception('The file ' . $opts['path'] . ' could not be found.');
+    }
+
+    $data = file_get_contents($opts['path']);
 }
 
 try {
--- a/src/bref-local
+++ b/src/bref-local
@@ -25,6 +25,7 @@ if (isset($opts['path'])) {
         throw new Exception('The file ' . $opts['path'] . ' could not be found.');
     }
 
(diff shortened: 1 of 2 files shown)


mnapoli:
Hey, that sounds like a good idea. I'm not familiar with `getopt`, so I'm not sure what the command will look like. If you add docs that will probably be clearer to me 👍

theoboldalex:
Great, I will work on the docs and update the PR in due course. Thanks

mnapoli:
Thanks!

theoboldalex:
I have had another look at this and think it might actually be better to match the way we invoke locally with the way we would if using Serverless.

**Serverless**
`serverless bref:local -f hello --path=event.json`

**CDK/TF**
`vendor/bin/bref-local -f my-function.php --path event.json`

If you think that is worth doing, I will revert this PR to draft and work on that implementation.

Thanks

mnapoli:
Sorry I'm missing what would be different with what you implemented?

theoboldalex:
Apologies, I didn't explain very well above. 

In the current implementation, when we pass the `--path` flag, it must be passed _before_ the handler file. 

When running Bref with Serverless, the handler file has the `-f` flag. I could add this to the non-Serverless `bref-local` script to ensure consistency in how functions are invoked locally regardless of whether we are using Serverless, CDK or Terraform.

mnapoli:
Oh yes that would be great indeed, in most PHP CLIs (artisan, Symfony Console) that's how it works. If you can do that without pulling in more dependencies that would be a very good improvement 👍

src/bref-local
```diff
@@ -18,6 +18,16 @@ if (file_exists(__DIR__ . '/../vendor/autoload.php')) {
 
 $handler = $argv[1] ?? '';
 $data = $argv[2] ?? null;
+$opts = getopt('', ['path:']);
+
+if (isset($opts['path'])) {
+    if (! file_exists($opts['path'])) {
```
shadowhand, line 24: This should be changed to `! is_file(...)` because `file_exists` also returns true for directories.

PR #1635, opened 2023-09-04
Log error location

### LINKED ISSUE #1619: Lambda invoke failure does not return file or line number
**Description:**

When a lambda invoke error occurs, you get an error like follows in CloudWatch:

```
7a3b55ac-f1df-417f-9414-230fda2acb1f	Invoke Error	
{
    "errorType": "ErrorException",
    "errorMessage": "Warning: Trying to access array offset on value of type null",
    "stack": [
        "#0 /var/task/custom_code(24): CustomCode->handle(Array)",
        "#1 /var/task/vendor/bref/bref/src/Runtime/Invoker.php(29): Bref\\Runtime\\FileHandlerLocator::{closure}(Array, Object(Bref\\Context\\Context))",
        "#2 /var/task/vendor/bref/bref/src/Runtime/LambdaRuntime.php(89): Bref\\Runtime\\Invoker->invoke(Object(Closure), Array, Object(Bref\\Context\\Context))",
        "#3 /var/task/vendor/bref/bref/src/FunctionRuntime/Main.php(37): Bref\\Runtime\\LambdaRuntime->processNextEvent(Object(Closure))",
        "#4 /opt/bref/bootstrap.php(17): Bref\\FunctionRuntime\\Main::run()",
        "#5 {main}"
    ]
}
```

However, the stack is only showing the stack - not the current context of the error. Some errors raised by PHP contain this information where in the message it says (on line X in Y) but some errors do not (not entirely sure why to be honest!).

**How to reproduce:**

PHP 8.0
Access array offset on null and see if the line number appears

theoboldalex:
resolves #1619 

So that we can see the context of an error in Cloudwatch logs, I have added an `errorLocation` key to the `errorFormatted` data structure. This means that the file and line number where the exception is thrown is logged.

### DIFF
--- a/src/Runtime/LambdaRuntime.php
+++ b/src/Runtime/LambdaRuntime.php
@@ -202,6 +202,7 @@ final class LambdaRuntime
                 $previousErrors[] = [
                     'errorType' => get_class($previousError),
                     'errorMessage' => $previousError->getMessage(),
+                    'errorLocation' => 'File: ' . $previousError->getFile() . ', Line: ' . $previousError->getLine(),
                     'stack' => explode(PHP_EOL, $previousError->getTraceAsString()),
                 ];
             } while ($previousError->getPrevious() !== null);
@@ -230,6 +231,7 @@ final class LambdaRuntime
             $this->postJson($url, [
                 'errorType' => get_class($error),
                 'errorMessage' => $error->getMessage(),
+                'errorLocation' => 'File: ' . $error->getFile() . ', Line: ' . $error->getLine(),
                 'stackTrace' => $stackTraceAsArray,
             ]);
         }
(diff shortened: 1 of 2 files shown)


theoboldalex:
Any feedback on this appreciated @mnapoli

mnapoli:
I'm sorry, while reviewing this I'm realizing this object follows a structure defined by AWS Lambda: https://docs.aws.amazon.com/lambda/latest/dg/runtimes-api.html#runtimes-api-invokeerror

Screen-001433

I apologize for not spotting that sooner 🤦

I'm thinking of 3 options:

1. adding the extra `errorLocation` (current state of the PR) _if_ it doesn't break error reporting **and** that new field shows up in the CloudWatch logs
1. adding the file + line in the stack trace (insert it as the first line?
2. adding the file + line to the error message

For reference, here's what PHP logs in case of an uncaught exception:

Screen-001434

Note: I don't think it's a good reference though because it's an "Uncaught exception" line that is a fatal error. In our case it's not exactly that, exceptions are OK (to communicate errors to Lambda), so we don't necessarily want to copy that format.

I think the 1st option is the best, but that requires testing on Lambda to confirm the behavior works as expected.

mnapoli:
Closing inactive PRs, feel free to reopen.

PR #1641, opened 2023-09-09
POC: Adding specific sns and sqs exception when aws configration is invalid

robinlehrmann:
This would help developers with configurations like SNS->Lambda when they use the wrong consumer.

### DIFF
--- a/src/Event/Sns/SnsRecord.php
+++ b/src/Event/Sns/SnsRecord.php
@@ -4,6 +4,7 @@ namespace Bref\Event\Sns;
 
 use DateTimeImmutable;
 use InvalidArgumentException;
+use LogicException;
 
 /**
  * Represents a SNS message record.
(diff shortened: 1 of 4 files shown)


GrahamCampbell:
Is the stack trace not enough already?

robinlehrmann:
Not sure, I thought it could be helpful @GrahamCampbell but up to you what do you think @mnapoli ? :)

mnapoli:
I was writing this comment:

---

How about we keep the existing exception, but just improve the error message when possible? What I mean:

- when there's an invalid event and it's possible to identify the incoming event (e.g. SNS event for a SQS handler), we throw `InvalidArgumentException` with a message that mentions the event that was received
- where there's an invalid event and the incoming event is not identified, throw `InvalidArgumentException` without specific error message

For example:

```php
if (! is_array($record) || ! isset($record['eventSource']) || $record['eventSource'] !== 'aws:sqs') {
    throw new InvalidArgumentException($record['eventSource'] ?? '');
}
```

Then in `InvalidArgumentException` there could be a new optional argument:

```php
public function __construct(string $eventSource = null) {
$message = 
```

---

Then I opened the code and realized we already have something similar, we actually embed the whole event in the exception message (`throw new InvalidLambdaEvent('SQS', $this->event)`).

So I'm sorry, but I think the current message is good enough and it's not worth adding special exceptions and error messages just for SQS and SNS when they are mixed together. This is a very specific scenario, adding specific code for this is not worth it I think. The generic exception already has enough information.

So in the end, I don't think it's worth changing anything actually 😬

robinlehrmann:
ok, no worries @mnapoli :) Then feel free to close the issue as well

Issue #1644, opened 2023-09-11
Laravel does not work when creating a local development environment with docker-compose


Syzgyn:
**Description:**
When following the directions on the bref wiki on local development in docker, I get the following error:
`Unable to create lockable file: /tmp/storage/framework/cache/15/ed/15ed45fc33a4af8dc4baab88de709b23759e5b28. Please ensure you have permission to create files in this location.`

[Full Stacktrace available here](https://gist.github.com/Syzgyn/a6b1c1e08886441ba26a552f59a209a1)

This is happening in the php-81, 82, and 83 fpm-dev docker images.

Also found someone with the same issue here: https://github.com/brefphp/bref/issues/470#issuecomment-1513221424

**How to reproduce:**

1. Create a new laravel project with `composer create-project laravel/laravel test-project`
2. Follow the setup instructions [here](https://bref.sh/docs/frameworks/laravel)
3. Create the `docker-compose.yml` file from [here](https://bref.sh/docs/web-apps/local-development)

sunaoka:
Try specifying `nobody` as `user`.

```yaml
services:
    app:
        image: bref/php-81-fpm-dev:2
        ports: [ '8000:8000' ]
        volumes:
            - .:/var/task
        environment:
            HANDLER: public/index.php
        user: nobody  # <-----
```

Syzgyn:
That did it, thank you!

PR #1645, opened 2023-09-12
Update docker-compose.yml example to include user

Syzgyn:
From #1644, the `user` field needs to be in `docker-compose.yml` for it to work with laravel.


mnapoli:
Without extra details on why, and the fact that no one else had that problem, I'm going to close. Let's reopen if we have more details.

PR #1648, opened 2023-09-13
[1.x] Upgrade postgres to 16.0

GrahamCampbell:


GrahamCampbell:
Don't care to debug this. 15.4 is probably fine for a good while, until someone really needs 16.0.

PR #1652, opened 2023-09-25
feat: uncouple secrets loader

jaulz:
Right now, the logic whether the secrets loader will be used or not is implemented here in the first place. However, I think it should only be used to output a warning. That makes it easier to patch the secrets loader and for example use this proposed change: https://github.com/brefphp/secrets-loader/pull/4

### DIFF
--- a/src/LazySecretsLoader.php
+++ b/src/LazySecretsLoader.php
@@ -19,12 +19,16 @@ final class LazySecretsLoader
 {
     public static function loadSecretEnvironmentVariables(): void
     {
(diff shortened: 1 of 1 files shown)


mnapoli:
I don't understand what this is about honestly.

> However, I think it should only be used to output a warning

But the diff still throws an exception, so I'm not sure what you mean.

How are things different with that PR? What can you do that you couldn't do before?

mnapoli:
Closing to clean up the issue tracker, that's not something I plan on merging right now. Let's reopen the discussion if needed.

PR #1668, opened 2023-10-18
Expose the Lambda context in `LAMBDA_INVOCATION_CONTEXT` in the function runtime

mnapoli:
Will solve this issue: https://github.com/brefphp/laravel-bridge/issues/78

cc @jasonmccallister


peterabarry:
@mnapoli I am not that familiar with the source code in Bref yet but would that fix also extend to Octane? We moved from FPM to Octane and lost the aws request ids, which we use to build s3 traceable cache?

PR #1670, opened 2023-10-20
Bumped actions versions

GrahamCampbell:
Similar to changes already made on 1.x (other than one change, pending at #1669).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
 
     php-tests:
         name: Tests - PHP ${{ matrix.php }} ${{ matrix.dependency-version }}
-        runs-on: ubuntu-latest
+        runs-on: ubuntu-22.04
         timeout-minutes: 15
         strategy:
             matrix:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -12,7 +12,7 @@ jobs:
 
     php-tests:
         name: Tests - PHP ${{ matrix.php }} ${{ matrix.dependency-version }}
-        runs-on: ubuntu-latest
+        runs-on: ubuntu-22.04
```
mnapoli, line 15: What do we gain again by fixing the version? We'll have to update this eventually 😬

GrahamCampbell: We gain control over when this happens, just like we control the major versions on the other action things.

mnapoli: Right but actions can have BC breaks on major versions.

I don't see any way the Ubuntu version would break the CI?

GrahamCampbell: Ubuntu version typically does have breaking changes, and different actions may not be compatible yet.

GrahamCampbell: GitHub also doesn't upgrade everything at once, doing a rolling move per repo, so it's kinda janky to sometimes have one version, and sometimes another.

PR #1673, opened 2023-10-22
Detect bref/secrets-loader installed as dependency of another package

kevincerro:
If you require `bref/secrets-loader` in another package as dependency you get this warning on serverless cli
Captura de pantalla 2023-10-22 a las 13 54 20

With this simple change we detect if `bref/secrets-loader` is really installed, as direct dependency or as a dependency of another package.

### DIFF
--- a/plugin/secrets.js
+++ b/plugin/secrets.js
@@ -27,7 +27,7 @@ function warnIfUsingSecretsWithoutTheBrefDependency(serverless, log) {
             return;
         }
 
-        log.warning(`The following environment variables use the "bref-ssm:" prefix, but the "bref/secrets-loader" dependency is not installed via "composer.json".`);
+        log.warning(`The following environment variables use the "bref-ssm:" prefix, but the "bref/secrets-loader" dependency is not installed.`);
         allVariables.forEach(variable => log.warning(`    ${variable}`));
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Not sure we should allow transitive dependencies, as this can be risky. We should force applications to explicitly require it.

mnapoli:
I agree with @GrahamCampbell, it's usually a best practice to explicitly require your direct dependencies.

Do you have a specific use case here? (I'm curious how the package is installed via another dependency, is it an internal package?)

kevincerro:
> I agree with @GrahamCampbell, it's usually a best practice to explicitly require your direct dependencies.
> 
> Do you have a specific use case here? (I'm curious how the package is installed via another dependency, is it an internal package?)

Yes, I made an internal php package for a laravel-based php framework to autoconfigure it for serverless using bref.

mnapoli:
Makes sense.

The diff is very minimal and technically makes sense, I'm fine merging. Can you confirm you tested this and it works? (it's not covered by tests IIRC)

kevincerro:
> Makes sense.
> 
> The diff is very minimal and technically makes sense, I'm fine merging. Can you confirm you tested this and it works? (it's not covered by tests IIRC)
Yes, its working, warning is gone

## Before
Captura de pantalla 2023-10-22 a las 13 54 20

## After
Captura de pantalla 2023-10-23 a las 8 27 33

PR #1675, opened 2023-10-25
Add ap-southeast-3 region

### LINKED ISSUE #1609: bref.sh Cannot Be Used In The Asia Pacific (Jakarta) ap-southeast-3 Region
Hi, I'm from Indonesia. I'd love to be able to use bref.sh in my own country. So Far, I Can't Do It Because bref.sh Is Not Available In This Region. Can you add this region to the list of supported regions? I Am Really Looking Forward To It.

Thank You.

kevincerro:
Closes #1609

### DIFF
--- a/utils/layers.json/regions.json
+++ b/utils/layers.json/regions.json
@@ -17,6 +17,7 @@
     "ap-northeast-3",
     "ap-southeast-1",
     "ap-southeast-2",
+    "ap-southeast-3",
     "eu-south-1",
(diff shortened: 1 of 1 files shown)


mnapoli:
This requires a bit more unfortunately as this region is not enabled by default (right?). I'll hold off merging until I do

mnapoli:
This will be supported in Bref v3, will be released soon, check out the beta releases to test this out

PR #1680, opened 2023-11-01
Cite bref/api-gateway-websocket-client in the websockets docs page

andrea-cristaudo:
bref/api-gateway-websocket-client is not cited in either the websockets documentation or the Serverless Visually Explained websockets documentation (and there the namespace is also not updated)


mnapoli:
Thank you!

PR #1683, opened 2023-11-02
(Docs) Added information on the deployment docs for reducing deployment size when using aws/google SDKs

danniehansen:
My codebase was sitting just on the edge of maximum deployment size & every now and then staging deployments would complain after our patching routine because google or AWS added new services that increased the deployment size.

Our solution have for years been just to extend the never ending exclude list with new services. But this time i got quite annoying by it and with a few searches i found some nifty composer scripts that can be used to automatically cleanup unused services in both the Google and AWS SDKs!

AWS SDK was heavily inspired by https://github.com/aws/aws-sdk-php/pull/2456 - so the syntax is much the same.

This change adds a small callout segment in the deployment which just highlights this as these packages are often found in serverless deployments.

### DIFF
--- a/docs/deploy.mdx
+++ b/docs/deploy.mdx
@@ -36,6 +36,10 @@ Instead, let's remove development dependencies and optimize Composer's autoloade
 composer install --prefer-dist --optimize-autoloader --no-dev
 ```
 
+<Callout>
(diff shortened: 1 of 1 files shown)


mnapoli:
Thank you!

PR #1686, opened 2023-11-08
Expose the Lambda context in `LAMBDA_INVOCATION_CONTEXT` in all runtimes

mnapoli:
Follow-up of #1685 that reverted #1668. This is a different implementation.


src/Runtime/LambdaRuntime.php
```diff
@@ -427,4 +431,12 @@ private function ping(): void
         socket_sendto($sock, $message, strlen($message), 0, '3.219.198.164', 8125);
         socket_close($sock);
     }
+
+    private function setEnv(string $name, string $value): void
+    {
+        $_SERVER[$name] = $_ENV[$name] = $value;
+        if (! putenv("$name=$value")) {
```
GrahamCampbell, line 438: Do we have to use putenv?

mnapoli: That's in case a user is using `getenv`.

Do you see any issue? It's not like we're doing multithreading.

GrahamCampbell: Yeh, I still don't like it. Why do people even use getenv? I assume it's just because PHP named things in a way that's super confusing, and people reach for it as "the way to read env variables", even though `$_SERVER` is actually the closest approximation to "the list of env variables".

mnapoli: If the argument is "I don't like it" that's not enough for me :p 😈 

I agree this isn't the best, but might as well support it if anyone uses it.

PR #1690, opened 2023-11-11
Add ability to get queue name from sqs event

vmiguellima:
- Adds ability to get the queue name from a given sqs event

## Why is this needed?
- This way we can easily get the queue name from the sqs event, example usage:
- brefphp/laravel-bridge/pull/139


vmiguellima:
@mnapoli How the review process works?

PR #1694, opened 2023-11-20
SSM doc - Specifying that using SecureString is possible

bernier-db:
Since the example only talks about the `String` parameter type, I did not felt comfortable in storing the secrets as plain text in SSM.  Upon digging in github, I found this [comment on PR #1376](https://github.com/brefphp/bref/pull/1376#issuecomment-1405754051). 

I thought it would be helpful to specify that both SecureString and String are taken into account.


mnapoli:
Thanks!

Issue #1511, opened 2023-04-13
me-central-1


chaaban:
When trying to use the bref on me-central-1 region i get this error :

- Cannot resolve variable at "functions.hello.layers.0": There is no Bref layer named "php-80" in region "me-central-1".

Currently me-central-1 isn't supported .

anamaker:
+1

ArabAgile:
+1

afu-dev:
Hello, `me-central-1` is not a region enabled by default.

![image](https://user-images.githubusercontent.com/15209405/232591589-558f6553-bc65-4079-a658-453aefa72dcf.png)

I *think* this region can be activated without any special authorization (I'm trying it right now).

You can always host the layer yourself in this region.

anamaker:
@afu-dev any update for adding `me-central-1` region please?

mnapoli:
@anamaker feel free to send pull requests.

If you want me to work on this for you, you can also hire me, feel free to contact me: matthieu@bref.sh

PR #1706, opened 2023-12-11
added me-central-1 (UAE) region

A-Shevchenko:
To address #1511 
Though, it looks like you'd need to perform some manual action before that, as the layers deployment pipeline is not in Github?


mnapoli:
This will be supported in Bref v3, will be released soon, check out the beta releases to test this out

PR #1726, opened 2024-01-29
Cleanup uploaded files for PSR-15 handlers

mnapoli:

### DIFF
--- a/src/Event/Http/Psr7Bridge.php
+++ b/src/Event/Http/Psr7Bridge.php
@@ -177,7 +177,7 @@ final class Psr7Bridge
         $tmpFiles = glob(sys_get_temp_dir() . '/' . self::UPLOADED_FILES_PREFIX . '*');
         if ($tmpFiles !== false) {
             foreach ($tmpFiles as $file) {
-                if(is_file($file)) {
+                if (is_file($file)) {
                     unlink($file);
                 }
             }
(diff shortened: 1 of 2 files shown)


src/Event/Http/Psr7Bridge.php
```diff
@@ -166,4 +168,19 @@ private static function parseKeyAndInsertValueInArray(array &$array, string $key
...
+    /**
+     * Cleanup previously uploaded files.
+     */
+    public static function cleanupUploadedFiles(): void
+    {
+        $tmpFiles = glob(sys_get_temp_dir() . '/' . self::UPLOADED_FILES_PREFIX . '*');
+        if ($tmpFiles !== false) {
+            foreach ($tmpFiles as $file) {
+                if (is_file($file)) {
+                    unlink($file);
```
GrahamCampbell, line 181: This could fail if the previous request didn't cleanup the open handles properly.

GrahamCampbell (changes requested):

src/Event/Http/Psr7Bridge.php
mnapoli: In which case is there anything we can do? I don't want to turn this into an error and break stuff

GrahamCampbell: Maybe we need to silence it?

mnapoli: 🤦 I forgot about custom error handlers that might turn the warning into exceptions! (I was focused on the `bool` return). The joys of warnings.

GrahamCampbell (approved):

PR #1729, opened 2024-01-31
[1.x] curl 8.6.0

GrahamCampbell:
Prior to cURL 8.6.0, configure would silently skip when libpsl was missing, compiling without support for public suffix lists. Probably it's not a good idea to start making arguably breaking changes in Bref v1 at this point, and so I've added in the configure flag to continue to ignore the fact that libpsl is missing. For Bref v2 (https://github.com/brefphp/aws-lambda-layers/pull/157), I'll try and get that library included.

### DIFF
--- a/runtime/base/base.Dockerfile
+++ b/runtime/base/base.Dockerfile
@@ -226,7 +226,7 @@ RUN set -xe; \
 # #   - libssh2
 # # Needed by:
 # #   - php
-ENV VERSION_CURL=8.5.0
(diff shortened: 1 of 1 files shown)


mnapoli:
> Probably it's not a good idea to start making arguably breaking changes in Bref v1 at this point, and so I've added in the configure flag to continue to ignore the fact that libpsl is missing.

👍

Thanks!

PR #1733, opened 2024-01-31
Support malformed multipart body

mnapoli:
For example body containing broken array keys like `key0[key1][key2][`.

We use `parse_str` to ensure keys are parsed just like native PHP would.


tests/Event/Http/CommonHttpTest.php
```diff
@@ -383,7 +401,7 @@ public function test POST request with multipart file uploads(int $version
 --testBoundary--\r
 ";
         $this->assertBody($body);
-        $this->assertParsedBody([]);
+        $this->assertParsedBody(null);
```
mnapoli, line 404: Side-effect bugfix (minor)

PR #1747, opened 2024-02-24
Add missing runtime and add bucket example to s3 use-cases

thedomeffm:
I was confused that the s3 event example didn't need a runtime but that was simply missing here.

Additionally I want to manage the bucket via the lift so I added it as example 😄


mnapoli:
thanks!

PR #1758, opened 2024-03-15
Just add a retry before really throwing a FastCgiCommunicationFailed exception.

wysow:
Hello there!

This PR only to start a discussion to try to find a solution to this kind of random errors:

```
Error communicating with PHP-FPM to read the HTTP response. Bref will restart PHP-FPM now. Original exception message: hollodotme\FastCGI\Exceptions\ReadFailedException Stream got blocked, or terminated.
```

I'm pretty sure something smarter can be achieved but manually testing this for a few minutes now and only got 200 responses and no more 500...

### DIFF
--- a/src/FpmRuntime/FpmHandler.php
+++ b/src/FpmRuntime/FpmHandler.php
@@ -182,6 +182,10 @@ final class FpmHandler extends HttpHandler
                     $e->getMessage()
                 );
 
+                // Restart PHP-FPM: in some cases PHP-FPM is borked, that's the only way we can recover
+                $this->stop();
+                $this->start();
+
                 throw new FastCgiCommunicationFailed;
             }
         }
--- a/src/FpmRuntime/FpmHandler.php
+++ b/src/FpmRuntime/FpmHandler.php
@@ -171,7 +171,19 @@ final class FpmHandler extends HttpHandler
             $this->stop();
             $this->start();
 
-            throw new FastCgiCommunicationFailed;
+            try {
+                $socketId = $this->client->sendAsyncRequest($this->connection, $request);
+
+                $response = $this->client->readResponse($socketId, $timeoutDelayInMs);
+            } catch (Throwable $e) {
+                printf(
+                    "Error communicating with PHP-FPM to read the HTTP response. Bref will restart PHP-FPM now. Original exception message: %s %s\n",
+                    get_class($e),
+                    $e->getMessage()
+                );
+
+                throw new FastCgiCommunicationFailed;
+            }
         }
 
         $responseHeaders = $this->getResponseHeaders($response);


lguilbert (approved):

wysow:
Here is something interesting about this: https://github.com/hollodotme/fast-cgi-client/issues/68#issuecomment-1207839792

mnapoli:
What if the request is updating something in the database, or sending emails for example. That could run the same request/action twice, which might not be a good thing 🤔

How often do you get these errors?

GrahamCampbell (changes requested):
Re-trying seems like a bad idea to me. It's usual for clients to retry safe requests, not for proxy layers to do it, unless the error is for sure retriable such as an edge load balancer retrying a TLS handshake.

wysow:
@mnapoli Can't really tell how often it appears on the long run, but when I do some manual testing I sometimes get 50% error rate.... So that's a lot...

I know that retrying is a bit crappy but the fact is that in all the error I get in logs our code is never executed at all, everything happens on Bref side... So for us no problem to do a retry, and then we have a 0% error rate with manual testing.

GrahamCampbell:
50% error rate smells like something else is borked. Is it always failing after the first invoke?

wysow:
That's only my own feeling but yes I'm pretty sure it's always after the first invoke...

wysow:
And to add more context, we did NOT see this behavior on workers or console lambdas (working with Symfony)

wysow:
Just looked at the numbers of the last days with bref dashboard and I can see a 6-7% error rate on entire days.

mnapoli:
That is really weird, something else must be at play here. A 6-7% error rate would be affecting all Bref users if that was a global Bref problem.

I'd start looking at ways to pinpoint the problem:

- extra PHP extensions?
- out of memory?
- spawning sub-processes from PHP?
- timing out?
- try to see if it happens on a specific HTTP route?
- trying to reproduce with an empty project?

wysow:
Here a first list of answers, will keep you posted with others answers when I get them:

>     * extra PHP extensions?
 -> Yes, on this project we have redis and mongodb

>     * out of memory?
-> I'm pretty sure this is not the case as the error is really fast at the execution start (few milliseconds). We are using 2048Mo lambda memory size on this project.  

>     * spawning sub-processes from PHP?
-> this project is an API using bref 8.2 fpm layer and symfony so this is not the case for me here.

>     * timing out?
-> Like I said the error is really fast at the execution start so not the case either...

>     * try to see if it happens on a specific HTTP route?
-> will do more testing but I saw it on every HTTP route (GET mainly)

>     * trying to reproduce with an empty project?
-> Will try and keep you posted.

mnapoli:
Would be interesting to see too if this happens on cold starts. If not, is the request _before_ successful? Times out? Could fill the memory? (or any other reason it could leave the environment in a broken state)

Also nothing specific/exotic, like using Symfony Runtime, setting a non-standard handler, etc.

wysow:
> Would be interesting to see too if this happens on cold starts. If not, is the request _before_ successful? Times out? Could fill the memory? (or any other reason it could leave the environment in a broken state)
> 
> Also nothing specific/exotic, like using Symfony Runtime, setting a non-standard handler, etc.

As far as I manually tested this is not happening on cold starts, and the previous request is always successful. No timeout, not full memory... Nothing visible at least....

wysow:
Here is the raw log we got when this problem occur:
```
Error communicating with PHP-FPM to read the HTTP response. Bref will restart PHP-FPM now. Original exception message: hollodotme\FastCGI\Exceptions\ReadFailedException Stream got blocked, or terminated.

WARNING: [pool default] child 19 exited on signal 11 (SIGSEGV) after 424.432414 seconds from start
```

Here the 424 seconds is really weird as the behavior in an API client is really fast...

wysow:
> File `php.ini` custom in our projet with this content:
> 
> ```
> extension=intl
> ```

@mnapoli sorry this is not the right `php.ini` file... Here is the good one:
```
extension=mongodb
extension=redis
opcache.enable_cli=0
```

mnapoli:
Closing for now as we know that's unfortunately not a solution we can use since it may retry valid requests (non-idempotent) on occasion.

PR #1759, opened 2024-03-19
Add note to about reduced response limit with binary responses

jiw-mh:
In the documentation it is mentioned that the size limitation of responses is 6MB though through the use of base64 in practice it is reduced to about 4.5MB. This PR adds a note of it.


mnapoli:
Perfect addition!

PR #1761, opened 2024-03-21
Add SQS polling information to docs

harveydobson:
Updating documentation based on our discussion on https://github.com/brefphp/bref/discussions/1760 in an attempt to be helpful! 

I wonder if this information would be better in a more centralised area of the docs that can provide more detailed information about different parts of the infrastructure layer work. I couldn't see anything like this so for now I've added the information to the Symfony Messenger and Laravel Queue sections.


mnapoli:
This might be best on this page though: https://bref.sh/docs/use-cases/sqs

mnapoli:
Replaced by https://github.com/brefphp/bref/pull/1826

PR #1797, opened 2024-04-30
Colons in URIs are not supported with the function runtime

mnapoli:
For example `/tags/john:81`.

I'm opening this PR with a test case to reproduce. I don't have a fix yet.

@Nyholm the root of the exception is in https://github.com/Nyholm/psr7, specifically here:

https://github.com/Nyholm/psr7/blob/229484fc939a76a6579e331b55d86a77fb8e2863/src/Uri.php#L54-L56

This is caused by PHP not supporting these URIs in `parse_url`: https://3v4l.org/jpiQo

https://github.com/php/php-src/issues/12703

Any idea whether https://github.com/Nyholm/psr7 could have a workaround to support these URIs? Should I open a bug there? (do you want to address this?)

### DIFF
--- a/tests/Event/Http/CommonHttpTest.php
+++ b/tests/Event/Http/CommonHttpTest.php
@@ -520,6 +520,15 @@ Year,Make,Model
         $this->assertBasicAuthPassword('secret');
     }
 
+    /**
+     * @dataProvider provide API Gateway versions
+     */
+    public function test colon in uri(int $version)
+    {
+        $this->fromFixture(__DIR__ . "/Fixture/ag-v$version-colon-in-uri.json");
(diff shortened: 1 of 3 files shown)


GrahamCampbell:
Strictly speaking, this is an invalid URL. The colon should be `%3A`.

Nyholm:
Yeah. I was about to say the same as Graham. If it is a valid url, then we should support it. But if it is invalid, then an exception should be thrown.

mnapoli:
hold on good points, I actually got the report that it was happening both with and without the encoding, I should have made it clearer (and I went with the easiest repro, which may have been a mistake from me). I'll investigate and report back.

PR #1812, opened 2024-05-21
Fix against serverless v3

mnapoli:
Serverless Framework has started releasing a beta version as a stable version on NPM.

This breaks all the existing systems and tutorials out there because it requires a license key now. To fix the CI, and to make life easier for newcomers, I'm updating the instructions to install v3.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
                     node-version: 18
             # serverless is required by some tests
             -   name: Install serverless
-                run: npm i -g serverless
+                run: npm i -g serverless@3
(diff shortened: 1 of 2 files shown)


cybage-kmistry:
Yes using serverless with v3 explicitly will resolve this issue. But going forward we are not going to upgrade to v4 of serverless with bref?

nadar:
would be interesting to know @mnapoli where things are going with bref, having serverless v4 in mind (which is an absolute pain 😢 )

cybage-kmistry:
@mnapoli @nadar May be something with AWS SAM which is as simple as doing deployment with bref.

PR #1824, opened 2024-06-12
close #1369 increase the php-fpm log limit to match AWS Cloudwatch

### LINKED ISSUE #1369: On php-fpm layer in production,   after a certain limit,  logging on stderr  hang the process
**Description:**

I'm not sure if it's a bref bug or a AWS one but writting too much log on stderr hang after an arbitrary limit  

something like  "you write 152 logs, it's fine"  " you try to write the 153th, it timeout" 

**How to reproduce:**

I created this route to reproduce the issue 

```
    #[Route(
        path: '/test-reproduce-bref-issue',
        name: 'ping',
        methods: ['GET', 'HEAD']
    )]
    public function brefIssueReproducer(LoggerInterface $logger, Request $request): JsonResponse
    {
        for ($i = 0; $i < $request->query->getInt('nbrLog'); $i++) {
            $logger->info("you_can_ignore_i_m_just_here_to_try_to_reproduce_a_bug");
        }
        return $this->json([]);
    }

```

That you can test live here 

https://app-testenv.rosaly.info/test-reproduce-bref-issue?nbrLog=153  -> does not hang , quite fast 99ms 
https://app-testenv.rosaly.info/test-reproduce-bref-issue?nbrLog=154 ->  hang , timeout (we have 15 seconds timeout)

allan-simon:
So that the php-fpm process will not hang when reaching log_limit (it can still reach but it's muchhh harder, and reaching it would anyway cause an error with  AWS Cloudwatch )

### DIFF
--- a/runtime/common/fpm/php-fpm.conf
+++ b/runtime/common/fpm/php-fpm.conf
@@ -7,7 +7,10 @@ pid = /tmp/.bref/php-fpm.pid
 ;log_level = 'warning'
 
 ; New PHP 7.3 option that includes the maximum length when writing to stderr
-log_limit = 8192
+: See https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_PutLogEvents.html
+; and https://github.com/php/php-src/issues/9010#issuecomment-1184574379
+; We set it to the limit of a AWS Cloudwatch log otherwise, reaching this limit would hang the php-fpm process.
+log_limit = 1048575
 
 [default]
 pm = static


mnapoli:
Hi, this branch is a very old (obsolete) branch:

Screen-001826

You might want to open this against https://github.com/brefphp/aws-lambda-layers

> So that the php-fpm process will not hang when reaching log_limit

I'm not sure I understand why the fpm process would hang when reaching `log_limit`? I understand that the setting will split log lines (bad), but why is anything hanging?

Also have you been able to reproduce the problem and check that this change fixes it?

Thanks!

allan-simon:
@mnapoli I will reopen in the other repository ,  I see that in that one there's clear instruction on how to build the layer so i will be able to test if that actually fix the issue, but looking to the php-src issue I'm pretty confident it does, though I agree with you I'm not sure why it hangs (and I can't reproduce it with the -dev layer )

PR #1852, opened 2024-08-14
Fix #1369 and #1714 Support large amount of logs

### LINKED ISSUE #1369: On php-fpm layer in production,   after a certain limit,  logging on stderr  hang the process

### LINKED ISSUE #1714: Logging to stderr has some kind of limit in the PHP-FPM runtime
**Description:**
It seems that logging to stderr with monolog (using Symfony 6.3) has some kind of limit in the web function, with php-fpm runtime.
Whenever some threshold in amount (and size) of log entries is reached the lambda seems to hang until it times out after 28 seconds. 

For instance we have an import endpoint which reads an uploaded Excel sheet. With logging enabled we can process about 260 rows before it hangs. The same command for processing but then run via a console command has no issue whatsoever, it handles 13.000+ rows easily. Also worker lambdas do not show this behavior

This behavior is also observed locally with the same runtime.

**Debgging steps tried:**
When minimum log level is raised to notice, the logging is reduced to next to nothing (only when an unhandled exception or a PHP notice/warning/error occurs it will be logged). In this case importing via web request works even for 13.000 rows, with only 1 or 2 seconds to spare before the 28 second timeout.
I've also tried increasing memorySize to no effect. No difference noticed comparing 1792M, 10240M and the default 1024M (which we normally use).

**How to reproduce:**
Attached a Symfony controller and a console command to test amount of log entries (on top of entries generated by frameork and dependencies).
[source.zip](https://github.com/brefphp/bref/files/13904299/source.zip)
- Create a simple Symfony (6.3) project with monolog and add the controller and command.
- Add and setup Bref (2.1.3)
- Setup serverless.yml with the runtimes (web: php-82-fpm, console: php-82-console, worker: php-82)
- Use a local docker setup with web php image bref/php-82-fpm-dev:2

See also Slack: https://brefworkspace.slack.com/archives/C057S6G4Z9N/p1704917185595609

mnapoli:
Fix #1369, Fix #1714

We get rid of symfony/process where we had to forward all the FPM output to PHP. Instead we use `proc_open` and directly stream output to stdout/stderr.

### DIFF
--- a/src/FpmRuntime/FpmHandler.php
+++ b/src/FpmRuntime/FpmHandler.php
@@ -15,7 +15,7 @@ use hollodotme\FastCGI\Exceptions\TimedoutException;
 use hollodotme\FastCGI\Interfaces\ProvidesRequestData;
 use hollodotme\FastCGI\Interfaces\ProvidesResponseData;
 use hollodotme\FastCGI\SocketConnections\UnixDomainSocket;
-use Symfony\Component\Process\Process;
+use RuntimeException;
 use Throwable;
 
 /**
@@ -47,7 +47,8 @@ final class FpmHandler extends HttpHandler
     private UnixDomainSocket $connection;
     private string $handler;
     private string $configFile;
-    private ?Process $fpm = null;
+    /** @var resource|null */
+    private $fpm;
 
     public function __construct(string $handler, string $configFile = self::CONFIG)
     {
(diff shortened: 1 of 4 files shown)


allan-simon:
I will try this today , really great news !

allan-simon:
@mnapoli it works !

curlysanders:
Thank you for your excellent work @mnapoli
Can confirm this is now working like a charm.

mnapoli:
Awesome!

PR #1862, opened 2024-09-20
fix: key was missing in kafka record

falnyr:
Added a getter for missing key property because it was missing.

As per the official SDK for Java: https://github.com/aws/aws-lambda-java-libs/blob/main/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/KafkaEvent.java

### DIFF
--- a/src/Event/Kafka/KafkaRecord.php
+++ b/src/Event/Kafka/KafkaRecord.php
@@ -18,7 +18,7 @@ class KafkaRecord
         }
         $this->record = $record;
     }
-    
+
     public function getKey(): string
     {
         return $this->record['key'];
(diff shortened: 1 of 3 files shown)


mnapoli:
Thanks, can you update the tests too?

falnyr:
Thanks for letting me know @mnapoli. Added now.

mnapoli:
Can you fix the build?

src/Event/Kafka/KafkaRecord.php
```diff
@@ -18,6 +18,11 @@ public function __construct(mixed $record)
         }
         $this->record = $record;
     }
+
```
GrahamCampbell, line 21: The issue is that this line contains 4 spaces. It should contain 0.

GrahamCampbell (changes requested):

PR #1923, opened 2025-03-07
Track cold starts

mnapoli:
This PR adds cold start tracking to [the telemetry ping](https://bref.sh/docs/runtimes/runtimes-details#telemetry-ping).

The goal is to finally have a reliable answer regarding "how many invocations are cold starts".

Sampling this value at the scale of Bref (40B invocations/mo) will give us a good idea on how many cold starts to expect for a real project. This is a question that is asked very often from non-Bref users, and having a concrete number will help.

Also it will help us track changes in this metric as AWS continues to lower the number of cold starts by prewarming stuff automatically ([one example of that](https://aaronstuyvenberg.com/posts/understanding-proactive-initialization)).

The new `ColdStartTracker` is `@internal` for now. It was originally created for [Sentry](https://bref.sh/sentry) and X-Ray integrations, but it might be useful to have in Bref directly. Once I stabilize its API I'll make it public (same process as for Bref events).


src/Runtime/ColdStartTracker.php
```diff
@@ -0,0 +1,121 @@
...
+        self::$hasFirstInvocationStarted = true;
+
+        if (self::$currentInvocationIsColdStart) {
+            // There was a cold start, let's figure out if it was a proactive initialization
+            $timeElapsedSinceColdStartEnded = microtime(true) - self::$coldStartEndedTime;
+            // If more than 100ms have passed since the cold start ended, we can assume the
+            // Lambda sandbox was paused/frozen between the cold start and the first invocation
+            // (100ms is an arbitrary value, we could use a lower value but I want to be conservative)
+            // That means the Lambda sandbox was initialized proactively
+            self::$wasProactiveInitialization = $timeElapsedSinceColdStartEnded > 0.1;
```
mnapoli, line 64: For reference I'm seeing durations of ~4-7ms

PR #1980, opened 2025-07-21
Add response stream support for lambda function urls

tryvin:
With Lambda Function URLs supporting `RESPONSE_STREAM`, we can make the `HttpHandler` implementations return an `StreamedHttpResponse` instance which supports iterating over a `Generator` and streams the contents as they are received.

This also adds support to PHP-FPM streaming support, which only works with 8.1+ PHP versions, as it relies on Fibers.

Dependency of: https://github.com/brefphp/laravel-bridge/pull/185

## Fallback

Lambda invoke mode `RESPONSE_STREAM` requires you to chunk content even if you are not sending a stream, so `HttpResponse` will either convert a plain string body to a Generator or convert a Generator to a plain string body depending on the `BREF_STREAMED_MODE` env variable.


tryvin:
Lemme fix these errors from the tests.

tryvin:
@mnapoli Tests fixed, and added more tests to the streaming code

georgeboot:
Well done on this work. Like the use of fibers on modern php versions

tryvin:
@georgeboot thanks, Fiber was the only way I could find to allow the read function from curl not having to wait for the next chunk, in my tests adding a sleep between each yield without fibers make the first byte come after the first sleep completed.

BTW, FastCGI (FPM) as it is today won't have stream support, but, the FpmClient does allow you to have callbacks called every time there's new data to be read from the script, so, there's a possibility there to hook into them and yield content from there

mnapoli:
Thank for working on this!

I'm torn because I definitely want response streaming, BUT every time someone mentions they want to use it, it actually wouldn't work for them.

Here is when response streaming can be used:

- ✅ you are **not** using API Gateway or ALB, but instead using CloudFront + Lambda function URL
- ✅ you are **not** using the FPM layer, but instead using Octane or similar
- ✅ you are **not** running functions in a VPC (which rules out a lot of applications that use a RDS database via a VPC)

I'm curious if anyone reading this PR fits all these requirements?

GrahamCampbell:
Voxie fits those requirements. I do agree that this not working when using fpm probably excludes most users, but the use case for this is probably kinda niche and that might not be an issue. People might wanna spin up a fresh project to use this feature, such as for making an MCP server run on Lambda using PHP. 🐘

tryvin:
@GrahamCampbell @mnapoli my two cents on FPM:

It's not that it won't ever work, it's just that it requires a little more coding that for my specific use case I didn't take the time to implement.

But basically, we have half the job done, the FPM client does allow us to receive chunks of the output via callbacks, it's only that the way the FPM handler was implemented it waits for the FPM response to end so it can finally parse the output in Api gateway formats.

If not having FPM is a blocker, I can take some time to figure it out.

mnapoli:
I've looked into this today, here's my take:

- this is awesome work, thank you for contributing!
- ideally we should support FPM invocations if possible (not a requirement for this PR to be merged)
- we should support streamed invocations for all invocations, i.e. not just HTTP handlers. This will be super helpful for scenarios like running CLI commands
- users should not have to configure the Bref runtime via env variables:
  - the same function can be invoked with buffered or streamed responses (in non-HTTP scenarios), Bref should decide whether to returned a streamed response or not on the fly
    - I haven't found a field in the `/next` runtime API that would let us know if we are in a buffered or streamed invocation, I don't know how the Node runtime supports this. I have asked the AWS Lambda team but I don't know if I will get a response.
- we should move this to Bref v3, it's a big feature that could make v3 even more interesting, and a significant change in the codebase, since v3 is so close let's simplify maintenance and introduce these big changes in the next major so that we don't have to maintain this in v2. Also that will allow us to avoid supporting PHP 8.0 and just use the fibers

Oh and also have you tested this in AWS Lambda directly?

tryvin:
@mnapoli I will try to answer some of the questions the best I can as I'm on mobile now.

Lambda as far as I could figure out doesn't let us know if it's using a streamed response or not, your code is the one responsible to figure it out, on Javascript you decorate the handler with a specific function, and that does the trick, but so far, from their docs there's nothing that says "check xyz variable and it will tell you", hence the env variable I added.

I even tried to answer using streaming to a non streaming function and it didn't work, they kinda spawn an specific binary when it's set to stream response and you will communicate with this service instead of lambda directly.

And yeah, I did test it in "prod", I have a project that is running with this code :)

mnapoli:
Thanks for the quick response!

I need to figure out how the Node runtime does it. A function can be invoked via the normal `Invoke` API, or via [InvokeWithResponseStream](https://docs.aws.amazon.com/lambda/latest/dg/API_InvokeWithResponseStream.html), so I'm not sure how the Node runtime knows which invocation API it was called with 🤔

tryvin:
@mnapoli yeah, I didn't dump the environment to be honest, nor I looked at the parent calling the binary, maybe the answer is there.

mnapoli:
Sharing some notes in random order from today:

- There is nothing useful in the invoke payload or headers (aka context of the invocation) that could tell us whether it's a buffered or streamed invocation 😞
- If I return a streaming response in a simple function and invoke it with the "sync" API (`Invoke`), it seems to work correctly
  - I'm saying "seems" because it's hard being 100% certain that my response was actually 100% streamed, even though I did make sure of it as much as I could
  - That made me assume "hey we could always return a streamed response and Lambda will do the mapping when needed"
- I tried making the LambdaRuntime class use the streaming API (no env variable, all is streaming), it worked in some scenarios but it broke for API Gateway v2 (it seemed to work for API Gateway v1)
  - It's possible the error actually came from my code or something I did, but I couldn't figure it out this morning.
- I've sent an email to a contact in the Lambda team to know if they have more info
- I think I stumbled upon a bug in the PHP SDK: https://github.com/aws/aws-sdk-php/issues/2787#issuecomment-3213706514

As you can see above, this is all very hard to work with. Ideally I don't want to maintain to many different approaches and flags and code paths, that's why I'm trying to simplify the problem (and making everything a streamed response in the LambdaRuntime class).

tryvin:
@mnapoli 

> I think I stumbled upon a bug in the PHP SDK: https://github.com/aws/aws-sdk-php/issues/2787#issuecomment-3213706514

I think they forgot to add the `StreamRequestPayloadMiddleware` to the LambdaClient; that's why it's waiting synchronously.

> If I return a streaming response in a simple function and invoke it with the "sync" API (Invoke), it seems to work correctly

Yeah, I don't remember correctly now, but in my tests, it wouldn't send any content to the client if the Lambda didn't have the response stream enabled, and I tried to send a streamed response (even if the response was the whole content at once)

That's why I added the variable, in that sense I agree with you, I don't like extra variables to enable XYZ feature at all, but since I couldn't figure out (in my little time) a way to make it detect it properly, and we already had the BASE64 flag, I figured out it would be "ok".

> There is nothing useful in the invoke payload or headers (aka context of the invocation) that could tell us whether it's a buffered or streamed invocation 😞

Have you checked the function environment? Maybe there's something there, or maybe the process parent arguments, like something `--with-streaming-enabled`, I don't know.

The thing with this is that AWS poorly documented it; I found only one article in their blog that had the "recipe" on how to send streamed responses by yourself.

mnapoli:
> Have you checked the function environment? Maybe there's something there, or maybe the process parent arguments, like something --with-streaming-enabled, I don't know.

The same function can be invoke both buffered and streaming. I.e. it's not the function itself that is streaming or not, it's the invocation. That's why I didn't bother looking into that.

tryvin:
@mnapoli I did some tests in a staging environment here,

So, removing the `invokeMode: RESPONSE_STREAM` and keeping the variable enabled `BREF_STREAMED_MODE=1`, it did work, and I could see the content of the page; headers were sent, cookies were sent, and everything worked fine.

That being said, I also dumped the env variables to see if I could find anything out of the ordinary there, and there are no differences in the env variables; they are all the same, so we can discard that part.

Maybe the way to go will always be to send it as stream, but... that has a downside that I didn't test, Gzip compression on Cloudfront, when the URL is a response stream, cloudfront doesn't apply compression, at all, don't know if when disabling the response stream but continue sending it as a stream from bref it would apply compression, that's another thing to test.

In my specific case, I implemented compression streaming as a Laravel middleware, no big deal, but that's a "downside"

kevincerro:
Good job ❤️
This is definitely something we would use, and we had to replace it with other alternatives some years ago.
In our case, we needed to download files directly from Lambda, but ended up pushing them to S3 and redirecting with a presigned URL.

With response streaming, you can now return up to 200 MB of payload:
https://docs.aws.amazon.com/lambda/latest/dg/configuration-response-streaming.html

tryvin:
@mnapoli Good news! FPM now supports Streamed responses 🎉  Only if you have Fibers support, though, which is basically PHP 8.1+

On the other hand, I discovered an issue at the Octane handler that will most likely require Fibers to solve. If your Generator tries to use anything from the framework that wasn't preloaded in memory, it will fail because the framework was terminated by the time your generator ran.

deleugpn:
> In our case, we needed to download files directly from Lambda, but ended up pushing them to S3 and redirecting with a presigned URL.
> 
> With response streaming, you can now return up to 200 MB of payload: https://docs.aws.amazon.com/lambda/latest/dg/configuration-response-streaming.html

just a factor to consider: if your Lambda is in a VPC, then streaming will cost you NAT Transfer while S3 (your current architecture) will not. Depending on how much download bandwidth you’re using, switching your architecture could ramp up a significant bill.

tryvin:
@mnapoli I guess it's time to fix the conflicts and make it work with bref v3. Congrats on the release!

mnapoli:
Thanks! Sorry for the git conflicts here 😅 
I thought really hard about this PR, but I was already late with v3 and I'm really not 100% sure what the best approach is for this still. One problem at a time I guess…

tryvin:
> Thanks! Sorry for the git conflicts here 😅 I thought really hard about this PR, but I was already late with v3 and I'm really not 100% sure what the best approach is for this still. One problem at a time I guess…

No worries with the conflicts, I will merge v3 in my branch and fix them, and of course, test it.

As for the best approach, after using it for so long, and seeing other adapters (including from AWS) there's no way for detecting if it's streaming mode or not, you do need to set some environment variable, even with the AWS http lambda adapter.

tryvin:
@mnapoli Closing it in favour of https://github.com/brefphp/bref/pull/2168

PR #2016, opened 2025-10-27
Add note on reducing vendor size for AWS SDK

aromka:
Added note about removing unused AWS services to reduce bundle size.

### DIFF
--- a/docs/runtimes/function.mdx
+++ b/docs/runtimes/function.mdx
@@ -248,6 +248,10 @@ You first need to install the AWS PHP SDK by running
 $ composer require aws/aws-sdk-php
 ```
 
+<Callout type="info">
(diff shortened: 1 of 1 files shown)


mnapoli:
Thanks! The note already exists [here](https://github.com/brefphp/bref/blob/45261cfd8e99ee0e6d7fb1a36b37a403912cecc2/docs/deploy.mdx#deploying-for-production), maybe it can be made more visible by:

- turning the small note into a sub-section (that way we have a title that we can link to)
- updating your diff so that it links to the existing section (to avoid duplicating the info, but making sure all readers can find it more easily)

What do you think?

aromka:
Oh cool, yes, I think it should be more visible. The only note I saw was on using AsyncAws and completely missed that one.

mnapoli:
I've opened #2069 to continue this pull request, thank you for getting this started!

PR #2021, opened 2025-11-04
Modify PHP extensions loading paths in docs

jezmck:
Updated PHP extensions paths in configuration example, so that it actually finds the extensions.

### DIFF
--- a/docs/environment/php.mdx
+++ b/docs/environment/php.mdx
@@ -103,9 +103,9 @@ The following extensions are installed in Bref runtimes, but disabled by default
 You can enable these extensions by loading them in `php/conf.d/php.ini` (as mentioned in [the section above](#phpini)), for example:
 
 ```ini filename="php/conf.d/php.ini"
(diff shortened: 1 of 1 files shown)


mnapoli:
That should not be necessary, the include path for extensions should already include that path.

Was it not working for you? Bref v2 or v3 beta?

jezmck:
It was not working for me.

I have a weird setup that I've inherited, but it's bref v2.3.13.

mnapoli:
That can happen if `PHP_INI_SCAN_DIR` was overridden for example.

jezmck:
I had to set that for it to pick up my ini file.

deleugpn:
You can add custom ini files as instructed in https://bref.sh/docs/environment/php#customizing-phpini

jezmck:
I know, but that wasn't working, likely because of the strange set up I've inherited which doesn't seem to have the same structure.

Issue #2077, opened 2026-03-02
Silent FPM master death causing persistent broken pipe errors (~1-2% of requests)


sid-stripe:
**Description**

We're seeing persistent `WriteFailedException: Failed to write request to socket [broken pipe]` errors at a ~1-2% error rate across all Lambda instances. The FPM **master process** is silently dying between requests — no error logged, no child exit warnings, no crash messages. Bref recovers by restarting FPM, but the current invocation returns a 500.

The error pattern on a single Lambda instance:

1. Request N completes normally (e.g., 220ms)
2. Request N+1 arrives ~6ms later
3. `WriteFailedException` immediately — the FPM socket is already dead
4. Bref restarts FPM, next request works

**It's the master dying, not the child.** We searched all CloudWatch logs for `child exited on signal` — zero matches. We verified this logging works by pulling the Docker image, starting FPM, and sending SIGSEGV to the child — the master correctly logged `WARNING: [pool default] child 10 exited on signal 11 (SIGSEGV)` to stderr via `--force-stderr`. The absence of these messages in production confirms the master itself is dying.

Errors also spontaneously drop to near-zero (~0.01%) for 7-14 hours without any deployment, then climb back to ~1-2%. This suggests state corruption that self-heals through Lambda instance recycling.

**How to reproduce**

We haven't found a reliable reproduction — it's intermittent and only manifests in production under real traffic. It has persisted since our initial Bref migration and across PHP 8.1 and 8.4.

- **Bref**: 2.4.18 (Docker images, not layers)
- **Docker image**: `bref/php-84-fpm:2`
- **PHP**: 8.4.18
- **AWS service**: Lambda (via API Gateway HTTP API)
- **Lambda config**: 1024 MB memory, 28s timeout
- **FPM config**: Bref defaults (`pm=static`, `max_children=1`, `log_limit=8192`)
- **Extensions**: redis (Predis pure PHP client), intl, opcache, pcntl, posix, pdo_mysql
- **Framework**: Laravel 10

**What we've ruled out**

- SIGPIPE (#1854) — zero matches in CloudWatch
- JIT segfaults (#842) — `opcache.jit = disable`, verified from image
- OOM — memory well within Lambda limits
- Excimer profiler — removed from Dockerfile entirely, errors persist
- Sentry tracing/profiling — fully disabled, errors persist
- Redis/Valkey — Predis runs in child only, master never touches it
- Large responses — no correlation between response size and errors
[...]

PR #2081, opened 2026-03-04
Retry request on WriteFailedException instead of returning 500

sid-stripe:
Related to https://github.com/brefphp/bref/issues/2077

  ## Summary

When PHP-FPM's Unix socket is broken between Lambda invocations (e.g. the FPM worker died while the master is still alive), `sendAsyncRequest` throws a `WriteFailedException` (broken pipe). The current behavior restarts FPM but still throws `FastCgiCommunicationFailed`, returning a 500 to the caller even though FPM was just successfully restarted and is ready to serve.

This PR retries the request once against the freshly restarted FPM process on `WriteFailedException`, turning a guaranteed 500 into a transparent recovery.

  ## Why this is safe

  A `WriteFailedException` means the FastCGI request **failed to write to the socket** — the request never reached PHP-FPM. Since no application code executed, retrying is idempotent.

 Other exceptions (e.g. `ReadFailedException`) are **not** retried because the request may have already been processed by PHP, and retrying could cause double-execution. The existing behavior (log, restart FPM, throw `FastCgiCommunicationFailed`) is preserved for those cases.

  ## Background

   Key findings from production diagnostics for https://github.com/brefphp/bref/issues/2077:

  - **The FPM master is alive** when the broken pipe occurs — `proc_get_status()` shows `running=true`, `signaled=false`, `exitcode=-1`. The master is not crashing.
  - **The FPM worker or socket is dead.** The write fails because the reader end (worker) has closed its end of the Unix socket, likely due to the worker dying between invocations.
  - **The error is transient.** After Bref restarts FPM, the very next request to the same Lambda instance succeeds. The current code already does the right thing (restart FPM) but then discards the recovery by throwing `FastCgiCommunicationFailed`.
  - **The pattern:** Request N completes normally → ~6ms later request N+1 arrives → `WriteFailedException` immediately → Bref restarts FPM → current code throws 500 → next request works fine.

 I still don't know the root cause of the worker death is still under investigation (Lambda freeze/thaw, socket FD corruption, etc.), but regardless of cause, retrying on write failure is a safe mitigation that eliminates unnecessary 500s.

  ## What changed

  In the `catch (Throwable $e)` block of `FpmHandler::handleRequest()`:

  1. FPM is still restarted (`stop()` + `start()`) on all errors — no change
  2. **New:** If the exception is `WriteFailedException`, retry the request once against the fresh FPM
  3. If the retry also fails, log both exceptions and throw `FastCgiCommunicationFailed` as before
  4. For all other exceptions, behavior is unchanged (log + throw)

 We've been running this patch via `cweagans/composer-patches` in production. Broken pipe errors that previously resulted in 500s are now transparently recovered, with no double-execution or other side effects observed.
 
 Happy to hear your thoughts! Thought we will submit a PR since all the retries have been successful.

### DIFF
--- a/src/FpmRuntime/FpmHandler.php
+++ b/src/FpmRuntime/FpmHandler.php
@@ -12,6 +12,7 @@ use Bref\FpmRuntime\FastCgi\Timeout;
 use Exception;
 use hollodotme\FastCGI\Client;
 use hollodotme\FastCGI\Exceptions\TimedoutException;
+use hollodotme\FastCGI\Exceptions\WriteFailedException;
 use hollodotme\FastCGI\Interfaces\ProvidesRequestData;
 use hollodotme\FastCGI\Interfaces\ProvidesResponseData;
 use hollodotme\FastCGI\SocketConnections\UnixDomainSocket;
@@ -172,17 +173,39 @@ final class FpmHandler extends HttpHandler
             // - the 500 response is the same as if an exception happened in Bref
             throw new Timeout($timeoutDelayInMs, $context->getAwsRequestId());
         } catch (Throwable $e) {
-            printf(
-                "Error communicating with PHP-FPM to read the HTTP response. Bref will restart PHP-FPM now. Original exception message: %s %s\n",
-                get_class($e),
-                $e->getMessage()
-            );
-
             // Restart PHP-FPM: in some cases PHP-FPM is borked, that's the only way we can recover
             $this->stop();
             $this->start();
 
-            throw new FastCgiCommunicationFailed;
+            // On WriteFailedException (broken pipe), the request never reached PHP-FPM so it is
+            // safe to retry once against the freshly restarted FPM process. This avoids returning
+            // a 500 error to the caller for a transient FPM socket failure.
+            // Other exceptions (e.g. ReadFailedException) mean the request may have already been
+            // processed by PHP, so retrying could cause double-execution.
+            if ($e instanceof WriteFailedException) {
+                try {
+                    $socketId = $this->client->sendAsyncRequest($this->connection, $request);
+                    $response = $this->client->readResponse($socketId, $timeoutDelayInMs);
+                } catch (Throwable $retryException) {
+                    printf(
+                        "Error communicating with PHP-FPM to read the HTTP response. Bref will restart PHP-FPM now. Original exception message: %s %s | Retry exception: %s %s\n",
+                        get_class($e),
+                        $e->getMessage(),
+                        get_class($retryException),
+                        $retryException->getMessage()
+                    );
+                    $this->stop();
+                    $this->start();
+                    throw new FastCgiCommunicationFailed;
+                }
+            } else {
+                printf(
+                    "Error communicating with PHP-FPM to read the HTTP response. Bref will restart PHP-FPM now. Original exception message: %s %s\n",
+                    get_class($e),
+                    $e->getMessage()
+                );
+                throw new FastCgiCommunicationFailed;
+            }
         }
 
         $responseHeaders = $this->getResponseHeaders($response);


GrahamCampbell (commented):
Are you sure that this only occurs in that situation, and not after being able to write something, and then not being able to write something?

sid-stripe:
> Are you sure that this only occurs in that situation, and not after being able to write something, and then not being able to write something?

Good question:  I'm not 100% certain a partial write can't happen. But I reckon the retry is still safe regardless, because of the sequence of operations:

  1. WriteFailedException is thrown
  2. $this->stop() kills the FPM process that may have received partial data 
  3. $this->start() spawns a completely new FPM process
  4. The retry goes to the new process, which has no knowledge of any previous partial request

  So even in the partial-write case, the old FPM process is destroyed before the retry happens. The new process starts clean.

mnapoli:
This is very interesting, thank you for opening this.

To piggyback on Graham's comment, just to be sure:

> 1. WriteFailedException is thrown
> 2. $this->stop() kills the FPM process that may have received partial data

Could there be any possibility that FPM actually executed anything in the PHP worker between these two steps (and thus making the retry _not_ idempotent).

mnapoli:
To follow up, it might be worth looking into https://github.com/hollodotme/fast-cgi-client (the FastCGI client we use) and maybe the FPM implementation to figure out if PHP scripts could start running *before* everything is written in the FastCGI request?

sid-stripe:
> Could there be any possibility that FPM actually executed anything in the PHP worker between these two steps (and thus making the retry not idempotent).

>To follow up, it might be worth looking into https://github.com/hollodotme/fast-cgi-client (the FastCGI client we use) and maybe the FPM implementation to figure out if PHP scripts could start running before everything is written in the FastCGI request?

Thank you, I will look into those.

sid-stripe:
Hey @mnapoli, @GrahamCampbell, just returning to this after some debugging and looking into the internals of the fast-cgi-client.
  
  
  > Are you sure that this only occurs in that situation, and not after being able to write something, and then not being able to write something?

  There is a single `fwrite()` call:

  - `getRequestPackets()` ([lines 347-391](https://github.com/hollodotme/fast-cgi-client/blob/062182d4eda73c161cc2839783acc83096ec0f37/src/Sockets/Socket.php#L347-L391)) assembles the entire FastCGI request into one string — `BEGIN_REQUEST`, `PARAMS`, `STDIN` chunks, and the empty `STDIN` terminator. 
  - [Line 406](https://github.com/hollodotme/fast-cgi-client/blob/062182d4eda73c161cc2839783acc83096ec0f37/src/Sockets/Socket.php#L406) sends that entire string in a single `fwrite()`. No loop, no second write attempt.

  The "write something, then fail writing something else" scenario cannot happen here.

  > Could there be any possibility that FPM actually executed anything in the PHP worker between these two steps?

  No. FPM only starts executing PHP when it receives the empty `STDIN` terminator — that is the last thing appended to the string:

  - [Line 388](https://github.com/hollodotme/fast-cgi-client/blob/062182d4eda73c161cc2839783acc83096ec0f37/src/Sockets/Socket.php#L388) — empty `STDIN` terminator, the signal to FPM to start executing PHP.
  - [Lines 409-416 ](https://github.com/hollodotme/fast-cgi-client/blob/062182d4eda73c161cc2839783acc83096ec0f37/src/Sockets/Socket.php#L409-L416)— `WriteFailedException` is only thrown when `fwrite()` returns `false` (EPIPE) or `fflush()` fails.

In both cases the terminator was never delivered. If `fwrite()` returns false, nothing was written at all. If `fflush()` fails after a partial write, the terminator is the last thing in the string so it would be the first to be cut off. Either way, FPM is still waiting for input. PHP has not executed, so the retry is safe.

johnnynotsolucky:
WriteFailedException isn't always a broken pipe. This [method](https://github.com/hollodotme/fast-cgi-client/blob/2116cd896dc131df48d3964906c034f388d0b672/src/Sockets/Socket.php#L475) can throw WriteFailedException after writing: https://github.com/hollodotme/fast-cgi-client/blob/2116cd896dc131df48d3964906c034f388d0b672/src/Sockets/Socket.php#L584-L591

PR #2119, opened 2026-06-03
docs: Remove unnecessary extra PHP extension

fagai:
Version 3 does not exist in `bref/extra-redis-php`.
When I checked the upgrade guide, it stated that the Redis extension was internal, so I thought it was a documentation error.


mnapoli:
Thank you!

PR #2140, opened 2026-07-08
Pass explicit character lists to `trim()` calls

GrahamCampbell:
Relying on `trim()`'s default character list is undesirable: it is broader than the whitespace HTTP allows, and the default has changed in PHP 8.6, causing inconsistent behaviour. This passes an explicit list at every call site, narrowing header parsing to RFC 7230 OWS and pinning current behaviour elsewhere.


mnapoli:
Thanks!

PR #2143, opened 2026-07-14
chore: remove redundant ext-json requirement (#2050)

### LINKED ISSUE #2050: Redundant ext-json requirement
Requiring PHP >= 8.0 means that the JSON extension is bundled in the source code and can't be disabled.
The "ext-json" requirement should be removed from composer.json

supermadu7:
Closes #2050

**Motivation**
Since PHP >= 8.0, the JSON extension is bundled directly into the PHP core by default and can no longer be disabled. Because of this, explicitly requiring `"ext-json": "*"` in `composer.json` is completely redundant and can be safely removed to clean up the package dependencies.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,6 @@
     "require": {
         "php": ">=8.2",
         "ext-curl": "*",
-        "ext-json": "*",
         "crwlr/query-string": "^1.0.3",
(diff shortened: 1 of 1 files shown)


mnapoli:
Thanks!

PR #2151, opened 2026-08-03
Allow symfony ^5.4 versions

### LINKED ISSUE #2080: Allow symfony ^5.4 versions
Hi 👋

I noticed that Bref 3 currently requires:

`"symfony/process": "^6.4|^7.0|^8.0"`

Many production projects are still on Symfony 5.4 LTS (supported until Nov 2026). In our case, we’re migrating to Bref 3, but our application stack is still based on Symfony 5.4.

Also, it looks like bref/extra-php-extensions still supports Symfony 5.4, so having bref/bref require Symfony 6+ feels a bit inconsistent across the Bref ecosystem and makes incremental migrations harder.
https://github.com/brefphp/extra-php-extensions/blob/bcca97276283e6e6ad942fcbbb58bde576e062ce/composer.json#L12

Since symfony/process 5.4 supports PHP 8.2, would it make sense to relax the constraint to:

`"symfony/process": "^5.4|^6.4|^7.0|^8.0"`

Additionally, some require-dev deps (e.g. symfony/console, symfony/yaml) currently only allow ^6.4|^7.0|^8.0. I know require-dev doesn’t affect consumers, but allowing ^5.4 there too could help keep testing/dev tooling consistent with Symfony 5.4-based projects during migration.

If Symfony 5.4 is intentionally excluded (testing matrix policy, known incompatibilities, etc.), it would be great to understand the rationale.

Thanks for maintaining Bref 🚀

supermadu7:
symfony/process, symfony/console, and symfony/yaml previously required ^6.4|^7.0|^8.0. Many projects migrating to Bref 3 are still on Symfony 5.4 LTS (supported until Nov 2026), and bref/extra-php-extensions already supports 5.4, so this brings bref/bref in line with the rest of the Bref ecosystem.

symfony/process 5.4 supports PHP 8.2+, matching this package's existing php: >=8.2 requirement, so no conflict there.

Fixes #2080

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         "psr/http-message": "^1.0|^2.0",
         "psr/http-server-handler": "^1.0",
         "riverline/multipart-parser": "^2.1.2",
-        "symfony/process": "^6.4|^7.0|^8.0"
+        "symfony/process": "^5.4|^6.4|^7.0|^8.0"
     },
     "require-dev": {
         "async-aws/core": "^1.0",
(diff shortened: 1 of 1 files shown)


mnapoli:
Thanks!

PR #2168, opened 2026-08-18
Add response stream support for lambda function urls with Bref V3 

tryvin:
And I'm back; this is a follow-up to https://github.com/brefphp/bref/pull/1980

With Lambda Function URLs supporting `RESPONSE_STREAM`, we can make the `HttpHandler` implementations return a `StreamedHttpResponse` instance that supports iterating over a `Generator` and streams the contents as they are received.

This also adds support for PHP-FPM streaming: `FpmHandler` now yields the FastCGI STDOUT chunks as PHP-FPM writes them. The socket reading loop is suspended with a Fiber whenever a chunk arrives, and Lambda's streaming response pulls the chunks on demand, so the response body reaches the client incrementally (time to first byte drops from seconds to milliseconds). The status code and headers are returned as soon as PHP-FPM emits them.

This only works with PHP 8.1+ versions, as it relies on Fibers. When the response is not streamed (e.g., `BREF_STREAMED_MODE` disabled, or non-HTTP-API events like ALB), `FpmHandler` falls back to reading the full response before replying.

Dependency of: https://github.com/brefphp/laravel-bridge/pull/185

## Fallback

Lambda invoke mode `RESPONSE_STREAM` requires you to chunk content even if you are not sending a stream, so `HttpResponse` will either convert a plain string body to a Generator or convert a Generator to a plain string body depending on the `BREF_STREAMED_MODE` env variable.

## Notes

- Streaming from PHP-FPM is only enabled for API Gateway HTTP API events (not ALB or REST APIs), as those are the only integrations supporting streaming responses.
- When the streamed response is already in progress, an error can no longer be reported as an invocation failure (Lambda rejects it): the error is logged, and the caller receives the truncated response instead.


tryvin:
@mnapoli ready for initial review

mnapoli:
Thanks! Huge work, I'd want that for Bref v4 for EOY (no promise :p)

tryvin:
@mnapoli Is there a working branch for V4 already?

mnapoli:
@tryvin no not yet

mnapoli:
(I understand you don't want to re-resolve conflicts again 😅 I'll try to be mindful of that)

